PR #10, opened 2020-03-30
Feature fix include installs

lukasdreyer:


holke:
Before i merge this i will have to clean up the t8_forest_private.h header.
Sticking to its name this header should not be installed und hence included by a user.
However, i understand that some functions that are currently in it should be exposed to the user (such as finding the neighbor elements).

PR #11, opened 2020-03-31
Feature tests

Davknapp:
I implemented a test, that checks up to a given level, if the parent of an element is computed correctly.


holke:
Please issue the pull-request again but into the develop not the master branch

PR #13, opened 2020-03-31
Added test for parent-child-relation

Davknapp:
I implemented a test, that checks if the parent of an element was computed correctly

### DIFF
--- a/test/t8_test_find_parent.cpp
+++ b/test/t8_test_find_parent.cpp
@@ -34,30 +34,27 @@ t8_recursive_child_find_parent (t8_element_t * element,
                                 t8_eclass_scheme_c * ts, int level,
                                 int maxlvl)
 {
-  T8_ASSERT (level <= maxlvl
-             || maxlvl < ts->t8_element_num_children (element));
+  T8_ASSERT (level <= maxlvl &&
+             maxlvl <= ts->t8_element_maxlevel() - 1);
   int                 num_children, i;
   t8_element_t       *child, *test_parent;
-  /*Get number of children */
+  /* Get number of children */
   num_children = ts->t8_element_num_children (element);
-  /*Get child and test_parent, to check if test_parent = parent of child */
+  /* Get child and test_parent, to check if test_parent = parent of child */
   ts->t8_element_new (1, &child);
   ts->t8_element_new (1, &test_parent);
   if (level == maxlvl)
     return;
   for (i = 0; i < num_children; i++) {
-    /*Compute child i */
+    /* Compute child i */
     ts->t8_element_child (element, i, child);
-    /*Compute parent of child */
+    /* Compute parent of child */
     ts->t8_element_parent (child, test_parent);
-    /*If its equal, call child_find_parent, to check if parent-child relation
+    /* If its equal, call child_find_parent, to check if parent-child relation
      * is correct in next level until maxlvl is reached*/
-    if (ts->t8_element_compare (element, test_parent)) {
-      SC_ABORT ("Computed child_parent is not the parent");
-    }
-    else {
-      t8_recursive_child_find_parent (child, ts, level + 1, maxlvl);
-    }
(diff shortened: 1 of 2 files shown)


test/t8_test_find_parent.cpp
```diff
@@ -0,0 +1,96 @@
...
+    /*Compute parent of child */
+    ts->t8_element_parent (child, test_parent);
+    /*If its equal, call child_find_parent, to check if parent-child relation
+     * is correct in next level until maxlvl is reached*/
+    if (ts->t8_element_compare (element, test_parent)) {
+      SC_ABORT ("Computed child_parent is not the parent");
+    }
+    else {
+      t8_recursive_child_find_parent (child, ts, level + 1, maxlvl);
+    }
```
holke:
```suggestion
    SC_CHECK_ABORT (!ts->t8_element_compare (element, test_parent), "Computed child_parent is not the parent");
    t8_recursive_child_find_parent (child, ts, level + 1, maxlvl);
```

We use SC_CHECK_ABORT to check for a condition and abort with a message if the condition is not met.
You can also use SC_CHECK_ABORTF for formatted output in the message.

test/t8_test_find_parent.cpp
```diff
@@ -0,0 +1,96 @@
...
+{
+  T8_ASSERT (level <= maxlvl
+             || maxlvl < ts->t8_element_num_children (element));
+  int                 num_children, i;
+  t8_element_t       *child, *test_parent;
+  /*Get number of children */
+  num_children = ts->t8_element_num_children (element);
+  /*Get child and test_parent, to check if test_parent = parent of child */
+  ts->t8_element_new (1, &child);
+  ts->t8_element_new (1, &test_parent);
```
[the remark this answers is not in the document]
holke: Also make sure that child and test_parent are destroyed again after use.
Currently the memory is not free'd

test/t8_test_find_parent.cpp
```diff
@@ -0,0 +1,96 @@
...
+{
+  t8_element_t       *element;
+  t8_scheme_cxx      *scheme;
+  t8_eclass_scheme_c *ts;
+  int                 eclassi;
+  t8_eclass_t         eclass;
+  scheme = t8_scheme_new_default_cxx ();
+  for (eclassi = T8_ECLASS_ZERO; eclassi < T8_ECLASS_PYRAMID; eclassi++) {
+    eclass = (t8_eclass_t) eclassi;
+    /*Get scheme for eclass */
```
holke:
```suggestion
    /* Get scheme for eclass */
```

test/t8_test_find_parent.cpp
```diff
@@ -0,0 +1,96 @@
...
+    ts = scheme->eclass_schemes[eclass];
+    /*Get element and initialize it */
```
holke:
```suggestion
    /* Get element and initialize it */
```

test/t8_test_find_parent.cpp
```diff
@@ -0,0 +1,96 @@
...
+    ts->t8_element_new (1, &element);
+    ts->t8_element_set_linear_id (element, 0, 0);
+    /*Check for correct parent-child relation */
```
holke:
```suggestion
    /* Check for correct parent-child relation */
```

test/t8_test_find_parent.cpp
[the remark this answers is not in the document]
Davknapp: > Allocate these outside of the recursion and reuse them.
> This way you allocate more elements on each level, which should not be necessary.

I think I can reuse the child, but must store somehow the parent. 
I guess, you have something in mind like
parent = child,
t8_recursive_child_find_parent(child, parent ....)?
But in that way, if we go up again in the recursion, the parent is overwritten and the next computed child is computed at the deepest level, not the next level.

holke:
And right when you think you are done, the tests fail.
Looks to me like we are out of memory in the non-debug case, but i'm just guessing.
We need to investigate further why it crashes now.

PR #45, opened 2020-10-14
Feature notification wrong gmsh version

Niklas997:
An error message will be thrown, if the provided msh-file has an incompatible MeshFormat number and therefore cannot be read by "t8_cmesh_readmshfile.c". Currently supported are msh-files of version 2 or lower.

### DIFF
--- a/scripts/pre-commit
+++ b/scripts/pre-commit
@@ -11,7 +11,7 @@
 
 # This is the indent script in the project's directory
 
-INDENT=./t8indent
+INDENT=./p4estindent
 
 if git rev-parse --verify HEAD >/dev/null 2>&1
 then
--- a/scripts/pre-commit
+++ b/scripts/pre-commit
@@ -10,7 +10,6 @@
 #
 
 # This is the indent script in the project's directory
-
 INDENT=./p4estindent
 
 if git rev-parse --verify HEAD >/dev/null 2>&1
--- a/scripts/pre-commit
+++ b/scripts/pre-commit
@@ -10,7 +10,8 @@
 #
 
 # This is the indent script in the project's directory
-INDENT=./p4estindent
+
+INDENT=./t8indent
 
 if git rev-parse --verify HEAD >/dev/null 2>&1
 then
--- a/src/t8_cmesh/t8_cmesh_readmshfile.c
+++ b/src/t8_cmesh/t8_cmesh_readmshfile.c
@@ -174,17 +174,17 @@ t8_msh_file_node_compare (const void *node_a, const void *node_b,
 }
 
 /* Read an open msh-file and checks whether the MeshFormat-Version is supported by t8code or not. */
-static void
+static int
 t8_cmesh_check_version_of_msh_file (FILE * fp)
 {
-
   char               *line = (char *) malloc (1024);
   char                first_word[2048] = "\0";
-  T8_ASSERT (fp != NULL);
   size_t              linen = 1024;
   int                 retval;
-  int                 max_version_supported = 2;
   int                 version_number, sub_version_number;
+  int                 check_format;
+
+  T8_ASSERT (fp != NULL);
 
   /* Go to the beginning of the file. */
   fseek (fp, 0, SEEK_SET);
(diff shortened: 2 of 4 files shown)


holke:
Could you please add the `author.txt` as we discussed.

scripts/pre-commit
```diff
@@ -10,7 +10,8 @@
 #
 
 # This is the indent script in the project's directory
-INDENT=./p4estindent
+
+INDENT=./t8indent
```
holke, line 14: unnecessary new line here
```suggestion
INDENT=./t8indent
```

holke (commented):
The `p4estindent` replacement should be in a different pull request, since it is unrelated to the gmsh change.
There are two ways to revert the change for this branch:
1. Beginner: Just change the file back, git add and git commit it
2. Advanced: Use interactive rebase to edit the commit and git push -f it (https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History)

src/t8_cmesh/t8_cmesh_readmshfile.c
```diff
@@ -173,6 +173,63 @@ t8_msh_file_node_compare (const void *node_a, const void *node_b,
...
+t8_cmesh_check_version_of_msh_file (FILE * fp)
+{
+
+  char               *line = (char *) malloc (1024);
+  char                first_word[2048] = "\0";
+  T8_ASSERT (fp != NULL);
+  size_t              linen = 1024;
+  int                 retval;
+  int                 max_version_supported = 2;
+  int                 version_number, sub_version_number;
```
holke, line 187: If you can avoid it, try not  to mix declarations and code in `C`
```suggestion
  size_t              linen = 1024;
  int                 retval;
  int                 max_version_supported = 2;
  int                 version_number, sub_version_number;
 
  T8_ASSERT (fp != NULL);
```

src/t8_cmesh/t8_cmesh_readmshfile.c
```diff
@@ -173,6 +173,63 @@ t8_msh_file_node_compare (const void *node_a, const void *node_b,
...
+static void
+t8_cmesh_check_version_of_msh_file (FILE * fp)
+{
+
+  char               *line = (char *) malloc (1024);
+  char                first_word[2048] = "\0";
+  T8_ASSERT (fp != NULL);
+  size_t              linen = 1024;
+  int                 retval;
+  int                 max_version_supported = 2;
```
holke, line 186: Lets make this a macro and put it in the header file.
`#define T8_CMESH_MAX_SUPPORTED_FILE_VERSION 2`

holke:
We should additionally check that the file type is correct, since we only support ASCII and not binary.
The file type is the integer after the version number and must be 0 for ascii.

See also here: https://gmsh.info/doc/texinfo/gmsh.html#MSH-file-format-version-2-_0028Legacy_0029

src/t8_cmesh/t8_cmesh_readmshfile.c
```diff
@@ -173,6 +173,63 @@ t8_msh_file_node_compare (const void *node_a, const void *node_b,
...
+  /* Check if MeshFormat-number is compatible. */
+  if (version_number <= max_version_supported) {
+    t8_debugf ("Succesfully read the MeshFormat-Number.\n");
+    t8_debugf ("This version of msh-file (%d.%d) is supported.\n",
+               version_number, sub_version_number);
+
+  }
+  else {
+    t8_debugf
+      ("This version of msh-file is currently not supported by t8code.\n");
```
holke, line 225: I think this output is unnecessary since it is covered in the error message below.

src/t8_cmesh/t8_cmesh_readmshfile.c
```diff
@@ -173,6 +173,63 @@ t8_msh_file_node_compare (const void *node_a, const void *node_b,
   return Node_a->index == Node_b->index;
 }
 
+/* Read an open msh-file and checks whether the MeshFormat-Version is supported by t8code or not. */
+static void
+t8_cmesh_check_version_of_msh_file (FILE * fp)
+{
+
+  char               *line = (char *) malloc (1024);
```
holke, line 181: There is no `free` to this `malloc`.
Have a look at the other functions and their `goto die...` statements.
You have several points in the functions where you must abort and return. For example when you check the return value of sscanf. In this case, you must free the allocated memory and return. Both currently does not happen.

holke (changes requested):
On the first impression this code is looking good to me.
You kept the implementation close to the other functions in the file and commented everything well. :+1: 

I marked some issues that should be resolved.

Also a main issue: What should happen if the version number is wrong?
In the current state, the function prints an error message, but meshreading continues anyways, which will result in a crash.
The new check function should return an int, which encodes at least these 3 states:
 - correct file version
 - wrong file version
 - could not read file version
A calling function can then decide what to do.
The currently calling `t8_cmesh_from_msh_file`  should then abort the reading with a proper message, close the file and return a NULL cmesh.

At the end, the code should not crash, but rather abort peacefully and tell the user whats wrong.

scripts/pre-commit
```diff
@@ -10,6 +10,7 @@
 #
 
 # This is the indent script in the project's directory
+
```
holke, line 13:
```suggestion
```
Somehow a new line got introduced here.

src/t8_cmesh/t8_cmesh_readmshfile.c
```diff
@@ -834,7 +851,13 @@ t8_cmesh_from_msh_file (const char *fileprefix, int partition,
       return NULL;
     }
     /* Check if msh-file version is compatible. */
-    t8_cmesh_check_version_of_msh_file (file);
+    if (t8_cmesh_check_version_of_msh_file (file) != 1) {
+      /* If reading the MeshFormat-number failed or the version is incompatible, close the file */
+      fclose (file);
+      t8_debugf
+        ("The reading process of the msh-file has failed and the file has been closed.\n");
```
holke, line 858: This return leaves the allocated cmesh dangling and the memory is not freed -> memory leak.

We can fix it by destroying the cmesh before returning:
`t8_cmesh_destroy (&cmesh);`

holke (changes requested):
Very good update.
 This adresses all my previous concerns and looks good now.

I only have minor corrections. Once these are changed we can merge.

The next stept then could be to update the gmsh example such that it does not crash anymore when the file version is wrong, but exits properly.

src/t8_cmesh/t8_cmesh_readmshfile.c
```diff
@@ -776,6 +850,14 @@ t8_cmesh_from_msh_file (const char *fileprefix, int partition,
       t8_global_errorf ("Could not open file %s\n", current_file);
```
holke, line 850: Definitely not your fault, but while checking this i notices that a cmesh destroy is also missing in this line, causing a memory leak when the file cannot be opened.
Could you please add this as well?

src/t8_cmesh/t8_cmesh_readmshfile.c
```diff
@@ -198,36 +198,53 @@ t8_cmesh_check_version_of_msh_file (FILE * fp)
...
-    t8_global_errorf ("The MeshFormat-number was not read correctly.\n");
+  if (retval != 3) {
     t8_debugf ("Reading of the MeshFormat-number failed.\n");
+    goto die_format;
+  }
+
+  /* Checks if the file is of Binary-type. */
+  if (check_format) {
+    t8_global_errorf
+      ("Incompatible file-type. t8code works with ASCII-type msh-files of version %d.\n",
```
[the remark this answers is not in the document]
Niklas997: Aren't the msh-files of version 1 created by gmsh structured slighlty different than those of version 2? (https://gmsh.info/doc/texinfo/gmsh.html#MSH-file-format-version-1-_0028Legacy_0029)  There seems to be no MeshFormat number and the parts containing the nodes and elements are introduced by '$NOD' and '$ELM' instead of '$Nodes' and '$Elements' like in files of version 2. Apart from that, in the element section the third entry in version 2 is the 'number-of-tags' followed by these specified tags before the 'node-number-lists' starts. These tags are skipped by the function 't8_cmesh_msh_file_read_eles()'  by reading the number of tags and leaving the following entries out, whereas, in version 1 are always 3 entries before the 'node-number-list', describing the elements, starts. But they aren't preceded by the 'number-of-tags'. Therefore, the function doesn't read the elements properly, does it? So, version 1 files can't be read or am I wrong?

src/t8_cmesh_readmshfile.h
```diff
@@ -27,12 +27,13 @@
 
 #ifndef T8_CMESH_READMSHFILE_H
 #define T8_CMESH_READMSHFILE_H
-#define T8_CMESH_MAX_SUPPORTED_FILE_VERSION 2
 
 #include <t8.h>
 #include <t8_eclass.h>
 #include <t8_cmesh.h>
```
holke, line 34:
```suggestion

/* The maximum supported .msh file version.
 * Currently, we support gmsh's file version 2 in ASCII format.
 */
```

holke:
This looks good now and i will merge it.

I just added a comment to the macro.

holke (approved):
Approved :+1:

PR #50, opened 2020-10-28
Feature notification wrong gmsh version

holke:
These are updates to https://github.com/holke/t8code/pull/45.
In the previous version we did not complain about msh file version below 2.2; however, we actually do not support these.
This should be fixed now with appropriate error messages.

Since #45 was merged already, we need to open a new pull request.


holke (commented):
This is all looking good for me.

PR #51, opened 2020-10-30
Created a test in order to check the reading of msh-files

Niklas997:


test/t8_test_cmesh_readmshfile.c
```diff
@@ -0,0 +1,164 @@
...
+  SC_CHECK_MPI (mpiret);
+
+  sc_init (sc_MPI_COMM_WORLD, 1, 1, NULL, SC_LP_ESSENTIAL);
+  p4est_init (NULL, SC_LP_ESSENTIAL);
+  t8_init (SC_LP_DEFAULT);
+
+/* The mesh is unpartitioned and all trees are local to the master rank, no other rank has any trees. */
+
+  t8_global_productionf ("Testing: Reading of msh-files.\n");
+
```
holke, line 117: Please write an extra function for each test (or one test function with a parameter specifying the test).

The main function of the tests should not contain too much code, rather only call other functions.

test/t8_test_cmesh_readmshfile.c
```diff
@@ -0,0 +1,164 @@
...
+  /* Testing supported msh-file version 2 (as example). */
+  cmesh =
+    t8_cmesh_from_msh_file ("test/test_msh_file_vers2_ascii", 1,
+                            sc_MPI_COMM_WORLD, 2, 0);
+  retval = t8_test_supported_msh_file (cmesh);
+  if (retval != 1) {
+    t8_global_errorf ("Reading of supported msh-file version failed.\n");
+    goto end_test;
```
holke, line 125: We should avoid goto here.
It is only necessary in a few cases, which are mostly related to file i/o.

When you integrate these lines in their own function (see comment above) the go to will go away anyways i guess.

test/t8_test_cmesh_readmshfile.c
```diff
@@ -0,0 +1,164 @@
...
+  SC_CHECK_MPI (mpiret);
+/* If the test was succesful. */
+  return 0;
+
+/* In case anything failed, the test is going to be closed. */
+end_test:
+  t8_debugf ("The test was not succesful.\n");
+  if (t8_cmesh_is_committed (cmesh)) {
+    t8_cmesh_destroy (&cmesh);
+  }
```
holke, line 162: Even if the test fail you should properly finalize sc and MPI.

test/t8_test_cmesh_readmshfile.c
```diff
@@ -0,0 +1,164 @@
+#include <t8_eclass.h>
```
holke, line 1: The licence information is missing.

Look at the other files, each file in t8code needs a licence text at the beginning.

holke, line 1: After the licence text, please add a brief description of the purpose of this test.

holke (changes requested):
This looks very good to me.
Just a few remarks, mostly about comments and code structure.

The code itself looks good and is well documented.

We still need to find out how to move files during build though...

holke:
If we do not execute this test from the main folder, the files are not found.

If the files are not found an error message should print:
`msh files not found. Please ensure that you execute this test from the main directory (t8code)`

PR #61, opened 2021-01-19
Feature googletest

Davknapp:


test/t8_gtest_main.cxx
```diff
@@ -0,0 +1,29 @@
+#include <mpi.h>
```
holke, line 1: The mpi.h include should be wrapped in #ifdef T8_ENABLE_MPI

holke (changes requested):
Thanks, good work. Only minor changes needed.

There is some unrelated stuff here, please put it in a seperate PR "Code beautifications" or similar.

I thought about it and do not want to integrate this into develop until we have the parallel gtest running.
So we should change the base branch to holke/feature-googletest

holke:
ahh, changed the base to holke/feature-googletest and  90% of my remarks are outdated now ;)

PR #67, opened 2021-03-05
Fix first_tree_shared computation in uniform_bounds

holke:


src/t8_cmesh/t8_cmesh_cxx.cxx
```diff
@@ -124,9 +125,10 @@ t8_cmesh_uniform_bounds (t8_cmesh_t cmesh, int level,
     T8_ASSERT (0 <= last_global_child
                && last_global_child <= global_num_children);
     *first_local_tree = first_global_child / children_per_tree;
+    *child_in_tree_begin_temp =
```
holke, line 128: `child_in_tree_begin_temp =`

holke:
```suggestion
    child_in_tree_begin_temp =
```

PR #73, opened 2021-03-18
Point inside check

holke:
Implement a check whether a point lies inside an element and add a test.

Also required to implement face normals for lines.

Note: PR #74  depends on this PR

### DIFF
--- a/src/t8_forest/t8_forest_cxx.cxx
+++ b/src/t8_forest/t8_forest_cxx.cxx
@@ -989,6 +989,218 @@ t8_forest_element_face_normal (t8_forest_t forest, t8_locidx_t ltreeid,
   }
 }
 
+int
+t8_forest_element_point_inside (t8_forest_t forest, t8_locidx_t ltreeid,
+                                const t8_element_t * element,
+                                const double *tree_vertices,
+                                const double point[3], const double tolerance)
+{
+  t8_eclass_t         tree_class = t8_forest_get_tree_class (forest, ltreeid);
+  t8_eclass_scheme_c *ts = t8_forest_get_eclass_scheme (forest, tree_class);
+  t8_element_shape_t  element_shape = ts->t8_element_shape (element);
+  int                 num_faces = ts->t8_element_num_faces (element);
+  int                 iface;
+  int                 afacecorner;
+  double              face_normal[3];
+  double              dot_product;
+  double              point_on_face[3];
+
+  switch (element_shape) {
+  case T8_ECLASS_VERTEX:
+    {
+      /* A point is 'inside' a vertex if they have the same coordinates */
+      double              vertex_coords[3];
+      /* Get the vertex coordinates */
+      t8_forest_element_coordinate (forest, ltreeid, element, tree_vertices,
+                                    0, vertex_coords);
+      /* Check whether the point and the vertex are within tolerance distance
+       * to each other */
+      if (t8_vec_dist (vertex_coords, point) > tolerance) {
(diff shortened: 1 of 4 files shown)


src/t8_forest/t8_forest_cxx.cxx
```diff
@@ -960,6 +989,218 @@ t8_forest_element_face_normal (t8_forest_t forest, t8_locidx_t ltreeid,
   }
 }
 
+int
```
Davknapp, line 992: Shouldn't these functions be element-functions and therefore be implemented there?
This would avoid this very long switch blocks. 
On the other hand this would mean code-duplication for the bilinear-interpolated elements. If there was a way to call a more general function for these elements, this function would be much more pretty.

src/t8_forest/t8_forest_cxx.cxx
```diff
@@ -960,6 +989,218 @@ t8_forest_element_face_normal (t8_forest_t forest, t8_locidx_t ltreeid,
...
+t8_forest_element_point_inside (t8_forest_t forest, t8_locidx_t ltreeid,
+                                const t8_element_t * element,
+                                const double *tree_vertices,
+                                const double point[3], const double tolerance)
+{
+  t8_eclass_t         tree_class = t8_forest_get_tree_class (forest, ltreeid);
+  t8_eclass_scheme_c *ts = t8_forest_get_eclass_scheme (forest, tree_class);
+  t8_element_shape_t  element_shape = ts->t8_element_shape (element);
+  int                 num_faces = ts->t8_element_num_faces (element);
```
Davknapp, line 1001:
```suggestion
 const  int           num_faces = ts->t8_element_num_faces (element);
```

Davknapp (commented):
We should discuss if there is a way to make the point inside function to element-functions. If so, we should change that before merging. Otherwise, this can be merged.

holke:
> 
> 
> We should discuss if there is a way to make the point inside function to element-functions. If so, we should change that before merging. Otherwise, this can be merged.

No it cannot be an element function, since it is a question of geometry. The element functions are topology only and do not know anything about geometry.
This becomes much clearer when we include high-order geometries with curved elements.
The underlying element functions are still the same nevertheless.

test/t8_test_point_inside.cxx
```diff
@@ -0,0 +1,286 @@
...
+     * the sum of the already set entries is.
+     * This dampening increases the percentage of points that lie inside the element.
+     *
+     * The number of steps depends on the number of corners.
+     * The more corners the smaller the number of stepes in order to limit the
+     * number of points that we generate.
+     */
+    num_points = 10000;         /* Desired number of points */
+    /* Compute the number of steps needed */
+    num_steps = pow (num_points, 1. / (num_corners - 1));
```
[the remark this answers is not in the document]
holke: No. There is a double^int one, but not the other way round.

src/t8_forest/t8_forest_cxx.cxx
holke: No, as said above they are part of the geometry. I guess this handling will also change once we merge https://github.com/holke/t8code/tree/feature-geometry

src/t8_forest/t8_forest_cxx.cxx
holke: Done

holke:
I adressed your comments. If you are satisfied please approve :)

Davknapp (approved):

PR #76, opened 2021-03-19
Improve the github CI by uploading logs of failed runs

holke:
The CI script now uploads configure and make logs as artifacts.

See my test runs and their artifacst https://github.com/holke/t8code/actions/runs/2301835043 and https://github.com/holke/t8code/actions/runs/2301750767 to see how it works.


scripts/pre-commit
```diff
@@ -30,6 +30,7 @@
 # exit with non-zero status after issuing an appropriate message if
 # it wants to stop the commit.
 #
+# Copyright Johannes Holke 2020
```
holke, line 33:
```suggestion
```

holke:
meanwhile we handled the indentation in a different PR.

This PR now only adds uploading of the configure and make check error logs as artifacts.

holke:
I tested this with a failing configure: https://github.com/holke/t8code/actions/runs/2301835043

configure.ac
```diff
@@ -30,6 +30,8 @@ echo "o---------------------------------------"
 echo "| Checking MPI and related programs"
 echo "o---------------------------------------"
 
+
+
```
holke, line 34:
```suggestion
```

config/t8_netcdf.m4
```diff
@@ -31,6 +31,7 @@ if test "x$T8_WITH_NETCDF" != xno ; then
 ]],[[
   #include <netcdf.h>
 
+  this line will break the code
```
holke, line 34:
```suggestion

```

config/t8_netcdf.m4
```diff
@@ -31,6 +31,7 @@ if test "x$T8_WITH_NETCDF" != xno ; then
 ]],[[
   #include <netcdf.h>
```
holke, line 33:
```suggestion

```

config/t8_netcdf.m4
```diff
@@ -30,7 +30,6 @@ if test "x$T8_WITH_NETCDF" != xno ; then
 [[
 ]],[[
   #include <netcdf.h>
```
holke, line 32:
```suggestion
  #include <netcdf.h>
  
```

holke:
> After one test failing, the github CI skips the remaining tests, but the OnFailUploadLog still gets triggered. They then fail themselves, because there in no log to upload, because the test never ran. It is not exactly a problem, but it does not look nice https://github.com/holke/t8code/runs/6373185043?check_suite_focus=true

Yes, i agree. We have this behavior from the beginning and its not directly related to uloading the logs.
Maybe there is a way to exit the action once it has failed...if so i could add it.

.github/workflows/tests.yml
```diff
@@ -291,70 +291,125 @@ jobs:
...
-    - name: OnFailPrintLog
+    - name: OnFailUploadLog
       if: failure()
-      run: cat build_release/config.log
+      uses: actions/upload-artifact@v1
+      with:
+        name: config_release_MPI.log
+        path: build_release/config.log
     - name: make
       run: cd build_release && make -j V=0 CFLAGS="$CFLAGS_var -Werror" CXXFLAGS="$CXXFLAGS_var -Werror"
```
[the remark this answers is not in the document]
holke: The `make` step does not produce any log files that we could upload.
We get the output of `make` in the console which we can look at in the browser.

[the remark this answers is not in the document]
sandro-elsweijer: We could use `make > make_release_MPI.log 2>&1`. But this was just an idea, I do not know how relevant this is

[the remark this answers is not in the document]
holke: I think its not necessary to do that.

holke:
From my point of view this is done.
You can approve and merge :)

sandro-elsweijer (approved):

Issue #79, opened 2021-03-19
Cmesh_new_hypercube does not work with partition flag


holke:
It seems that the vertex coordinates are not partitioned correctly.

To reproduce:

```
 cmesh = t8_cmesh_new_hypercube (T8_ECLASS_TET, comm, 0, 1, 0);
 t8_cmesh_vtk_write_file (cmesh, "vtk_filename", 1.0);
```

holke:
This bug was also verified for the `cmesh_new_from_p{4,8}est` functions.
The `example/cmesh/t8_cmesh_partition` example does not work anymore.

lukasdreyer:
First thing that I stumbled upon:

The function `t8_cmesh_add_attributes` is called by `t8_cmesh_commit_replicated_new` and `t8_cmesh_commit_partitioned_new`, so for me it looks like the comment 
```
      /* attribute->id is a gloidx that is casted to a locidx here.
       * Should not cause problems, since mesh is replicated */
```
from below might not be true.

```
void
t8_cmesh_add_attributes (t8_cmesh_t cmesh)
{
  t8_stash_attribute_struct_t *attribute;
  t8_stash_t          stash = cmesh->stash;
  t8_locidx_t         ltree;
  size_t              si, sj;

  ltree = -1;
  for (si = 0, sj = 0; si < stash->attributes.elem_count; si++, sj++) {
    attribute = (t8_stash_attribute_struct_t *)
      sc_array_index (&stash->attributes, si);
    if (cmesh->first_tree <= attribute->id &&
        attribute->id < cmesh->first_tree + cmesh->num_local_trees) {
      if (attribute->id > ltree) {
        /* Enter a new tree */
        ltree = attribute->id;
        sj = 0;
      }
      /* attribute->id is a gloidx that is casted to a locidx here.
       * Should not cause problems, since mesh is replicated */
      T8_ASSERT (attribute->id - cmesh->first_tree ==
                 (t8_locidx_t) attribute->id - cmesh->first_tree);
      t8_cmesh_trees_add_attribute (cmesh->trees, 0, attribute,
                                    attribute->id - cmesh->first_tree, sj);
    }
  }
}
```
I will try to gain a deeper understanding about the different kinds of local and global indices from the wiki page.

lukasdreyer:
I think I need a highlevel overview of the capabilities that `t8_cmesh_commit_partitioned_new` should have.
[...]

PR #80, opened 2021-03-19
Feature error message in cmesh hypercube

holke:
Print an error message when cmesh hypercube is called with partition flag.
See #79


holke:
We cannot merge this, since it breaks a lot of tests.

Which is good! It shows us how often we use cmesh_new_hypercube with the partition flag on.

PR #88, opened 2021-03-29
Feature write netcdf files

Niklas997:
Made changes regarding the parallel writing performance of a forest to the netCDF File and added an extended function to select between different per variable storage and mpi-access patterns.


src/t8_forest/t8_forest_netcdf.cxx
```diff
@@ -1059,6 +1168,21 @@ t8_forest_write_netcdf (t8_forest_t forest, const char *file_prefix,
...
   context.convention = "UGRID v1.0";
+  /* Check the given 'netcdf_storage_mode' and 'netcdf_mpi_access' */
+  if ((netcdf_var_storage_mode != NC_CONTIGUOUS
+       && netcdf_var_storage_mode != NC_CHUNKED)
+      || (netcdf_mpi_access != NC_INDEPENDENT
+          && netcdf_mpi_access != NC_COLLECTIVE)) {
+    t8_global_productionf
+      ("Illegal input parameters for either the storage-mode (NC_CONTIGUOUS or NC_CHUNKED) or the netcdf variable mpi-access (NC_INDEPENDENT or NC_COLLECTIVE) were given.\nTherefore, default values (NC_CONTIGUOUS, NC_INDEPENDENT) will be used.\n");
+    context.netcdf_var_storage_mode = NC_CONTIGUOUS;
+    context.netcdf_mpi_access = NC_INDEPENDENT;
```
holke, line 1179: I guess this function should exit now with an error code?

src/t8_forest_netcdf.h
```diff
@@ -33,6 +33,29 @@
...
+ * \param [in]  forest    A forest.
+ * \param [in]  file_prefix    A string which holds the file's name (output file will be 'file_prefix.nc').
+ * \param [in]  file_title    A string to caption the NetCDF-File.
+ * \param [in]  dim    The Dimension of the forest mesh (2D or 3D).
+ * \param [in]  num_extern_netcdf_vars    The number of extern user-defined variables which hold elementwise data (if none, set it to 0).
+ * \param [in]  ext_variables An array of pointers of the herein before mentioned user-defined variables (if none, set it to NULL).
+ * \param [in]  comm The sc_MPI_Communicator to use.
+ * \param [in]  netcdf_var_storage_mode This defines whether the variables in the netCDF-4 File are stored contiguously (NC_CONTIGUOUS) or in chunks (NC_CHUNKED (in this case netCDF's default chunking scheme is used)).
+ * \param [in]  netcdf_var_mpi_access This defines whether the mpi processes are writing the variables' data collectively or independently.
+ */
```
holke, line 45:
```suggestion
 * \note Only use this extended version of \ref t8_forest_write_netcdf if you know exactly what you are doing. In all other cases, we recommend using \ref t8_forest_write_netcdf instead.
 */
```

example/netcdf/t8_write_forest_netcdf.cxx
```diff
@@ -13,6 +13,99 @@
...
+  /* Create a default scheme */
+  default_scheme = t8_scheme_new_default_cxx ();
+
+  /* Construct a 3D hybrid hypercube as a cmesh */
+  cmesh = t8_cmesh_new_hypercube_hybrid (3, comm, 0, 0);
+
+  /* Build a (partioined) uniform forest */
+  forest =
+    t8_forest_new_uniform (cmesh, default_scheme, forest_refinement_level, 0,
+                           comm);
```
holke, line 65: Possible extension for the future: Adapt the forest before writing so that we do not always write a uniform forest.

example/netcdf/t8_write_forest_netcdf.cxx
```diff
@@ -13,6 +13,99 @@
...
+static void
+t8_example_time_netcdf_writing_operation (t8_forest_t forest,
+                                          sc_MPI_Comm comm,
+                                          int netcdf_var_storage_mode,
+                                          int netcdf_var_mpi_access,
+                                          const char *title)
+{
+#if T8_WITH_NETCDF
+  double              start_time, end_time, duration, global;
+  int                 retval;
```
holke, line 26: I would prefer if we had additional data on the forest as is the case for an application code.

Can we add at lease one integer and one double field here?
Maybe fill with random data or (if we are interested in compression) some function values?

example/netcdf/t8_write_forest_netcdf.cxx
```diff
@@ -13,6 +13,99 @@
...
+  sc_MPI_Barrier (comm);
+  start_time = sc_MPI_Wtime ();
```
holke, line 28: Note: In future we can do the time measurement via sc_statistics.
If you look for example into the t8_time_forest_partition example how its done.
The sc_statistics can compute min,max and average for all MPI ranks by itself.

holke (changes requested):
This look very good, thanks.

Only some minor comments, the most important for me would be to add additional data for a more realistic use case feel.
(The data could also be optional, if a bool is true or something)

Also i noticed your name is not in the AUTHORS file.
Could you please add yourself under Contributors

PR #94, opened 2021-04-13
Bug fix negative prism volume

Davknapp:


holke (approved):
Nothing much to review here ;)
Thanks for the fix, looks good to me, you are changing the order of two vertex pairs, which turn negative volumes into positive ones.


Overall, this function could benefit from more comments.
How does the geometry look like?
Where do the different prisms start and end?
Why cos(300 ....)?

You could add these here or open a new PR.

test/t8_test_find_parent.cpp
```diff
@@ -28,6 +28,7 @@
  * again the input element. Then call this function for every child, until maxlevel is
  * reached.
  */
+
```
holke, line 31:
```suggestion
```

Davknapp:
> Nothing much to review here ;)
> Thanks for the fix, looks good to me, you are changing the order of two vertex pairs, which turn negative volumes into positive ones.
> 
> Overall, this function could benefit from more comments.
> How does the geometry look like?
> Where do the different prisms start and end?
> Why cos(300 ....)?
> 
> You could add these here or open a new PR.

Added Comments in the latest commit. I construct the prisms, by diverting the unit-circle into chunks of 60 degrees. At this point of the computation I need the fifth chunk -> 5*60 = 300

src/t8_cmesh/t8_cmesh.c
```diff
@@ -2722,14 +2722,20 @@ t8_cmesh_new_prism_cake_funny_oriented (sc_MPI_Comm comm)
   return cmesh;
 }
 
+/*Creates a mesh consisting of 8 prisms. The first 6 prisms are constructed, by
```
holke, line 2725:
```suggestion
/* Creates a mesh consisting of 8 prisms. The first 6 prisms are constructed, by
```

PR #119, opened 2021-06-14
Feature use vtk features

Veli-hub:
write vtu files with vtk own library structure and functions


src/t8_forest/t8_forest_vtk.cxx
```diff
@@ -94,6 +122,190 @@ typedef int         (*t8_forest_vtk_cell_data_kernel) (t8_forest_t forest,
...
+  T8_ASSERT (forest->rc.refcount > 0);
+  T8_ASSERT (forest->committed);
+  T8_ASSERT (fileprefix != NULL);
+
+  long int            point_id = 0;     /* The id of the point in the points Object. */
+  t8_cmesh_t          cmesh;
+  t8_locidx_t         ielement;
+  t8_tree_t           tree;
+  t8_locidx_t         itree, ivertex;
+  t8_ctree_t          ctree;
```
holke, line 140: ctree can be removed

src/t8_forest/t8_forest_vtk.cxx
```diff
@@ -94,6 +122,190 @@ typedef int         (*t8_forest_vtk_cell_data_kernel) (t8_forest_t forest,
...
+  /*Check assertions: forest and fileprefix are not NULL and forest is commited */
+  T8_ASSERT (forest != NULL);
+  T8_ASSERT (forest->rc.refcount > 0);
+  T8_ASSERT (forest->committed);
+  T8_ASSERT (fileprefix != NULL);
+
+  long int            point_id = 0;     /* The id of the point in the points Object. */
+  t8_cmesh_t          cmesh;
+  t8_locidx_t         ielement;
+  t8_tree_t           tree;
```
holke, line 138: can be removed and replaced with a `t8_eclass_t eclass`

src/t8_forest/t8_forest_vtk.cxx
```diff
@@ -94,6 +122,190 @@ typedef int         (*t8_forest_vtk_cell_data_kernel) (t8_forest_t forest,
...
+  cmesh = t8_forest_get_cmesh (forest);
+  /* The cellTypes Array stores the element types as integers(see vtk doc) */
+  int                *cellTypes =
+    T8_ALLOC (int, t8_forest_get_num_element (forest));
+/* We iterate over all local trees*/
+  for (itree = 0;
+       (int) itree < t8_forest_get_num_local_trees (forest); itree++) {
+
+    vertices = t8_forest_get_tree_vertices (forest, itree);
+    tree = t8_forest_get_tree (forest, itree);
```
holke, line 175: no need to get the tree.
Use `t8_forest_get_tree_class` instead to get the trees class.

src/t8_forest/t8_forest_vtk.cxx
```diff
@@ -94,6 +122,190 @@ typedef int         (*t8_forest_vtk_cell_data_kernel) (t8_forest_t forest,
...
+    t8_eclass_scheme_c *scheme =
+      t8_forest_get_eclass_scheme (forest, tree->eclass);
+    t8_locidx_t         elems_in_tree =
+      (t8_locidx_t) t8_element_array_get_count (&tree->elements);
```
holke, line 179: Use `t8_forest_get_tree_element_count` instead

src/t8_forest/t8_forest_vtk.cxx
```diff
@@ -94,6 +122,190 @@ typedef int         (*t8_forest_vtk_cell_data_kernel) (t8_forest_t forest,
...
+    /* We iterate over all elements in the tree */
+    for (ielement = 0; ielement < elems_in_tree; ielement++) {
+      t8_element_t       *element =
+        t8_forest_get_element_in_tree (forest, itree, ielement);
+      T8_ASSERT (element != NULL);
+      t8_element_shape_t  element_shape = scheme->t8_element_shape (element);
+      /* For each element we iterate over all points */
+      for (ivertex = 0; ivertex < t8_eclass_num_vertices[tree->eclass];
+           ivertex++, point_id++) {
```
holke, line 188: Use `t8_element_num_corners` instead of `t8_eclass_num_vertices`.
The eclass is only for the tree.
Imagine a pyramid tree that has a tet element.
You want 4 vertices for the tet. The eclass will however give you 5 since it stores 'pyramid'.

src/t8_forest/t8_forest_vtk.cxx
```diff
@@ -94,6 +122,190 @@ typedef int         (*t8_forest_vtk_cell_data_kernel) (t8_forest_t forest,
...
+        /* We take the element coordinates in vtk order */
+        t8_forest_element_coordinate (forest, itree, element,
+                                      vertices,
+                                      t8_eclass_vtk_corner_number
+                                      [tree->eclass]
+                                      [ivertex], coordinates);
+        x = coordinates[0];
+        y = coordinates[1];
+        z = coordinates[2];
```
holke, line 197: x,y,z are superfluos.
Just use coordinates[0],[1],[2] below

src/t8_forest/t8_forest_vtk.cxx
```diff
@@ -94,6 +122,190 @@ typedef int         (*t8_forest_vtk_cell_data_kernel) (t8_forest_t forest,
...
+          tet->GetPointIds ()->SetId (ivertex, point_id);
+          break;
+        case T8_ECLASS_PRISM:
+          prism->GetPointIds ()->SetId (ivertex, point_id);
+          break;
+        case T8_ECLASS_PYRAMID:
+          pyramid->GetPointIds ()->SetId (ivertex, point_id);
+          break;
+        default:
+          printf ("Kein Elementtyp.\n");
```
holke, line 227:
```suggestion
          SC_ABORT_NOT_REACHED ();
```

This will properly abort the code.

src/t8_forest/t8_forest_vtk.cxx
```diff
@@ -94,6 +122,190 @@ typedef int         (*t8_forest_vtk_cell_data_kernel) (t8_forest_t forest,
...
+        cellArray->InsertNextCell (tet);
+        break;
+      case T8_ECLASS_PRISM:
+        cellArray->InsertNextCell (prism);
+        break;
+      case T8_ECLASS_PYRAMID:
+        cellArray->InsertNextCell (pyramid);
+        break;
+      default:
+        printf ("Kein Elementtyp.\n");
```
holke, line 259:
```suggestion
          SC_ABORT_NOT_REACHED ();
```

src/t8_forest/t8_forest_vtk.cxx
```diff
@@ -94,6 +122,190 @@ typedef int         (*t8_forest_vtk_cell_data_kernel) (t8_forest_t forest,
...
+        break;
+      }
+      cellTypes[elem_id] = t8_eclass_vtk_type[element_shape];
+      elem_id++;
+    }
+  }
+
+  // Write file
```
holke, line 267: // comment

src/t8_forest/t8_forest_vtk.cxx
```diff
@@ -94,6 +122,190 @@ typedef int         (*t8_forest_vtk_cell_data_kernel) (t8_forest_t forest,
...
+  vtkNew < vtkUnstructuredGrid > unstructuredGrid;
+  unstructuredGrid->SetPoints (points);
+  unstructuredGrid->SetCells (cellTypes, cellArray);
+
+  std::string mpifilename = fileprefix;
+  mpifilename += ".pvtu";
+
+//   auto pwriterObj = vtkSmartPointer<vtkXMLPUnstructuredGridWriter>::New();
```
holke, line 275: remove this line?

src/t8_forest/t8_forest_vtk.cxx
```diff
@@ -94,6 +122,190 @@ typedef int         (*t8_forest_vtk_cell_data_kernel) (t8_forest_t forest,
...
+  vtkSmartPointer < vtkXMLPUnstructuredGridWriter > pwriterObj =
+    vtkSmartPointer < vtkXMLPUnstructuredGridWriter >::New ();
+
+  pwriterObj->EncodeAppendedDataOff ();
```
holke, line 279: What does this do?
What would change if we did not call this?

src/t8_forest/t8_forest_vtk.cxx
```diff
@@ -94,6 +122,190 @@ typedef int         (*t8_forest_vtk_cell_data_kernel) (t8_forest_t forest,
...
+    }
+  }
+
+  // Write file
+  vtkNew < vtkUnstructuredGrid > unstructuredGrid;
+  unstructuredGrid->SetPoints (points);
+  unstructuredGrid->SetCells (cellTypes, cellArray);
+
+  std::string mpifilename = fileprefix;
+  mpifilename += ".pvtu";
```
holke, line 273: We avoid using std functionalities and should stick to raw c strings to be consistent with the rest of t8code.

```suggestion
  char mpifilename[BUFSIZ] = fileprefix;
  snprintf (mpifilename, BUFSIZ, "%s.pvtu", fileprefix);
```

src/t8_forest/t8_forest_vtk.cxx
```diff
@@ -94,6 +122,190 @@ typedef int         (*t8_forest_vtk_cell_data_kernel) (t8_forest_t forest,
...
+  vtkNew < vtkLine > line;
+  vtkNew < vtkQuad > quad;
+  vtkNew < vtkTriangle > tri;
+  vtkNew < vtkPyramid > pyramid;
+  vtkNew < vtkWedge > prism;
+  vtkNew < vtkTetra > tet;
+
+  std::string filename = fileprefix;
+  // Append process number
+  filename += "_" + std::to_string (forest->mpirank);
```
holke, line 161: use c style char. See comment below.

holke (changes requested):
All in all this looks very good.

I marked some comments on the macro usage, c++ style and usage of t8codes forest functions.

Could you please add a bit more documentation in the cxx file, what the different sections do.

src/t8_forest/t8_forest_vtk.cxx
Veli-hub: I cite: Get/Set whether the appended data section is base64 encoded. If encoded, reading and writing will be slower, but the file will be fully valid XML and text-only. If not encoded, the XML specification will be violated, but reading and writing will be fast. The default is to do the encoding. 

Do we want to keep this then?

src/t8_forest/t8_forest_vtk.cxx
Veli-hub: But we want to use t8_forest_get_tree_element_count instead of  t8_element_array_get_count (&tree->elements); and it expects the tree as input.

holke:
Thanks for the explanation. I think we should leave it then.
Let us include this text as a comment and also a link to the document.

src/t8_forest/t8_forest_vtk.cxx
holke: Thanks for the explanation. I think we should leave it then.
Let us include this text as a comment and also a link to the document.

src/t8_forest/t8_forest_vtk.cxx
holke: Correction:

We should use
```C
t8_locidx_t         t8_forest_get_tree_num_elements (t8_forest_t forest,
                                                     t8_locidx_t ltreeid);
```

src/t8_forest/t8_forest_vtk.cxx
holke: Sorry, my bad. `t8_forest_get_tree_num_elements` is the correct function. see comment below.

src/t8_forest/t8_forest_vtk.cxx
Veli-hub: I get '‘t8_element_num_corners’ was not declared in this scope' even though we have included:
#include <t8_schemes/t8_default_cxx.hxx>
#include <t8_schemes/t8_default/t8_default_common_cxx.hxx>

holke: This function is part of the element scheme, so you need to call it as `scheme->t8_element_num_corners`. This should work.

src/t8_forest/t8_forest_vtk.cxx
Veli-hub: This gives me an error: error: array must be initialized with a brace-enclosed initializer

holke: ah, yes. You cannot initialize a char array with a pointer.
(i really should check the code i i suggest here).
```C
  char mpifilename[BUFSIZ];
   snprintf (mpifilename, BUFSIZ, "%s.pvtu", fileprefix);
```

src/t8_forest/t8_forest_vtk.cxx
```diff
@@ -147,45 +158,53 @@ t8_write_vtk_via_API (t8_forest_t forest, const char *fileprefix)
...
-         ielement++) {
       t8_element_t       *element =
         t8_forest_get_element_in_tree (forest, itree, ielement);
-      T8_ASSERT(element!=NULL);
+      T8_ASSERT (element != NULL);
       t8_element_shape_t  element_shape = scheme->t8_element_shape (element);
 
-      for (ivertex = 0; ivertex < t8_eclass_num_vertices[tree->eclass];
+      /* For each element we iterate over all points */
+      for (ivertex = 0; ivertex < scheme->t8_element_num_corners (element);
```
holke, line 195: As a tip:

You should avoid using a function evaluation in a for loop.
The compiler does not know that the value does not change and thus the function is called in every step of the loop.
It's way better for the performance to compute the function once, store the value in a variable and then use the varible in the for-loop.

src/t8_forest_vtk.h
```diff
@@ -30,41 +30,19 @@
...
 
-void
- 
-             t8_write_vtk_via_API (t8_forest_t forest, const char *filename);
+/** Write the forest in .pvtu file format. Writes one .vtu file per
+ * process and a meta .pvtu file.
+ * \param [in]  forest    The forest.
+ * \param [in]  fileprefix  The prefix of the output files.
+ */
+#if T8_WITH_VTK
```
holke, line 42:
```suggestion
```
By moving the macro in the cxx file inside the function, the function should be compilable and callable without linking vtk.
Hence, we need the declarion even without vtk.

src/t8_forest_vtk.h
```diff
@@ -30,41 +30,19 @@
...
+void                t8_write_vtk_via_API (t8_forest_t forest,
+                                          const char *filename);
+#endif
```
holke, line 45:
```suggestion
```

holke:
Looks very good. Thanks for the changes.
I will have a final look at it in the upcoming days.

holke (approved):

PR #128, opened 2021-06-22
Improve vtk configuration

holke:
- Fix spaces between '-l' libs.
- Disable T8_VTK_VERSION_USED if T8_WITH_VTK is not defined


holke:
Can be merged when the checks pass.

PR #131, opened 2021-06-28
Feature vtk api writing data

Veli-hub:
We are now able to write metadata into the vtu files written with the vtk write API function.

### DIFF
--- a/src/t8_forest/t8_forest_vtk.cxx
+++ b/src/t8_forest/t8_forest_vtk.cxx
@@ -49,6 +49,8 @@
 #include <vtkXMLPUnstructuredGridWriter.h>
 #include <vtkXMLUnstructuredGridWriter.h>
 #include <vtkUnsignedCharArray.h>
+#include <vtkDoubleArray.h>
+#include <vtkTypeInt64Array.h>
 #include <vtkSmartPointer.h>
 #if T8_ENABLE_MPI
 #include <vtkMPI.h>
@@ -130,7 +132,9 @@ void
 t8_forest_write_vtk_via_API (t8_forest_t forest, const char *fileprefix,
                              int write_treeid,
                              int write_mpirank,
-                             int write_level, int write_element_id)
+                             int write_level,
+                             int write_element_id,
+                             int num_data, t8_vtk_data_field_t * data)
 {
 #if T8_WITH_VTK
   /*Check assertions: forest and fileprefix are not NULL and forest is commited */
@@ -168,13 +172,13 @@ t8_forest_write_vtk_via_API (t8_forest_t forest, const char *fileprefix,
   num_elements = t8_forest_get_local_num_elements (forest);
   int                *cellTypes = T8_ALLOC (int, num_elements);
   /*
-   * We have to define the vtk UnvtkUnsignedCharArrays that hold 
+   * We have to define the vtkTypeInt64Array that hold 
    * metadata if wanted. 
    */
-  vtkUnsignedCharArray *vtk_treeid = vtkUnsignedCharArray::New ();
-  vtkUnsignedCharArray *vtk_mpirank = vtkUnsignedCharArray::New ();
-  vtkUnsignedCharArray *vtk_level = vtkUnsignedCharArray::New ();
-  vtkUnsignedCharArray *vtk_element_id = vtkUnsignedCharArray::New ();
+  vtkTypeInt64Array  *vtk_treeid = vtkTypeInt64Array::New ();
+  vtkTypeInt64Array  *vtk_mpirank = vtkTypeInt64Array::New ();
+  vtkTypeInt64Array  *vtk_level = vtkTypeInt64Array::New ();
+  vtkTypeInt64Array  *vtk_element_id = vtkTypeInt64Array::New ();
 
 /* We iterate over all local trees*/
   for (itree = 0; itree < t8_forest_get_num_local_trees (forest); itree++) {
@@ -271,7 +275,7 @@ t8_forest_write_vtk_via_API (t8_forest_t forest, const char *fileprefix,
         vtk_level->InsertNextValue (scheme->t8_element_level (element));
       }
       if (write_element_id == 1) {
-        vtk_element_id->InsertNextValue (elem_id + offset + (long long)
+        vtk_element_id->InsertNextValue (elem_id + offset +
                                          t8_forest_get_first_local_element_id
                                          (forest));
       /* *INDENT-ON* */
@@ -365,6 +369,19 @@ pwriterObj->SetEndPiece (forest->mpirank);
     unstructuredGrid->GetCellData()->AddArray(vtk_element_id);
   }
   /* *INDENT-ON* */
+
+/*vtkDataArray *array = vtkDoubleArray::New();[idata].data
+array->SetNumberOfComponents(b->ndims);
+array->SetVoidArray(data, b->ndims * b->dims[0], 1);
+
+vtkPoints *points = vtkPoints::New();
+points->SetData(array);
+array->Delete();
+
+vtkUnstructuredGrid *ugrid = vtkUnstructuredGrid::New();
+ugrid->SetPoints(points);
+points->Delete();*/
+
 pwriterObj->SetInputData (unstructuredGrid);
 pwriterObj->Update ();
 pwriterObj->Write ();
--- a/src/t8_forest/t8_forest_vtk.cxx
+++ b/src/t8_forest/t8_forest_vtk.cxx
@@ -28,6 +28,7 @@
 #include <t8_vec.h>
 #include "t8_cmesh/t8_cmesh_trees.h"
 #include "t8_forest_types.h"
+#if T8_WITH_VTK
 #include <vtkActor.h>
 #include <vtkCellArray.h>
 #include <vtkDataSetMapper.h>
@@ -52,6 +53,7 @@
 #include <vtkMPI.h>
 #include <vtkMPICommunicator.h>
 #include <vtkMPIController.h>
+#endif
 #endif
 #include <t8.h>
 #include <t8_forest.h>
@@ -122,10 +124,11 @@ typedef int         (*t8_forest_vtk_cell_data_kernel) (t8_forest_t forest,
                                                        void **data,
                                                        T8_VTK_KERNEL_MODUS
                                                        modus);
-#if T8_WITH_VTK
+
 void
 t8_write_vtk_via_API (t8_forest_t forest, const char *fileprefix)
 {
+#if T8_WITH_VTK
   /*Check assertions: forest and fileprefix are not NULL and forest is commited */
   T8_ASSERT (forest != NULL);
   T8_ASSERT (forest->rc.refcount > 0);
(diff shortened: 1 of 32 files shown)


src/t8_forest/t8_forest_vtk.cxx
```diff
@@ -160,8 +165,16 @@ t8_forest_write_vtk_via_API (t8_forest_t forest, const char *fileprefix)
...
+  num_elements = t8_forest_get_local_num_elements (forest);
+  int                *cellTypes = T8_ALLOC (int, num_elements);
+  /*
+   * We have to define the vtk UnvtkUnsignedCharArrays that hold 
+   * metadata if wanted. 
+   */
+  vtkUnsignedCharArray *vtk_treeid = vtkUnsignedCharArray::New ();
+  vtkUnsignedCharArray *vtk_mpirank = vtkUnsignedCharArray::New ();
+  vtkUnsignedCharArray *vtk_level = vtkUnsignedCharArray::New ();
+  vtkUnsignedCharArray *vtk_element_id = vtkUnsignedCharArray::New ();
```
holke, line 177: Do we have to use unsignedChar?
Or are there other options like Integer?

If it really uses `unsigned char` then
a) Our data types are signed and could (in theory) be negative. This would then cause weird underflow conversion results.
b) An unsigned char can only hold 255 different values and we certainly want to store larger numbers.

src/t8_forest/t8_forest_vtk.cxx
```diff
@@ -236,11 +250,35 @@ t8_forest_write_vtk_via_API (t8_forest_t forest, const char *fileprefix)
...
+      if (write_mpirank == 1) {
+        vtk_mpirank->InsertNextValue (forest->mpirank);
+      }
+      if (write_level == 1) {
+        vtk_level->InsertNextValue (scheme->t8_element_level (element));
+      }
+      if (write_element_id == 1) {
+        vtk_element_id->InsertNextValue (elem_id + offset + (long long)
+                                         t8_forest_get_first_local_element_id
+                                         (forest));
```
holke, line 276: Do we need the conversion to long long here? Why?

src/t8_forest/t8_forest_vtk.cxx
holke: Yes, i just confirmed it with VTK v9.0
It only stores values up to 255.

src/t8_forest_vtk.h
```diff
@@ -43,7 +43,12 @@ T8_EXTERN_C_BEGIN ();
  * \note If t8code was not configured with vtk, use \ref t8_forest_vtk_write_file
  */
 void                t8_forest_write_vtk_via_API (t8_forest_t forest,
-                                                 const char *fileprefix);
+                                          const char *filename,
+                                          int write_treeid,
+                                          int write_mpirank,
+                                          int write_level,
+                                          int write_element_id);
+
```
holke, line 51: This file needs to be indented.

holke (changes requested):
Looks very good.
The only thing that should definitely change is using more than unsigned char as data type.

src/t8_forest/t8_forest_vtk.cxx
Veli-hub: I can change it to vtkTypeInt64Array or others. I'll do vtkTypeInt64Array for now, others are: vtkIntArray:The C++ standard does not define the exact size of the int type, so use of this type directly is discouraged. If an array of 32 bit integers is needed, prefer vtkTypeInt32Array to this class.

src/t8_forest/t8_forest_vtk.cxx
Veli-hub: It should work fine without the conversion. I removed it.

src/t8_forest/t8_forest_vtk.cxx
holke: In t8_vtk.h we define datatypes for the manual writing and distinguish between locidx and gloidx.
We should do something similar and add defines like
```
#if T8_WITH_VTK
#define t8_vtk_locidx_array_type vtkTypeInt32Array
#define t8_vtk_gloidx_array_type vtkTypeInt64Array
#endif
```
and then we can use t8_vtk_locidx_array_type and t8_vtk_gloidx_array_type.
We may also add assertions to ensure no type conversion errors occur...

src/t8_forest/t8_forest_vtk.cxx
holke: So far, no update arrived here, may be you did not push?

src/t8_vtk.h
```diff
@@ -56,6 +56,11 @@
 #define T8_VTK_FORMAT_STRING "binary"
 #endif
 
+#if T8_WITH_VTK
+#define t8_vtk_locidx_array_type vtkTypeInt32Array
+#define t8_vtk_gloidx_array_type vtkTypeInt64Array
```
holke, line 61: Nice.

These should have a "_t" at the end as all t8code defined datatypes have.
```suggestion
#define t8_vtk_locidx_array_type_t vtkTypeInt32Array
#define t8_vtk_gloidx_array_type_t vtkTypeInt64Array
```

src/t8_forest/t8_forest_vtk.cxx
```diff
@@ -126,7 +129,12 @@ typedef int         (*t8_forest_vtk_cell_data_kernel) (t8_forest_t forest,
                                                        modus);
 
 void
```
holke, line 131: This function should return something. Similar to the old vtk write function.

Writing to a file can always go wrong and the return value of this function should reflect this.
I.e. return 1 if successful, 0 if not.
The write () routine of the vtk xml writer will give you the corresponding output.
Could this also go wrong before this call? If so, we need to return 0 then as well (and clean up the memory).

src/t8_forest_vtk.h
```diff
@@ -40,10 +40,21 @@ T8_EXTERN_C_BEGIN ();
  * "--with-vtk" in order to use it.
  * \param [in]  forest    The forest.
  * \param [in]  fileprefix The prefix of the output files. The meta file will be named \a fileprefix.pvtu .
```
holke, line 42: The write_treeid etc fields are not docuemented.

src/t8_forest/t8_forest_vtk.cxx
```diff
@@ -306,14 +358,67 @@ t8_forest_write_vtk_via_API (t8_forest_t forest, const char *fileprefix)
...
-  pwriterObj->SetInputData (unstructuredGrid);
-  pwriterObj->Update ();
-  pwriterObj->Write ();
-/* We have to free the allocated memory for the cellTypes Array. */
-  T8_FREE (cellTypes);
+pwriterObj->SetNumberOfPieces (forest->mpisize);
+pwriterObj->SetStartPiece (forest->mpirank);
+pwriterObj->SetEndPiece (forest->mpirank);
+  /* *INDENT-OFF* */
+  if (write_treeid == 1) {
```
holke, line 365: You can remove the ==1 part here, and in the other ifs.
This also has the benefit of letting the user provide any non-zero value (which we usually interpret as true)

```suggestion
  if (write_treeid) {
```

holke (changes requested):
Looks very good to me.
Only some minor changes are needed :+1:

src/t8_forest/t8_forest_vtk.cxx
```diff
@@ -365,6 +369,19 @@ pwriterObj->SetEndPiece (forest->mpirank);
...
+array->SetVoidArray(data, b->ndims * b->dims[0], 1);
+
+vtkPoints *points = vtkPoints::New();
+points->SetData(array);
+array->Delete();
+
+vtkUnstructuredGrid *ugrid = vtkUnstructuredGrid::New();
+ugrid->SetPoints(points);
+points->Delete();*/
+
```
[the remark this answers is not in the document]
Veli-hub: Yes, this can be removed.

src/t8_forest/t8_forest_vtk.cxx
Veli-hub: I am not sure what else we should check. The other vtk functions do not return any information about whether the task was successful or not.

holke: Ok, then we should stick to write() and return 1 if it was successful and 0 otherwise

src/t8_forest/t8_forest_vtk.cxx
```diff
@@ -404,7 +406,10 @@ for (int idata = 0; idata < num_data; idata++) {
 /* We set the input data and write the vtu files. */
 pwriterObj->SetInputData (unstructuredGrid);
 pwriterObj->Update ();
-pwriterObj->Write ();
+if (pwriterObj->Write () == 0) {
+  /* Writing failed */
+  goto t8_forest_vtk_failure;
```
holke, line 411: The clean up is the same even if writing was successful, so the code is currently just duplicated, which we should avoid.

What you could do is store 1 or 0 in a variable depending on the result here, then do the clean up and return the variable.

src/t8_forest/t8_forest_vtk.cxx
```diff
@@ -418,12 +423,26 @@ for (int idata = 0; idata < num_data; idata++) {
...
+
+t8_forest_vtk_failure:
+t8_errorf ("Error when writing vtk file.\n");
+vtk_treeid->Delete ();
+vtk_mpirank->Delete ();
+vtk_level->Delete ();
+vtk_element_id->Delete ();
+for (int idata = 0; idata < num_data; idata++) {
+  dataArrays[idata]->Delete ();
+}
```
holke, line 437: See comment above, remove the duplicated code.
The error print is good and should be triggered if the return value of write was 0.

src/t8_forest_vtk.h
```diff
@@ -40,15 +40,20 @@ T8_EXTERN_C_BEGIN ();
...
+ * \param [in]  write_treeid If true, the global tree id is written for each element.
+ * \param [in]  write_mpirank If true, the mpirank is written for each element .
+ * \param [in]  write_level If true, the refinement level is written for each element.
+ * \param [in]  write_element_id If true, the global element id is written for each element.
  * \param [in]  num_data  Number of user defined double valued data fields to write.
  * \param [in]  data      Array of t8_vtk_data_field_t of length \a num_data
  *                        providing the used defined per element data.
  *                        If scalar and vector fields are used, all scalar fields
  *                        must come first in the array.
+ * \return  True if succesful, false if not (process local).
```
holke, line 52:
```suggestion
 * \return  True if successful, false if not (process local).
```

holke (changes requested):
Looks good, thanks for the changes.

Only some minor changes with the return value handling and we can merge

src/t8_forest/t8_forest_vtk.cxx
```diff
@@ -306,19 +361,81 @@ t8_forest_write_vtk_via_API (t8_forest_t forest, const char *fileprefix)
...
+
+T8_FREE (cellTypes);
+T8_FREE (dataArrays);
+/* Return wether writing was successful */
```
holke, line 430:
```suggestion
/* Return whether writing was successful */
```

holke:
All is looking fine :)

PR #133, opened 2021-07-02
fix forest statistics bug

holke:
- Correct number of stats and stats collection
- Fix stats computation bug in balance

    When more balance rounds are required than stats allocated,
    the stats array grows dynamically.
    However, we did not take into account that we do less partition
    and ghost runs than adapt. (Since the last adapt does nothing,
    so no ghost/partition needed).
    This resulted in use of uninitialized values when more balances
    rounds where needed.
    We fixed this behaviour.
    - Different stat counters for adapt,ghost,partition
    - Introduced stat chunk size for dynamic growth
    - Increase chunk size from 5 to 10

### DIFF
--- a/src/t8_forest/t8_forest.c
+++ b/src/t8_forest/t8_forest.c
@@ -1189,7 +1189,7 @@ t8_forest_print_profile (t8_forest_t forest)
     sc_stats_set1 (&stats[10], profile->ghost_runtime,
                    "forest: Ghost runtime.");
     sc_stats_set1 (&stats[11], profile->ghost_waittime,
-                   "forest: Ghost runtime.");
+                   "forest: Ghost waittime.");
     sc_stats_set1 (&stats[12], profile->balance_runtime,
                    "forest: Balance runtime.");
     sc_stats_set1 (&stats[13], profile->balance_rounds,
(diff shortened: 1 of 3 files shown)


src/t8_forest/t8_forest.c
```diff
@@ -1188,9 +1188,11 @@ t8_forest_print_profile (t8_forest_t forest)
                    "forest: Commit runtime.");
     sc_stats_set1 (&stats[10], profile->ghost_runtime,
                    "forest: Ghost runtime.");
-    sc_stats_set1 (&stats[11], profile->balance_runtime,
+    sc_stats_set1 (&stats[11], profile->ghost_waittime,
+                   "forest: Ghost runtime.");
```
holke, line 1192:
```suggestion
                   "forest: Ghost waittime.");
```

src/t8_forest/t8_forest_balance.cxx
```diff
@@ -144,7 +144,13 @@ t8_forest_balance (t8_forest_t forest, int repartition)
 {
   t8_forest_t         forest_temp, forest_from, forest_partition;
   int                 done = 0, done_global = 0;
-  int                 count = 0, num_stats, i;
+  int                 count_rounds = 0;
+  /* The following variables are only required if profiling is
+   * enabled. */
+  int                 num_stats_allocated, i;
```
holke, line 150: rename i to istats or smth

Issue #152, opened 2021-10-09
A few suggestions regarding the installation


jackerschott:
I have a few notes and suggestions regarding the installation of t8code on linux, as a few directories and files get installed into somewhat odd locations. The following is with `--prefix=/usr` in mind.

First the t8code headers (excluding the p4est and libsc ones) files get not installed into `/usr/include`, but rather into `/usr/include/src`, which is probably not intended, unnecessary and likely confusing for the end user. Note than one would have to either define `/usr/include/src` as an additional include directory or include t8code files as i.e. `#include <src/t8.h>`.
The source of this problem is that the header files are contained within the `src` subdirectory and are installed with the `nobase_` prefix. The easiest fix here would be to just remove this prefix, however that would also remove the `t8_forest`, `t8_cmesh`, `t8_data`, etc. subdirectories. Another way would be to define extra include directories for each of these subfolders.

Second there are `data` and `ini` directories which are installed directly into `/usr/share/`. As far as i know these directories are not considered standard (correct me if I am mistaken here) and therefore clutter `/usr/share` a bit. I would suggest putting these into `/usr/share/libsc/data`, `/usr/share/p4est/data`, `/usr/share/t8code/data` and `/usr/share/libsc/ini`.

Third you install `ax_prefix_config_h.m4`, `ax_split_version.m4` into `/usr/share/aclocal`, which are not t8code, p4est or libsc specific files but belong to GNU autoconf-archive. Therefore they raise conflicts with the respective package on probably most linux distributions (at least on Archlinux and Debian). This could be fixed by simply not installing these files and mention autoconf-archive as a dependency of t8code.

Fourth you install `Makefile.t8.mk`, `Makefile.p4est.mk`, `Makefile.sc.mk` directly into `/usr/etc` (or `/etc` on a FHS compliant system) which also clutters `/etc` similiar to the second point. The solution here could be the same: One could install these files into `/etc/t8code`, `/etc/p4est` and `/etc/libsc` subdirectories.

And last but not least, you could maybe consider mentioning libsc and p4est as dependencies rather than installing them through git submodules. I know that this is more of a conceptual problem than a practical one and it would probably render the installation of t8code by source a little bit more difficult. The last point could be maybe avoided though by making this optional. I mention this point because imagine it opposes the concept by which programs are packaged on probably any linux distribution.

[...]

PR #153, opened 2021-10-12
Fix installation location of header files

jackerschott:
Install header files in `<prefix>/include` instead of `<prefix>/include/src`. See #152.


holke:
Closed because it moved to #248

PR #158, opened 2021-11-17
Feature init shared mem

holke:
MPI shared memory needs to be initialized by setting inter and intranode communicators before using.
Added new t8_shmem_init and t8_shmem_finalize functions to t8_shmem.h and used them whenever shared memory is needed (cmesh_partition).

Adding warnings when used without initializing.


holke:
Converted to draft because i notice some bugs.
Currently working on unit tests.

holke:
I am now done.

The shmem array needed some reworking to properly function.

One has to keep in mind that for shared memory as libsc manages it, only those processes are allowed to write into it, that
sc_shmem_start_writing returns true for. For shmem types window and bgq this is only one process per shared memory region.
sc_shmem_start_writing and sc_shmem_end_writing must be called before by all processes.

We implemented this into t8_shmem_array with the new t8_shmem_array_start_writing and end_writing functions.

Davknapp (approved):

PR #175, opened 2022-01-26
Feature adapt with balance

holke:

Implemented the adapt with balance feature.
If a forest should be adapted and balanced, then these operations are now performened in a single step.

To this end, the adapt function will create a marker array of elements to refine, then the balance condition according to this
marker array will be checked and the markes modified, such that the resulting adapted mesh will be balanced.
At the end the adaptation is carried out according to the marker.

New function `t8_forest_set_balance_ext` can explicitely disable this behaviour, such that adapt and balance will be separate steps.

The `t8_test_forest_commit` test should test both variants.


holke:
Note: I coded this 1,5 years ago and did not look over the code again now.
But i believe it should be in a state where we can integrate it.

holke:
Well the tests fail.
But looks like minor problems (warnings).
Will update soon.

holke:
Converted to draft since tests still fail.

Davknapp:
Do not delete this branch, we will continue development!

PR #176, opened 2022-01-26
Feature googletest

lukasdreyer:
Enable googletest MPI for t8code. This includes changes in the Makefile, so that googletest is build with the correct flags, depending on whether it is build with or without MPI. 

There is an experimental feature, that allows to use `make gtest` for direct output of each test in the gtest suite, instead of having to call `make check` and looking at `test-suite.log` for that output. This currently displays the output of the last gtest run, if there were any compiler problems during the current build. Thus, this feature should be used with caution.

We use the current version of [googletest-mpi](https://github.com/DLR-SC/googletest_mpi/tree/feature/mpi_nonblocking_expect) that supports nonblocking expect calls. With this [script](https://github.com/DLR-SC/googletest_mpi/blob/feature/mpi_nonblocking_expect/googletest/scripts/fuse_gtest_files.py), all source files are fused into one source file, that is included in `/thirdparty/googletest-mpi` in t8code.

This branch also includes three gtest files, where one is currently just an assembly of things that should be tested in the future.

Additionally, some spelling and spacing is fixed in this branch.


test/t8_test_cmesh_partition.cxx
```diff
@@ -24,6 +24,7 @@
 #include <t8_schemes/t8_default_cxx.hxx>
 #include "t8_cmesh/t8_cmesh_trees.h"
 #include "t8_cmesh/t8_cmesh_partition.h"
+#include "src/t8_refcount.h"
```
Davknapp, line 27:
```suggestion
```

Do we need t8_refcount here? We do not use any functions of t8_refcount in this test

Davknapp (changes requested):
Looks good to me, I will approve the pull request after you have addressed my suggestions regarding test/t8_test_cmesh_partition.cxx

test/t8_test_cmesh_partition.cxx
lukasdreyer: I asked @holke the same thing in a private message, as this line also lead to a merge conflict with the line below. From my point, we could delete this line.
His answer was to include it in this merge, as it could lead to other merge conflicts in the future. Maybe we can discuss that together tomorrow.

Davknapp (approved):

PR #191, opened 2022-03-24
Feature write netcdf files

Niklas997:
Mainly, the netCDF examples were extended and documented.


lukasdreyer:
The build check with "-Werror" complains about some "defined but not used" functions

config/t8_netcdf.m4
```diff
@@ -28,11 +28,12 @@ if test "x$T8_WITH_NETCDF" != xno ; then
...
 [[
-]],[[
   #include <netcdf.h>
-
+  #include <netcdf_par.h>
+  #include <mpi.h>
+]],[[
   int ncid;
-  nc_open (NULL, NC_NOWRITE, &ncid);
+  nc_open_par (NULL, NC_NOWRITE, MPI_COMM_SELF, MPI_INFO_NULL, &ncid);
```
lukasdreyer, line 36: Does this mean that we always need a parallel build of netcdf? Is there a use case where only a serial build of netcdf is available?

example/netcdf/t8_write_cmesh_netcdf.c
```diff
@@ -0,0 +1,162 @@
...
+  /* Receive the process-local MPI rank */
+  mpiret = sc_MPI_Comm_rank (comm, &mpirank);
+  SC_CHECK_MPI (mpiret);
+
+  /* Construct a (partitioned) cmesh */
+  cmesh = t8_cmesh_new_hypercube (T8_ECLASS_TET, comm, 0, 1, 0);
+
+  /* Number of process local elements */
+  printf ("[t8] Number of local trees on process %d : %d\n", mpirank,
+          t8_cmesh_get_num_local_trees (cmesh));
```
holke, line 67: Use one of the t8code logging functions instead of printf.

example/netcdf/t8_write_cmesh_netcdf.c
```diff
@@ -0,0 +1,162 @@
...
+
+  for (j = 0; j < t8_cmesh_get_num_local_trees (cmesh); j++) {
+    random_values[j] = rand () / (double) rand ();
+  }
+  /* Create a new sc_array_t which provides the data for the NetCDF variables, in this case just random values */
+  var_random_values =
+    sc_array_new_data (random_values, sizeof (double),
+                       t8_cmesh_get_num_local_trees (cmesh));
+
+  /* Create the integer NetCDF variable; parameters are (name of the variable, descriptive long name of the variable, description of the data's unit (i.e. degrees Celsius), pointer to sc_array_t which provides the data) */
```
lukasdreyer, line 101:
```suggestion
  /* Create the double NetCDF variable; parameters are (name of the variable, descriptive long name of the variable, description of the data's unit (i.e. degrees Celsius), pointer to sc_array_t which provides the data) */
```

src/t8_netcdf.h
```diff
@@ -0,0 +1,84 @@
...
+/* Datatype for 32 bit NetCDF Data */
+typedef int32_t     t8_nc_int32_t;
+
+/* Struct for elementwise data variable for a NetCDF file */
+typedef struct
+{
+  const char         *variable_name;
+  const char         *variable_long_name;
+  const char         *variable_units;
+  int                 datatype;
```
lukasdreyer, line 44: I think an enum would be more clear here. If you stay with int, there needs to be at least an explanation that 0=int, 1=double

example/netcdf/t8_write_forest_netcdf.cxx
```diff
@@ -0,0 +1,503 @@
...
+#include <t8_cmesh.h>
+#include <t8_forest.h>
+#include <t8_forest/t8_forest_iterate.h>
+#include <t8_schemes/t8_default_cxx.hxx>
+#include <t8_forest_netcdf.h>
+#include <t8_netcdf.h>
+
+/* In this example is the use of the netcdf feature exemplary displayed
+* We show how to write out a forest in the netCDF format and how to create additional (integer/double) netCDF variables which hold element data.
+* There are mainly to funcitons implemented in 'src/t8_forest_netcdf.cxx' which allows the creation of netCDF file containing the data for a forest in the style of the URID conventions.
```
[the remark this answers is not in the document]
holke: Isnt it UGRID instead of URID?

holke:
Is currently halted bc test cases with cmesh fail.
cmesh hybrid is tested partitioned, but this configuration is not supported.

lukasdreyer:
Now the c++ compiler does not want to compile two of the c files.
It seems to have a problem with the following function declaration, although I can't figure out what the problem is.
```
static void
t8_cmesh_init_ugrid_namespace_context (t8_cmesh_netcdf_ugrid_namespace_t *
                                       namespace, int dim)
```
Additionally,
`ISO C++ forbids converting a string constant to ‘char*’ `. This problem could perhaps be fixed by changing `char*` to `const char*`

config/t8_netcdf.m4
lukasdreyer: I think that this is the problem why the check action fails now. The netcdf version from the repository only has `netcdf.h` and `nc_open`, but not `netcdf_par.h` and `nc_open_par`

src/t8_netcdf.h
```diff
@@ -0,0 +1,110 @@
...
+ * \param [in]  var_long_name    A string describing the variable a bit more and what it is about.
+ * \param [in]  var_unit    The units in which the data is provided.
+ * \param [in]  var_data    A sc_array_t holding the elementwise data of the variable.
+ * \param [in]  num_extern_netcdf_vars    The number of extern user-defined variables which hold elementwise data (if none, set it to 0).
+ */
+t8_netcdf_variable_t *t8_netcdf_create_var (t8_netcdf_variable_type_t
+                                            var_type, const char *var_name,
+                                            const char *var_long_name,
+                                            const char *var_unit,
+                                            sc_array_t * var_data);
```
lukasdreyer, line 75: Should this internal function have an interface to the outside?

src/t8_forest_netcdf.h
```diff
@@ -0,0 +1,78 @@
...
+ * \param [in]  forest    A forest.
+ * \param [in]  file_prefix    A string which holds the file's name (output file will be 'file_prefix.nc').
+ * \param [in]  file_title    A string to caption the NetCDF-File.
+ * \param [in]  dim    The Dimension of the forest mesh (2D or 3D).
+ * \param [in]  num_extern_netcdf_vars    The number of extern user-defined variables which hold elementwise data (if none, set it to 0).
+ * \param [in]  ext_variables An array of pointers of the herein before mentioned user-defined variables (if none, set it to NULL).
+ * \param [in]  comm The sc_MPI_Communicator to use.
+ * \param [in]  netcdf_var_storage_mode This defines whether the variables in the netCDF-4 File are stored contiguously (NC_CONTIGUOUS) or in chunks (NC_CHUNKED (in this case netCDF's default chunking scheme is used)).
+ * \param [in]  netcdf_var_mpi_access This defines whether the mpi processes are writing the variables' data collectively or independently.
+ * \note Only use this extended version of \ref t8_forest_write_netcdf if you know exactly what you are doing. In all other cases, we recommend using \ref t8_forest_write_netcdf instead.
```
lukasdreyer, line 45: If the user should normally use t8_forest_write_netcdf, I would prefer to have that function on top, so that it is found first when looking in the header file

src/t8_forest_netcdf.h
```diff
@@ -0,0 +1,78 @@
...
+ */
+void                t8_forest_write_netcdf_ext (t8_forest_t forest,
+                                                const char *file_prefix,
+                                                const char *file_title,
+                                                int dim,
+                                                int num_extern_netcdf_vars,
+                                                t8_netcdf_variable_t *
+                                                ext_variables[],
```
lukasdreyer, line 53: My C Knowledge is extremely rusty, what is the difference between using 
`t8_netcdf_variable_t * ext_variables[] ` and `t8_netcdf_variable_t ** ext_variables` here?

src/Makefile.am
```diff
@@ -25,6 +25,10 @@ libt8_installed_headers = \
   src/t8_geometry/t8_geometry_implementations/t8_geometry_linear.hxx \
   src/t8_geometry/t8_geometry_implementations/t8_geometry_zero.hxx \
   src/t8_vec.h src/t8_vtk.h \
+  src/t8_cmesh_netcdf.h \
+  src/t8_forest_netcdf.h \
+  src/t8_element_shape.h \
+  src/t8_netcdf.h \
```
lukasdreyer, line 31: @holke: What is the criterion for a header file to be an installed header? Do all of these headers need to be installed?

holke: Installed headers headers that should be seen from the outside.
So if it is necessary for user code (e.g. a simulation application) to include this header, then it should be installed.
If the header is only used internally in t8code (and does not contain declarations of function for a user) then it should not be installed but internal.

src/t8_netcdf.c
```diff
@@ -0,0 +1,87 @@
...
+
+/* Create an extern NetCDF integer variable */
+t8_netcdf_variable_t *
+t8_netcdf_create_integer_var (const char *var_name, const char *var_long_name,
+                              const char *var_unit, sc_array_t * var_data)
+{
+#if T8_WITH_NETCDF
+  t8_netcdf_variable_type_t var_type;
+  /* Check wheter 32-bit (4-byte) integer or 64-bit integer data sholud be written */
+  var_type = (var_data->elem_size > 4) ? T8_NETCDF_INT64 : T8_NETCDF_INT;
```
lukasdreyer, line 56: Are those the only 2 cases that can happen?

src/t8_netcdf.c
```diff
@@ -0,0 +1,87 @@
...
+#include <t8_netcdf.h>
+
+/* Create an extern NetCDF variable */
+t8_netcdf_variable_t *
+t8_netcdf_create_var (t8_netcdf_variable_type_t var_type,
+                      const char *var_name, const char *var_long_name,
+                      const char *var_unit, sc_array_t * var_data)
+{
+#if T8_WITH_NETCDF
+  T8_ASSERT (0 <= var_type && var_type <= 2);
```
lukasdreyer, line 33: This is probably also caught somewhere else, but maybe you could check here if the `elem_size` of `var_data` fits to the `var_type` given

example/netcdf/t8_write_cmesh_netcdf.c
```diff
@@ -0,0 +1,162 @@
...
+  t8_netcdf_variable_t *ext_var_random_values;
+  int                 mpiret;
+  int                 mpirank;
+  int                 j;
+
+  /* Receive the process-local MPI rank */
+  mpiret = sc_MPI_Comm_rank (comm, &mpirank);
+  SC_CHECK_MPI (mpiret);
+
+  /* Construct a hybrid cmesh (not partitioned) */
```
lukasdreyer, line 61: This is no longer true

src/t8_netcdf.h
Niklas997: I would rather see this as the standard function to use in order to create a netCDF variable, although there are already specialized functions for integer and double variables. I also added a note in the examples, that this general function exists and probably is more convenient since the datatype has to be explicitly stated (unlike in the t8_netcdf_create_integer_var function).

src/t8_forest_netcdf.h
Niklas997: The only difference is the syntax, I guess. In both cases it should resolve to `t8_netcdf_variable_t ** ext_variables`.

src/t8_forest/t8_forest_netcdf.cxx
```diff
@@ -0,0 +1,1188 @@
...
+
+/*
+Description:
+These functions write a file in the NetCDF-format which represents the given 2D- or 3D-forest
+*/
+
+#include <t8.h>
+#if T8_WITH_NETCDF
+#include <netcdf.h>
+/* do we need to question wheter this header is accessible? */
```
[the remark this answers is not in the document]
Niklas997: A configuration macro was added (`T8_WITH_NETCDF_PAR`) which automatically checks for the accessibility of the parallel netCDF routines (but only in case t8code is configured with the option `--with-netcdf`, otherwise it is `0`). (To force a serial netCDF usage, although the parallel routines are accessible, one can use `--with-netcdf=serial`)

src/t8_netcdf.c
Niklas997: Currently, yes. These are the only cases which we have implemented so far. But this evaluation has to be adapted if for example an option for `NC_SHORT` variables (signed 2-byte integer) is integrated.

src/t8_element_shape.h
```diff
@@ -0,0 +1,87 @@
...
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with t8code; if not, write to the Free Software Foundation, Inc.,
+  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+*/
+
+/** \file t8_element_shape.h
+ * TODO: comment
```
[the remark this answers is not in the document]
Niklas997: @holke Why did we decided to use `t8_element_shape_t` instead of `t8_eclass_t`?

example/netcdf/t8_write_forest_netcdf.cxx
```diff
@@ -224,12 +228,14 @@ t8_example_compare_performance_netcdf_var_properties (sc_MPI_Comm comm,
...
   }
+  num_elements = t8_forest_get_local_num_elements (forest);
+  printf ("[rank %d] Num elements: %ld\n", mpirank, num_elements);
 
   /* If additional data should be written to the netCDF file, the two variables are created in the following section */
   if (with_additional_data) {
     /* Get the number of process-local elements */
     num_elements = t8_forest_get_local_num_elements (forest);
-
+    printf ("[rank %d] Num elements: %ld\n", mpirank, num_elements);
```
lukasdreyer, line 238: Use t8code printing function (also in line 232)

src/t8_forest_netcdf.h
lukasdreyer: @holke : Can you expand on that? Do we have a coding guideline for which variant we prefer?

src/t8_forest/t8_forest_netcdf.cxx
```diff
@@ -0,0 +1,1244 @@
...
+  case 3:
+    t8_debugf ("Writing 3D forest to NetCDF.\n");
+    t8_forest_write_netcdf_file (forest, &context, &namespace_context,
+                                 num_extern_netcdf_vars, ext_variables, comm);
+    break;
+  default:
+    t8_global_errorf
+      ("Only writing 2D and 3D NetCDF forest data is supported.\n");
+    break;
+  }
```
lukasdreyer, line 1222: Same here as for the `cmesh_` function, no difference for different dimensions

lukasdreyer (commented):
I have not extensively checked all of the `t8_{cmesh|forest}_netcdf_write_*` functions, as I do not have the expertise in netcdf needed for that. I assume that you and @holke discussed that functionality in detail.

Other than a few last suggestions and some open questions to @holke in other threads, I am happy with this PR.

src/t8_element_shape.h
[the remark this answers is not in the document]
holke: eclass is for trees only and shape is for elements.
We started to introduce this with pyramids when we first got elements of a different shape than their tree (tets as elements in a pyramid tree).
Also, eclass is limited to the supported tree types (hex, tet, prism pyra in 3D), but we could think of a refinement scheme that generates complety new shapes (like a Hex that refines into 17 Hexes, 12 Docehedra and 13 Rhombicosidodecahedra). Adding these as element shapes should be possible without needing to add them as tree shapes (which would be more complex).

src/t8_forest_netcdf.h
holke: should be syntax only, i would prefer `**`

lukasdreyer (approved):
There are currently no tests, but a [discussion](https://github.com/holke/t8code/discussions/298) is ongoing on how we want to test output in general, so I will approve this PR and a test should be added once that discussion is concluded.

PR #196, opened 2022-04-01
Update gtest output

Davknapp:
grep -c WORD TXT only counts the lines where WORD occurs in TXT

With grep -o each occurance of WORD in TEXT is put out into a single line. We then count the number of lines.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -132,7 +132,7 @@ gtest:
 	@echo -e "\e[32m============================================================================\e[0m"
 	@echo "# Total:"; grep -c "RUN" test/t8_gtest_main.log
 	@echo -e "\e[32m# PASS:"
-	@grep -c "OK" test/t8_gtest_main.log
+	@grep -o -i OK test/t8_gtest_main.log | wc -l
 	@echo -e "\e[32m============================================================================\e[0m"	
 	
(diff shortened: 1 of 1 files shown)


holke:
sidenode: looks like a good change.

For a next improvement: We should not use "echo" and "grep" directly, since they may not be available on all systems or have different names. The configure script checks for this functionality and the exact names of the functions to call and should store them in internal variables (something like $GREP or similar) which we should use in the makefile.

lukasdreyer:
Could it happen that the output of two tests overlaps in a way that splits `OK` into `O Output of other test K`? I don't know how to account for that, and this solution is already better than the one we had before.

Davknapp:
> > Could it happen that the output of two tests overlaps in a way that splits `OK` into `O Output of other test K`? I don't know how to account for that, and this solution is already better than the one we had before.
> 
> Yes, that could happen. I don't know either how to account for that. But in that case, the user would still see, that all tests have passed, but the number of passed tests won't match the total number of tests.

we could use 
$(info text) to avoid the usage of echo

lukasdreyer (approved):
Other than the explicit use of grep and echo, which will be adressed separately, this looks good.

PR #197, opened 2022-04-04
Create mattermost.yml

Davknapp:
Created a workflow that sends a Message to our Mattermost-github channel everytime something is pushed to the main branch or a pr to the mainbranch is created.

  - [x] If not, does the specific use case compile and all tests pass (check manually)

### DIFF
--- a/.github/workflows/mattermost.yml
+++ b/.github/workflows/mattermost.yml
@@ -53,7 +53,7 @@ jobs:
         with:
           username: t8y
           url: ${{ secrets.MATTERMOST_WEBHOOK_URL }}
-          text: Pull request ${{ github.event.number }} has been ${{ github.event.action }}. See ${{ github.event.pull_request.url }} for more details.
+          text: Pull request ${{ github.event.number }} has been ${{ github.event.action }}. See https://github.com/holke/t8code/pull/${{ github.event.number }} for more details.
 # trigger this part of the workflow, when something new is pushed into a pull request.
       - name: synchronize
         uses: Mitigram/gh-action-mattermost@main
@@ -61,4 +61,4 @@ jobs:
         with:
           username: t8y
           url: ${{ secrets.MATTERMOST_WEBHOOK_URL }}
-          text: Updated Pull-Request ${{ github.event.number }}. See ${{ github.event.pull_request.url }} for more details.
+          text: Updated Pull-Request ${{ github.event.number }}. See https://github.com/holke/t8code/pull/${{ github.event.number }} for more details.
--- a/.github/workflows/mattermost.yml
+++ b/.github/workflows/mattermost.yml
@@ -1,0 +1,26 @@
+# This is a basic workflow to help you get started with Actions
+
+name: Mattermost_message
+
+# Controls when the workflow will run
+on:
+  # Triggers the workflow on push or pull request events but only for the develop branch
+  push:
+    branches: [ feature-CI_mattermost_messages, develop, main ]
+  pull_request:
+    branches: [ feature-CI_mattermost_messages, develop, main ]
+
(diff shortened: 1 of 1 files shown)


.github/workflows/mattermost.yml
```diff
@@ -0,0 +1,41 @@
+# This is a basic workflow to help you get started with Actions
```
holke, line 1: Please add the t8code licence statement.

holke, line 1:
```suggestion
# This is a basic workflow to help you get started with Actions
```
This comment is not valid.

Please add a description what this action does and if it has inputs, what the inputs are etc.

.github/workflows/mattermost.yml
```diff
@@ -0,0 +1,41 @@
...
+
+name: Mattermost_message
+
+# Controls when the workflow will run
+on:
+  # Triggers the workflow on push or pull request events but only for the develop branch
```
holke, line 7: The comment is not correct.

.github/workflows/mattermost.yml
```diff
@@ -0,0 +1,41 @@
...
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: push  
+        uses: Mitigram/gh-action-mattermost@main
+        if: github.event_name == 'push'
+        with: 
+          username: t8y
+          url: ${{ secrets.MATTERMOST_WEBHOOK_URL }}
+          text: ${{ github.event.pusher.name }} pushed into ${{ github.event.ref }}. See ${{ github.event.compare }} for more details.
```
holke, line 27: imho it is too much to trigger this for each push.
For a typical code review process this would spam our Chat.

.github/workflows/mattermost.yml
```diff
@@ -0,0 +1,41 @@
...
+      - name: pull_request
+        uses: Mitigram/gh-action-mattermost@main
+        if: github.event_name == 'pull_request' && github.event.action != 'synchronize'
+        with:
+          username: t8y
+          url: ${{ secrets.MATTERMOST_WEBHOOK_URL }}
+          text: Pull request ${{ github.event.number }} has been ${{ github.event.action }}.
```
holke, line 34: If we can we should post the link of the pr.

.github/workflows/mattermost.yml
```diff
@@ -0,0 +1,41 @@
...
+      - name: synchronize
```
holke, line 35: what does synchronize do?
Could you please explain to me and add a comment?

holke (changes requested):
I love this addition to t8code.
Thanks for taking care of it.

Some fine tuning remains.

.github/workflows/mattermost.yml
Davknapp: Thanks for the reminder!

.github/workflows/mattermost.yml
Davknapp: We can restrict it to certain actions. All actions possible are listed here: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request
I think, that opened, reopend, assigned and edited are probably the most important ones. We can restrict it to these cases.

holke: ah, wait.
This will later only fire when pushed into develop or main.
Not if we have a PR into develop and smth is pushed to the PR.

I guess then its fine.

.github/workflows/mattermost.yml
holke: ah, so this is what i thought the push above would do.
Maybe we should just see how it is in practice.

holke (approved):

holke:
Well merge it and see in iterate it when we see the perfomance in production.

PR #223, opened 2022-04-25
Feature nca test

Davknapp:
Currently, there is no test to check the computation of the nearest common ancestor for each element.
This PR implements such a function. Testing for pyramids is skipped in general. Two tests are implemented.

Test 1: nca_check_shallow
check the nca for all combination of children of the root-element

Test 2: nca_check_deep
Compute elements up to level 10 and check from the current level on the nca for the first and last descendant of this element. The first and last descendant must have this active element as the nca, This test is different for T8_ECLASS_VERTEX as the first/last descendant logic is not valid for vertices. Instead, the nca is checked for the correct level.

### DIFF
--- a/test/t8_gtest_nca.cxx
+++ b/test/t8_gtest_nca.cxx
@@ -1,0 +1,91 @@
+/*
+This file is part of t8code.
+t8code is a C library to manage a collection (a forest) of multiple
+connected adaptive space-trees of general element classes in parallel.
+
+Copyright (C) 2015 the developers
+
+t8code is free software; you can redistribute it and/or modify
+it under the terms of the GNU General Public License as published by
+the Free Software Foundation; either version 2 of the License, or
+(at your option) any later version.
+
+t8code is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with t8code; if not, write to the Free Software Foundation, Inc.,
+51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+*/
+
+/** \file t8_gtest_nca.cxx
+* Provide tests to check the functionality of the nearest-common-ancestor function
+* for every element.
+*/
+
+#include <gtest/gtest.h>
+#include <t8_eclass.h>
+#include <t8_schemes/t8_default_cxx.hxx>
+
+/* *INDENT-OFF* */
+class nca:public testing::TestWithParam < t8_eclass > {
+protected:
+    void SetUp () override {
+        eclass = GetParam ();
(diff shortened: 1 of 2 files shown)


test/t8_gtest_nca.cxx
```diff
@@ -0,0 +1,150 @@
...
+            /* Compute first and last descendant at every level up to elem_max_lvl. 
+                *They have the correct_nca as the nca*/
+            for(check_lvl = lvl + 1; check_lvl < elem_max_level; check_lvl++){
+                ts->t8_element_first_descendant(correct_nca, desc_a, check_lvl);
+                ts->t8_element_last_descendant(correct_nca, desc_b, check_lvl);
+                /* Compute the nca of desc_a and desc_b*/
+                ts->t8_element_nca(desc_a, desc_b, check);
+                if(eclass == T8_ECLASS_VERTEX){
+                    /* first- last-descendant logic does not hold for vertices.*/
+                    EXPECT_EQ(ts->t8_element_level(check), SC_MIN(ts->t8_element_level(desc_a), ts->t8_element_level(desc_b)));
```
lukasdreyer, line 126: Since both first and last descendant are called with the same check_lvl, is the following equivalent?
```suggestion
                    EXPECT_TRUE ((ts->t8_element_compare (desc_a, check) == 0));
                    EXPECT_TRUE ((ts->t8_element_compare (desc_b, check) == 0));
```

test/t8_gtest_nca.cxx
```diff
@@ -0,0 +1,150 @@
...
+        else{
+            ts->t8_element_copy(correct_nca, tmp);
+        }
+    }
+    ts->t8_element_destroy(1, &tmp);
+}
+
+
+INSTANTIATE_TEST_SUITE_P (t8_gtest_nca, nca,
+                        testing::Range (T8_ECLASS_ZERO, T8_ECLASS_COUNT));
```
lukasdreyer, line 149: You could also exclude Pyramids by adjusting the range here, but I think your solution is fine.
```suggestion
                        testing::Range (T8_ECLASS_ZERO, T8_ECLASS_PYRAMID));
```

lukasdreyer (commented):
It could be interesting to have an additional test, where both elements are not on the same level.
But I would be fine to address that in a future PR, as both of these tests look good.

test/t8_gtest_nca.cxx
Davknapp: Yes, I've done that first, but with the current setup you see that test is skipped for pyramids without having a detailed look into the file.

test/t8_gtest_nca.cxx
Davknapp: Yes, but this will not hold anymore, if we test on different levels, as you suggest. I would prefere to keep it that way, cause it probably will be changed back to this as soon as this test is adapted to test for different level

lukasdreyer (approved):

Davknapp:
Please let the branch active, I will develop the improvement of the test on it.

PR #225, opened 2022-04-29
Is_valid check for all prism-functions

### LINKED ISSUE #218: Prism elements are missing an is_valid check
All element types should have an `t8_element_is_valid` check and assert in each function of `t8_default_X_cxx.cxx` whether the input elements are valid.
The prisms are missing this.

Are any other elements missing it?

Davknapp:
Added is_valid checks for all prism-functions. 
Furthermore, the headers are indented according to the updated indent-script.

closes #218 


holke (changes requested):
Looks good.

For all functions that compute something that is not immediately returned, the result should be checked as well.

I know that we may not do this for the other elements - but we should.

holke (approved):

PR #230, opened 2022-05-03
Added T8_ECLASS_INVALID

Davknapp:
T8_ECLASS_INVALID was added to the enumerator t8_eclass to be used for the case t8code does not support a type of element that is used by a third-party library, for example vtk. As all other eclasses already have a meaning, this new class is introduced.

### DIFF
--- a/src/t8_eclass.h
+++ b/src/t8_eclass.h
@@ -56,7 +56,9 @@ typedef enum t8_eclass
   /** The pyramid has a quadrilateral as base and four triangles as sides. */
   T8_ECLASS_PYRAMID,
   /** This is no element class but can be used as the number of element classes. */
-  T8_ECLASS_COUNT
+  T8_ECLASS_COUNT,
+  /** This is no element class but can be used for the case a class of a third party library is not supported by t8code*/
+  T8_ECLASS_INVALID
 }
 t8_eclass_t;
 
--- a/src/t8_eclass.h
+++ b/src/t8_eclass.h
@@ -145,6 +145,13 @@ int                 t8_eclass_count_boundary (t8_eclass_t theclass,
 int                 t8_eclass_compare (t8_eclass_t eclass1,
                                        t8_eclass_t eclass2);
 
+/** Check, if a class is a valid class. Returns non-zero if it is a valid class,
+ *  returns zero, if the class is equal to T8_ECLASS_INVALID.
+ * 
+ * \param [in] eclass    The eclass to check.
+*/
+int                 t8_eclass_is_valid (t8_eclass_t eclass);
+
 T8_EXTERN_C_END ();
 
 #endif /* !T8_ELEMENT_H */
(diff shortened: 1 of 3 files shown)


holke:
good change.
Please add a small test to the eclass test. We should at least check that T8_ECLASS_INVALID is not equal to any other class (and not ECLASS_COUNT) and that iterating from ECLASS_FIRST to ECLASS_COUNT does not give an invalid class.

Additionally, we should add a function `t8_eclass_is_valid(eclass)` returning 0 or non-0 when the class is valid.

test/t8_gtest_eclass.cxx
```diff
@@ -24,31 +24,41 @@
...
     EXPECT_EQ (t8_eclass_to_dimension[eci], eclass_dims[eci]);
   }
 }
 
+TEST (t8_gtest_eclass, valid_class)
+{
+  int                 eclass;
+  for (eclass = T8_ECLASS_ZERO; eclass <= T8_ECLASS_COUNT; ++eclass) {
+    EXPECT_TRUE (t8_eclass_is_valid ((t8_eclass_t) eclass));
+  }
```
holke, line 46:
```suggestion
  }
  EXPECT_FALSE (t8_eclass_is_valid (T8_ECLASS_INVALID));
```

holke (changes requested):
Nice :)
 only a minor edit regarding the doxygen comment.

Next time please do not commit indented code that does not belong to the PR. (using git add -p)
You can make an extra PR for the indentation of other code.

holke:
Thinking about this...what about T8_ECLASS_COUNT.
This also is no valid eclass, so t8_eclass_is_valid should return 0 here, shouldnt it?

src/t8_eclass.h
```diff
@@ -145,6 +145,13 @@ int                 t8_eclass_count_boundary (t8_eclass_t theclass,
 int                 t8_eclass_compare (t8_eclass_t eclass1,
                                        t8_eclass_t eclass2);
 
+/** Check, if a class is a valid class. Returns non-zero if it is a valid class,
+ *  returns zero, if the class is equal to T8_ECLASS_INVALID.
+ * 
+ * \param [in] eclass    The eclass to check.
```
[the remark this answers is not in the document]
Davknapp: Added the documentation

Davknapp:
> Thinking about this...what about T8_ECLASS_COUNT. This also is no valid eclass in the sense that calling an eclass function with T8_ECLASS_COUNT as input generally is not defined. So t8_eclass_is_valid should return 0 here as well, shouldnt it?

I agree with you and adapted the function accordingly

src/t8_eclass.h
```diff
@@ -143,6 +145,14 @@ int                 t8_eclass_count_boundary (t8_eclass_t theclass,
 int                 t8_eclass_compare (t8_eclass_t eclass1,
                                        t8_eclass_t eclass2);
 
+/** Check, if a class is a valid class. Returns non-zero if it is a valid class,
```
holke, line 148:
```suggestion
/** Check whether a class is a valid class. Returns non-zero if it is a valid class,
```

holke (approved):

PR #235, opened 2022-05-10
Fix indent script paths

holke:

The path of the file with additional data types that is used by t8indent was a relative path in the t8indent script.
This made the other scripts in /scripts that use indent not work, since they are required to be called from within /scripts, but then t8indent did not find the path.

We fixed this by using the absolute path, querying git for the repos path.

### DIFF
--- a/scripts/t8indent
+++ b/scripts/t8indent
@@ -56,8 +56,23 @@ INDENT_BASE_OPTIONS="-bap -bli0 -br -nut -psl -sob -di20"
 #   static t8_cmesh_t
 #   afunction (int aparameter) {
 
+# Get the path of the file t8indent_custom_datatypes.txt in the
+# /scripts subfolder of the repo.
+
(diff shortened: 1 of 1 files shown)


scripts/t8indent
```diff
@@ -56,8 +56,23 @@ INDENT_BASE_OPTIONS="-bap -bli0 -br -nut -psl -sob -di20"
 #   static t8_cmesh_t
 #   afunction (int aparameter) {
 
+# Get the path of the file t8indent_custom_datatypes.txt in the
+# /scripts subfolder of the repo.
+
+repo_main_dir=`git rev-parse --show-toplevel`
+additional_data_types_file="${repo_main_dir}/scripts/t8indent_custom_datatypes.txt"
+
+echo $additional_data_types_file
```
holke, line 65:
```suggestion
```

Davknapp (approved):

PR #257, opened 2022-06-03
Feature ci new ubuntu

holke:

Update the CI to use Ubuntu 22.04.
This step is necessary to prepare for using the VTK library in the CI, since VTK 9.1 is only available from Ubuntu 22.04.

Needed to fix some new compiler warnings.

Unfortunately one warning cannot be fixed, since it is concerned with MPI usage and know by the MPI community.
Disabling the warning only temporarily did not work, so now the compiler option causing this warning (-Wstringop-overflow) is disabled for all CI runs.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -49,7 +49,7 @@ on:
 jobs:
   build:
 
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     
     timeout-minutes: 60
     steps:
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -258,8 +258,8 @@ jobs:
 #
     # build config vars
     - name: build CFLAGS and CXXFLAGS variables
-      run: echo CFLAGS_var="-Wall -pedantic -O3" >> $GITHUB_ENV
-        && echo CXXFLAGS_var="-Wall -pedantic -O3" >> $GITHUB_ENV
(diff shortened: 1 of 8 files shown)


holke:
The problem is an unsolvable issue with autotools library finding and newer C++ compilers.
It occurs when we build with CC=g++
We cannot fix this at the moment.

Proposed solution, that i will try: Only use Ubuntu 22.04. for the VTK runs and use older versions for the other runs.

holke:
> > The problem is an unsolvable issue with autotools library finding and newer C++ compilers. It occurs when we build with CC=g++ We cannot fix this at the moment.
> > Proposed solution, that i will try: Only use Ubuntu 22.04. for the VTK runs and use older versions for the other runs.
> 
> Does this mean, that t8code is not able to run on Ubuntu 22.04 with autotools in general?

No, t8code will (and does) run on Ubuntu 22.04.
Only the configuration `CC=g++ CXX=g++` (or `CC=mpicxx ...`) does fail.
This configuration is covered in several ways in our test cases and is somewhat important.
However, i do not see it as that critical, since most use cases would just use the C compiler.

holke:
For the record, here is a link to the discussion on the autotools mailing list.
https://www.mail-archive.com/bug-autoconf@gnu.org/msg04294.html

When using AC_SEACH_LIB (or similar), autotools builds a test program in the configure script to detect the library.
From gcc v8 on, the C++ compiler throws an error on the test script (instead of a warning as before).
This causes the configure script to fail in finding the library and thus abort.

holke:
I close this now since we cannot solve the autotools issue.

PR #261, opened 2022-06-09
Fix compiler warnings with newer gcc versions

holke:

Fixes compiler warnings that occur with gcc version 11.2.0.
This is the code part from previous PR #257 that we closed without merging.

### DIFF
--- a/src/t8_cmesh/t8_cmesh_stash.h
+++ b/src/t8_cmesh/t8_cmesh_stash.h
@@ -222,7 +222,7 @@ void                t8_stash_attribute_sort (t8_stash_t stash);
  *                  elements in the classes, joinfaces and attributes arrays.
  */
 t8_stash_t          t8_stash_bcast (t8_stash_t stash, int root,
-                                    sc_MPI_Comm comm, size_t elem_counts[]);
+                                    sc_MPI_Comm comm, size_t elem_counts[3]);
 
 /* TODO: specify equivalence relation. is a different order of data allowed? */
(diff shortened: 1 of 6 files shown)


sandro-elsweijer:
@holke 
Can you please elaborate which environment you use? I use gcc version 9.4.0 and the only warnings I get are from sc, p4est and vtk, but none from the compilation of the t8code source code. Therefore, there is also no difference in the make output between this branch and develop.

holke:
> @holke Can you please elaborate which environment you use? I use gcc version 9.4.0 and the only warnings I get are from sc, p4est and vtk, but none from the compilation of the t8code source code. Therefore, there is also no difference in the make output between this branch and develop.

I use an Ubuntu 22.04 with a gcc 11.2.0
```
gcc (Ubuntu 11.2.0-19ubuntu1) 11.2.0
Copyright (C) 2021 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

holke:
I think my comment regarding v >8 was misleading. The autotool errors from the PR before occur from v.8 on, that was my confusion.

sandro-elsweijer (approved):
Even with Ubuntu 22.04 jammy and gcc 11.2.0 I get no warnings. But even though I can't reproduce these warnings, the changes improve the overall usability of the functions and lgtm.

Issue #12, opened 2020-03-31
Find a new script to indent our source code


holke:
We are currently using the GNU `indent` script, however it is actually not designed for
`C++` codes and hence causes some malfunctions there (For example adding an extra const to const functions, which then causes compiler errors).

We should (at least for `C++` files) switch to a different indent program.

holke:
`astyle` seems to be a good choice.
I am currently figuring out the settings that come closest to the existing `indent` settings.

holke:
`clang-format` is a good choice as well.

The main challenge is to find settings that come as close to the current indent settings as possible.
Unfortunately, i was not able to reproduce them completely, neither with astyle nor clang-format.

chiaraMaHe:
still open

cburstedde:
What's the closest in settings that y'all could find? I'd like to experiment. This will be good for p4est and libsc as well.

sandro-elsweijer:
@cburstedde 
here is the clang-format file I created:
https://github.com/DLR-AMR/t8code/blob/feature-new_script_to_indent_our_source_code/.clang-format
Here is also a nice online tool describing all available options:
https://zed0.co.uk/clang-format-configurator/

sandro-elsweijer:
We switched to clang format

PR #262, opened 2022-06-10
Switch from gnuindent to clang-format

sandro-elsweijer:
Switch from `gnuindent` to `clang-format`, as described in #12.
The indentation from `gnuindent` is not perfectly replicable with `clang-format`, but I think I am as close as I can get.

There are still some options to experiment with (especially the C++ related options), so that our code looks nice. I did not indent the source code yet, because we are waiting for the updates for version 1.0 to be merged, so we can avoid as many merge conflicts as possible.
Feel free to look if the formatting looks good for you, I am happy over each suggestion.

The biggest changes are:
1. `gnuindent` aligns consecutive declarations like this:  
```c
int         var1;
char        var2;
longlong    var3;
```  
`clang-format` has the option to do this as well, but then it also aligns the declarations inside function definitions:  
```c
void myfunc (int         var1,
             char        var2,
             longlong    var3);
```
Therefore, the alignment of consecutive declarations is disabled.

2. The column limit is now 120 instead of 80
3. `gnuindent` sometimes breaks before `&&`, `||` etc. and sometimes after them. `clang-format` now always breaks before `&&`, `||` etc.
4. `/* *INDENT-OFF* */` has to be switched to `/* clang-format off */`. I have found no way to edit the clang macro.
5. C++ gets indented differently by design. This is the reason why we want to move away from `gnuindent`. 
6. `gnuindent` ignored the length of comments, `clang-format` does not. Therefore, long, but still nice looking lines from `gnuindent`  are now broken, because we often use trailing comments instead of leading comments.

### DIFF
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -11,8 +11,8 @@ The purpose of the indentation scripts is to help t8code developers to indent th
 Apply this script to a `.c`, `.cxx`, `.h` or `.hxx` file to indent it according to the guidelines.
 You can also provide a list of files and all will get indented.
 
-This script uses the `GNU indent` program with t8code specific settings.
-Sometimes `t8indent` does produce undesired results. Therefore, after indenting use `git add -p` or similar to check all changes before committing. You can protect code lines from being changed by the script by enclosing them in `/* *INDENT-ON* */` and `/* *INDENT-OFF* */` comments.
+This script uses the `clang-format` program with t8code specific settings.
+Sometimes `t8indent` does produce undesired results. Therefore, after indenting use `git add -p` or similar to check all changes before committing. You can protect code lines from being changed by the script by enclosing them in `/* clang-format off */` and `/* clang-format on */` comments.
 See also [Known issues with indent](https://github.com/holke/t8code/wiki/Known-issues-with-the-indent-script).
 
 #### pre-commit
@@ -34,10 +34,6 @@ This script indents all t8code source files at once. This script should only be 
 #### find_all_source_files.scp
 
 List all source files of t8code in the `src/` `example/` and `test/` subfolders.
-
-#### remove_double_const.scp
-
-One of the [Known issues with indent](https://github.com/holke/t8code/wiki/Known-issues-with-the-indent-script) is that it duplicates the `const` keyword in `C++` files in some cases. This script iterates over all `C++` files and tries to fix these duplicates. Handle with care.
 
 ## Others
 


holke:
When executing the `check_if_file_indendet` script my clang-format complains:
```
clang-format: Unknown command line argument '--dry-run'.  Try: '/usr/bin/clang-format -help'
clang-format: Did you mean '-icp-lto'?
clang-format: Unknown command line argument '--Werror'.  Try: '/usr/bin/clang-format -help'
clang-format: Did you mean '-cursor'?
```

i am running clang-format version 6.0.0:
```
$ clang-format --version
clang-format version 6.0.0-1ubuntu2 (tags/RELEASE_600/final)
```

If these arguments are needed we need to specify which clang-format version is needed and check that we fulfill before trying to indent the files.

holke:
clang-format-10 seems to be the first fully supported version.
It is available on my ubuntu, but i have to explicitly specify clang-format-10 in the indent scripts.

holke:
Do you know what causes this behaviour:
```
-  else {                        /* set_from != NULL */
-    t8_forest_t         forest_from = forest->set_from; /* temporarily store set_from, since we may overwrite it */
+  else { /* set_from != NULL */
+    t8_forest_t forest_from
+      = forest
+          ->set_from; /* temporarily store set_from, since we may overwrite it */
```

The new variant looks rather ugly tbh.

.clang-format
```diff
@@ -0,0 +1,74 @@
...
+  BeforeElse:      true
+  IndentBraces:    false
+  SplitEmptyFunction: true
+  SplitEmptyRecord: true
+  SplitEmptyNamespace: true
+BreakBeforeBinaryOperators: All
+BreakBeforeBraces: Custom
+BreakBeforeTernaryOperators: 'true'
+BreakInheritanceList: AfterColon
+ColumnLimit: '80'
```
holke, line 41: Shouldn't this by 120?

holke:
Thanks for adding this and putting the work into it.
We definitely need this change.

I am playing around with it a bit and add a few comments regarding fine tuning.

.clang-format
Davknapp: Changed

Davknapp:
> Do you know what causes this behaviour:
> 
> ```
> -  else {                        /* set_from != NULL */
> -    t8_forest_t         forest_from = forest->set_from; /* temporarily store set_from, since we may overwrite it */
> +  else { /* set_from != NULL */
> +    t8_forest_t forest_from
> +      = forest
> +          ->set_from; /* temporarily store set_from, since we may overwrite it */
> ```
> 
> The new variant looks rather ugly tbh.

I am unable to reproduce this behaviour using clang-format-16

Davknapp:
> > Do you know what causes this behaviour:
> > ```
> > -  else {                        /* set_from != NULL */
> > -    t8_forest_t         forest_from = forest->set_from; /* temporarily store set_from, since we may overwrite it */
> > +  else { /* set_from != NULL */
> > +    t8_forest_t forest_from
> > +      = forest
> > +          ->set_from; /* temporarily store set_from, since we may overwrite it */
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The new variant looks rather ugly tbh.
> 
> I am unable to reproduce this behaviour using clang-format-16

Retried with version 10.0.1 and can't reproduce this error either.

Davknapp (approved):

PR #270, opened 2022-06-23
Feature connectivity tests

Davknapp:
Added two tests that check the correctness of the relations between type, local_id, cube_id and type of the parent implement in t8_dpyramid_connectivity.c
During testing I also found a small bug and fixed it. 


test/t8_schemes/t8_test_pyra_connectivity.cxx
```diff
@@ -0,0 +1,107 @@
...
+      parent_type = t8_dpyramid_cid_type_to_parenttype[cid][type];
+      SC_CHECK_ABORTF (parent_type == pyra_parent_type,
+                       "Look-up tables are different");
+    }
+  }
+}
+
+/**
+ * Given the type of a parent element, iterate over all local ids of the children
+ * and look-up their type and cube-id. Usings these compute the parent-type again
```
sandro-elsweijer, line 57:
```suggestion
 * and look-up their type and cube-id. Using these to compute the parent-type again
```

sandro-elsweijer (changes requested):
I changed some values in the pyramid connectivity and almost every change resulted in a failing test, I am impressed about how good this test works.
@holke requested, that we implement the new tests as google test. Should we implement this test also as google test?

Davknapp:
> I changed some values in the pyramid connectivity and almost every change resulted in a failing test, I am impressed about how good this test works. @holke requested, that we implement the new tests as google test. Should we implement this test also as google test?

Thank you!
I will transform the test into a google-test

sandro-elsweijer (commented):
I was able to change the `1` in 
https://github.com/holke/t8code/blob/615500e272efd5d3a534d0b8fdbf4930935a01a2/src/t8_schemes/t8_default/t8_default_pyramid/t8_dpyramid_connectivity.c#L32
into a `2` without any test failing. Do you know why this happens?

Davknapp:
> I was able to change the `1` in
> 
> https://github.com/holke/t8code/blob/615500e272efd5d3a534d0b8fdbf4930935a01a2/src/t8_schemes/t8_default/t8_default_pyramid/t8_dpyramid_connectivity.c#L32
> 
> 
> into a `2` without any test failing. Do you know why this happens?

Yes, if you use 
```t8_debugf("cid: %i, type: %i \n", cid, type);```
when p_type is 3 and Iloc is 4, you see that the cid is 6 and the type is 1 (2, if you change it in the lookup).

line 77 computes the check_type and in:
https://github.com/holke/t8code/blob/615500e272efd5d3a534d0b8fdbf4930935a01a2/src/t8_schemes/t8_default/t8_default_pyramid/t8_dpyramid_connectivity.c#L73
there are multiple 3. That is why the test does not fail.

Idealy we should implement tests for all look-up-tables in the connectivity-files.

Davknapp:
> > I was able to change the `1` in
> > https://github.com/holke/t8code/blob/615500e272efd5d3a534d0b8fdbf4930935a01a2/src/t8_schemes/t8_default/t8_default_pyramid/t8_dpyramid_connectivity.c#L32
> > 
> > into a `2` without any test failing. Do you know why this happens?
> 
> Yes, if you use `t8_debugf("cid: %i, type: %i \n", cid, type);` when p_type is 3 and Iloc is 4, you see that the cid is 6 and the type is 1 (2, if you change it in the lookup).
> 
> line 77 computes the check_type and in:
> 
> https://github.com/holke/t8code/blob/615500e272efd5d3a534d0b8fdbf4930935a01a2/src/t8_schemes/t8_default/t8_default_pyramid/t8_dpyramid_connectivity.c#L73
> 
> 
> there are multiple 3. That is why the test does not fail.
> Idealy we should implement tests for all look-up-tables in the connectivity-files.

The problem is, that the current test does not test the arrays independantly. Theoratically one could change all three arrays such that the test works fine, but the arrays themself aren't correct anymore. Do you have an idea to check the arrays independantly (without creating another look-up-table at the test)?

sandro-elsweijer (approved):

PR #279, opened 2022-07-05
Extend t8_forest_write_netcdf_ext and add benchmark CLI

neoq:
Hi, I am a student of Prof. Julian Kunkel at University of Göttingen. In this PR I extend t8_forest_write_netcdf_ext to allow customizing the fillmode, the nc_create cmode (NC_NETCDF4 or NC_CLASSIC...), as well as the chunksize for the node coordinates. This gives the user more flexibility when writing netcdf files if desired. I also add a file-per-process mode where each MPI rank writes to its own file.
Additionally I add a benchmark under example/netcdf/benchmark/t8_write_forest_netcdf.cxx based on example/netcdf/t8_write_forest_netcdf.cxx that allows configuring these parameters with a CLI. It also takes a hint for the written storage space.
We want to use this to evaluate and possibly improve the performance of the netcdf writing routines.

I will make some adjustments before removing the draft status:
- Well documented code (doxygen)
- Tests
- verify build without MPI


neoq:
I think I need a little help with getting t8code to compile, firstly --without-netcdf, and secondly with --disable-mpi.
What should the `t8_forest_write_netcdf*` functions do without netCDF?
Can `T8_WITH_NETCDF_PAR` be true in a --disable-mpi configuration?

Thanks

Niklas997:
> I think I need a little help with getting t8code to compile, firstly --without-netcdf, and secondly with --disable-mpi. What should the `t8_forest_write_netcdf*` functions do without netCDF?

These functions should basically do "nothing" and maybe print out an error message stating that t8code is not configured with netCDF. But they should be compilable and callable despite the absence of netCDF (eventually by embracing the function bodies by something like "#if T8_WITH_NETCDF ... #endif"-directives). Maybe @holke can elaborate further on how the functions should behave in this case.

> Can `T8_WITH_NETCDF_PAR` be true in a --disable-mpi configuration?

This used to be true in some circumstances, but it is fixed in the current branch, so disabling MPI always sets "T8_WITH_NETCDF_PAR" to false.

Niklas997:
Thank you very much for your contribution! These are some nice features and the code looks good.
Prior to the pull request, the "example"-directory structure within the t8code repository changed. Therefore, there are currently some conflicting files, please resolve these conflicts.
There is also a "benchmark"-folder now, in which we could place the netcdf-benchmark file eventually (instead of placing it inside the "example"-folder).

PR #314, opened 2022-08-18
Enhancement pyramid switch shape at level

Davknapp:
This PR introduces a new field 'switch_shape_at_level' for pyramid elements. 
Everytime a pyramid is constructed we set this variable. If the shape of the constructed element is a pyramid, it is set to -1. Otherwise (the shape is a tetrahedron) we set it to the level of last ancestor in the shape of a tetrahedron. 
This variable is then used in every other algorithm that needs information about ancestors and speeds up the computation significantly. First tests showed a speed-up around 20% for adapt and balance algorithms, without any changes in the other high-level algorithms. 
This work has been done in collaboration with @lukasdreyer .


src/t8_schemes/t8_default/t8_default_pyramid/t8_default_pyramid_cxx.cxx
```diff
@@ -300,7 +300,7 @@ t8_default_scheme_pyramid_c::t8_element_is_family (t8_element_t **fam)
     T8_ASSERT (t8_element_is_valid (fam[i]));
   }
 #endif
-  return t8_dpyramid_is_family ((const t8_dpyramid_t **) fam);
+  return t8_dpyramid_is_family ((t8_dpyramid_t **) fam);
```
lukasdreyer, line 303: Reminder: Check why const is no longer possible here, because is_family should leave the elements const

lukasdreyer:
In the future, we should try to split MRs like this in smaller chunks.

For example here: One MR that introduces the new t8_dpyramid_t, and all the `.pyramid` changes, and one MR that actually contains the logic that we try to change.
In the current way, it is difficult to filter out all the `.pyramid` changes when trying to review the actual logic changes

lukasdreyer:
> In the future, we should try to split MRs like this in smaller chunks.
> 
> For example here: One MR that introduces the new t8_dpyramid_t, and all the `.pyramid` changes, and one MR that actually contains the logic that we try to change. In the current way, it is difficult to filter out all the `.pyramid` changes when trying to review the actual logic changes

I have introduced 3 PRs that split the functionality introduced in this PR to more managable chunks: https://github.com/DLR-AMR/t8code/pull/347 https://github.com/DLR-AMR/t8code/pull/348 https://github.com/DLR-AMR/t8code/pull/349
They replace this PR.

PR #347, opened 2022-10-12
Preparation for improved way to access parent/ancestor information of pyramids 

lukasdreyer:
Extend the t8_dpyramid_t struct, so that it is no longer the same as a t8_dtet_t struct, but instead has such a struct and additional information as variables.

In this PR, all source and test code is adapted to the new struct, so that additional functionality can be introduced in the upcoming PRs. This is mainly the inclusion of `pyramid.`, so `p->x` is changed to `p->pyramid.x`

The PR already includes the new variable switch_shape_at_level in the pyramid struct, although it is not yet used. The actual functionality will be introduced in the next PR.

### DIFF
--- a/src/t8_schemes/t8_default/t8_default_pyramid/t8_dpyramid.h
+++ b/src/t8_schemes/t8_default/t8_default_pyramid/t8_dpyramid.h
@@ -55,7 +55,17 @@ typedef int32_t     t8_dpyramid_coord_t;
 */
 typedef int8_t t8_dpyramid_type_t;
 
-/** This data type stores a pyramid. */
-typedef t8_dtet_t t8_dpyramid_t;
+/**
+ * This data type stores a pyramid. 
+ * The coordinates, the level and the type of a pyramid are stored in the tet-struct \a pyramid.
+ * Level, at which the shape switches from tet, to pyra. -1 if not computed for a pyramid with the shape of a tet
+ * undefined, if the pyramid has the shape of a pyramid.
+ */
+typedef struct t8_dpyramid
+{
+  t8_dtet_t           pyramid;  /* Coordinates, level and type */
+  int8_t              switch_shape_at_level;
+}
+t8_dpyramid_t;
 
 #endif /* T8_DPYRAMID_H */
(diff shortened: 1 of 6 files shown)


src/t8_schemes/t8_default/t8_default_pyramid/t8_dpyramid.h
```diff
@@ -55,7 +55,17 @@ typedef int32_t     t8_dpyramid_coord_t;
...
-typedef t8_dtet_t t8_dpyramid_t;
+/**
+ * This data type stores a pyramid. 
+ * The coordinates, the level and the type of a pyramid are stored in the tet-struct \a pyramid.
+ * Level, at which the shape switches from tet, to pyra. -1 if not computed for a pyramid with the shape of a tet
+ * undefined, if the pyramid has the shape of a pyramid.
+ */
+typedef struct t8_dpyramid
+{
+  t8_dtet_t           pyramid;  /* Coordinates, level and type */
```
jmark, line 66: I'm wondering if this is the most elegant solution in order to extend the tetraeder type to a pyramid type.

I think it is worth to think about if instead we define the fields explicitly:
```
typedef struct t8_dpyramid
{
#ifdef T8_ENABLE_DEBUG
  /* TODO: enum to int8??? #defines ?? */
  int8_t              eclass_int8;
#endif
  /** The refinement level of the tetrahedron relative to the root at level 0. */
  int8_t              level;

  /** Type of the tetrahedron in 0, ..., 5. */
  t8_dtet_type_t      type;

  t8_dtet_coord_t     x;        /**< The x integer coordinate of the anchor node. */
  t8_dtet_coord_t     y;        /**< The y integer coordinate of the anchor node. */
  t8_dtet_coord_t     z;        /**< The z integer coordinate of the anchor node. */

   /**< Pyramid specific fields. */
  int8_t switch_shape_at_level;
}
t8_dpyramid_t;
```

Whenever you are calling tetraeder functions you can do a type casting:
```
do_something_tet_specific((t8_dtet_t *) my_pyramid_ptr);
```
This actually works, since the tetraeder fields are at the beginning of the struct.

If you do it this way, you don't have to change so much in `*_bits.c`.

jmark (changes requested):

src/t8_schemes/t8_default/t8_default_pyramid/t8_dpyramid.h
Davknapp: I have mixed feelings about this. Yes, it would works, but that way we give over the controll to the typecasting. In the current way, all controll is in our hands, which I prefer.

jmark: So, you don't trust your compiler? :P

jmark (approved):

PR #357, opened 2022-10-26
Add a debug function that prints the trees of the cmesh

Davknapp:
This adds a function to print the vertices of each local tree of each cmesh.
If the cmesh is not partitioned, only process 0 prints the vertices of the cmesh. 
Otherwise the processes send their rank in a round-robin style, and after a process has received its message it prints the local trees and sends the next message. 

Why is this function necessary? I have a bug that mixes up the vertices and hope to find the source of the problem with this function. 

- [x] New tests use the Google Test framework

### DIFF
--- a/src/t8_cmesh.h
+++ b/src/t8_cmesh.h
@@ -644,7 +644,7 @@ void                t8_cmesh_print_profile (t8_cmesh_t cmesh);
 
 /** Return a pointer to the vertex coordinates of a tree.
  * \param [in]    cmesh         The cmesh.
- * \param [in]    ltreeid       The id of a loca tree.
+ * \param [in]    ltreeid       The id of a local tree.
  * \return    If stored, a pointer to the vertex coordinates of \a tree.
  *            If no coordinates for this tree are found, NULL.
  */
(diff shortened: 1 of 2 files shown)


src/t8_cmesh.h
```diff
@@ -769,6 +769,17 @@ void                t8_cmesh_new_translate_vertices_to_attributes (t8_locidx_t
...
+
+/**
+ * \warning This function is only available in debug-modus and should only 
+ * be used in debug-modus.
+ * 
+ * Prints the vertices of each tree of each process
+ * 
+ * \param[in] cmesh   Source-cmesh, which trees get printed.
+ */
+void                t8_cmesh_debug_print_trees (const t8_cmesh_t cmesh,
```
jmark, line 781: Put this function also in debug clause. This will error in case of `-Werror=unused-function`.

jmark (changes requested):

src/t8_cmesh.h
Davknapp: Just happened in the CI^^

jmark (approved):

PR #366, opened 2022-11-21
Test cmesh face is boundary

WeberJulia:
test to gtest: cmesh_face_is_boundary

- [x] New tests use the Google Test framework


test/t8_cmesh/t8_gtest_cmesh_face_is_boundary.cxx
```diff
@@ -0,0 +1,188 @@
...
+    cmesh = t8_cmesh_new_from_class (eclass, sc_MPI_COMM_WORLD);
+  }
+  void TearDown() override {
+    t8_cmesh_destroy(&cmesh);
+  }
+
+  t8_cmesh_t          cmesh;
+  t8_eclass           eclass;
+  int                 num_faces;
+  int                 iface;
```
Davknapp, line 45: This only used as an iterator in the for-loops. You can declare and initialize the variable in the for-loop via:
`for(int iface=1.....)`. That way, you can only use iface in the for-loop and not use it by accident somewhere else, which is a little safer.

test/t8_cmesh/t8_gtest_cmesh_face_is_boundary.cxx
```diff
@@ -0,0 +1,188 @@
...
+};
+
+TEST_P (cmesh_face_boundary_one_tree, check_face_is_boundary_one_tree) {
+  
+    EXPECT_TRUE(t8_cmesh_is_committed (cmesh));
+    /* We now check each face */
+    num_faces = t8_eclass_num_faces[(int) eclass];
```
Davknapp, line 52: Can this be moved in the SetUp()?

test/t8_cmesh/t8_gtest_cmesh_face_is_boundary.cxx
```diff
@@ -0,0 +1,188 @@
...
+    t8_test_compute_parallel_bounds (sc_MPI_COMM_WORLD, &first_tree, &last_tree);
+
+  }
+
+  t8_cmesh_t          cmesh;
+
+  t8_eclass           eclass;
+  int                 do_partition;
+  int                 num_faces;
+  int                 iface;
```
Davknapp, line 118: This should be moved out of the class

test/t8_cmesh/t8_gtest_cmesh_face_is_boundary.cxx
```diff
@@ -0,0 +1,188 @@
...
+  int                 checkface;
+  t8_gloidx_t         first_tree;
+  t8_gloidx_t         last_tree;
```
Davknapp, line 121: Please move all variables, that can not be initialized during SetUp out of the class. They are helper-variables for the test and it is ok to declare them there. We could write another test using this class which does not need ```checkface``` for example.

test/t8_cmesh/t8_gtest_cmesh_face_is_boundary.cxx
```diff
@@ -0,0 +1,185 @@
...
+/* *INDENT-OFF* */
+class cmesh_face_boundary_one_tree : public testing::TestWithParam<t8_eclass>{
+protected:
+  void SetUp() override {
+    eclass = GetParam();
+
+    /* For each eclass create a cmesh consisting only of one tree.
+     * We then check whether all faces of this tree are a boundary face. */
+    cmesh = t8_cmesh_new_from_class (eclass, sc_MPI_COMM_WORLD);
+    EXPECT_TRUE(t8_cmesh_is_committed (cmesh));
```
Davknapp, line 37: We should keep checks and construction seperated, even though we add a little bit of code duplication into the tests. But I think that a clear seperation of constructor and testing keeps the code more clear.

Davknapp (changes requested):
Please address my last comment. I will try the tests on my machine, then we are probably ready to merge.

Davknapp:
I have just realized, that this test appears also in PR #378 , where I added some remarks regarding the usage of `EXPECT_xx` from the google-test framework. Can you please:
1. address the comments made there
2. Make one PR per update/test in the future?

Davknapp (approved):

PR #368, opened 2022-11-22
Feature t8 io

Davknapp:
Implementation of a general IO Interface to centralize all IO-routines in t8code. 
Goal of this Interface is to:
- Move all IO routines (like gmsh-reader, vtk-output, netcdf-IO...) in one place
- Implement general IO routines, especially for serial/parallel readers. For objects, that define a cmesh we can extend a serial reader (read one object on one master proc) to a parallel reader (read multiple objects, each defining a cmesh on multiple processes). 
- Avoid code duplication for similar IO-routines but for different objects. 

This PR provides only the structure to meet these ideas. I will move the current implementations of supported IO-routines as soon as possible. 


src/t8_IO/t8_reader/t8_vtk_reader.hxx
```diff
@@ -0,0 +1,58 @@
...
+public:
+  vtk_path * filepath;
+  /* Constructor */
+  t8_vtk_reader ();
+  /* Destructor */
+  ~t8_vtk_reader ();
+  /* Write the output */
+  virtual t8_read_status_t read ();
+
+  virtual t8_read_status_t set_source (const t8_extern_t * source);
```
jmark, line 52: I don't think this is a good idea and exactly what we wanted to avoid, right? We should talk about this. :)

src/t8_IO/t8_IO_cxx.hxx
```diff
@@ -0,0 +1,195 @@
...
+  {
+    return partition;
+  }
+  /**
+   * Set the source object, if source is not NULL
+   * 
+   * \param[in] source an object to be filled.
+   * \return t8_write_status_t T8_WRITE_FAIL if it wasn't able to set the source, T8_WRITE_SUCCESS otherwise
+   */
+  virtual t8_read_status_t set_source (const t8_extern_t * source) = 0;
```
jmark, line 153: This routine does not have be part of the base class. We should talk about this.

src/t8_IO/t8_writer/t8_vtk_writer.cxx
```diff
@@ -0,0 +1,71 @@
...
+t8_vtk_writer::write ()
+/* *INDENT-ON* */
+{
+  t8_debugf ("[D] write_test\n");
+  return T8_WRITE_SUCCESS;
+}
+
+/* *INDENT-OFF* */
+t8_write_status 
+t8_vtk_writer::set_dest (const t8_extern_t * dest)
```
jmark, line 42: The same problem as with `set_source`.

jmark (changes requested):
Really nice starting point for future discussion! :)

PR #473, opened 2023-03-14
Replace non-standard u_int64_t by t8_linearidx_t

sloede:
In some locations (to be exact: 4), the non-standard type `u_int64_t` was used, which prevents compiling t8code on some platforms (e.g., Linux with musl instead of glibc). This PR fixes this by using the standardized type `uint64_t`.

cc @jmark

### DIFF
--- a/src/t8_schemes/t8_default/t8_default_prism/t8_dprism_bits.c
+++ b/src/t8_schemes/t8_default/t8_default_prism/t8_dprism_bits.c
@@ -53,7 +53,7 @@ int
 t8_dprism_compare (const t8_dprism_t *p1, const t8_dprism_t *p2)
 {
   int                 maxlvl;
-  uint64_t            id1, id2;
+  t8_linearidx_t            id1, id2;
   T8_ASSERT (p1->line.level == p1->tri.level);
   T8_ASSERT (p2->line.level == p2->tri.level);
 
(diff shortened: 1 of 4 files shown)


Davknapp (changes requested):
Thanks for catching that error! In all cases the variables describe a linear Id, for which we use the datatype ```t8_linearidx_t```, which is typedefed ```uint64_t```. With this change it should work.

src/t8_schemes/t8_default/t8_default_prism/t8_dprism_bits.c
```diff
@@ -53,7 +53,7 @@ int
 t8_dprism_compare (const t8_dprism_t *p1, const t8_dprism_t *p2)
 {
   int                 maxlvl;
-  u_int64_t           id1, id2;
+  t8_linearidx_t            id1, id2;
```
sloede, line 56:
```suggestion
  t8_linearidx_t      id1, id2;
```
Fix indentation

src/t8_schemes/t8_default/t8_default_tet/t8_default_tet.c
```diff
@@ -59,7 +59,7 @@ static int
 t8_default_tet_compare (const t8_element_t *elem1, const t8_element_t *elem2)
 {
   int                 maxlvl;
-  u_int64_t           id1, id2;
+  t8_linearidx_t            id1, id2;
```
sloede, line 62:
```suggestion
  t8_linearidx_t      id1, id2;
```
Fix indentation

Davknapp (approved):
Thank you for your addition to t8code

sloede:
It would be great if you can also tag v1.1.1 such that it can be used for T8code_jll.jl.

PR #476, opened 2023-03-16
Allow building t8code for Windows (x86_64)

sloede:
t8code is already set up to be portable to many other architectures thanks to the (mostly) rigorous use of standardized C/C++ code. For building t8code on 64-bit Windows systems via MinGW, there are two small exceptions: the functions `getline` and `strsep` are not part of the C standard, and they are not included in MinGW, thus compilation (or linking) fails on MinGW-based Windows systems.

This PR remedies this by providing an implementation for `getline` (based on public domain code) and `strsep` (own implementation) and includes them in the relevant source files of t8code. The inclusion of those files is guarded by a Windows-specific preprocessor directive, such that non-Windows platforms are not affected at all.

With this PR, it will be possible to, e.g., build t8code with MPI support for Julia users with a Windows system (I have done so successfully on a local machine).

cc @jmark @Davknapp 

### DIFF
--- a/src/t8_cmesh/t8_cmesh_readmshfile.cxx
+++ b/src/t8_cmesh/t8_cmesh_readmshfile.cxx
@@ -29,6 +29,10 @@
 #include "t8_cmesh_types.h"
 #include "t8_cmesh_stash.h"
 
+#ifdef _WIN32
+#include "t8_windows.h"
+#endif
+
 /* The supported number of gmesh tree classes.
  * Currently, we only support first order trees.
  */
--- a/src/t8_cmesh/t8_cmesh_triangle.c
+++ b/src/t8_cmesh/t8_cmesh_triangle.c
@@ -27,6 +27,10 @@
 #include "t8_cmesh_types.h"
 #include "t8_cmesh_stash.h"
 
+#ifdef _WIN32
+#include "t8_windows.h"
+#endif
+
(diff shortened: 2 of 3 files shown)


cburstedde:
This is nice; I'm wondering whether this may not live better in libsc. Libsc also has quite a bit of Windows CMake code etc. that may be useful.

I'm a bit concerned that we are diverging the build systems and portability code between p4est and t8code.

sloede:
> This is nice; I'm wondering whether this may not live better in libsc. Libsc also has quite a bit of Windows CMake code etc. that may be useful.
> 
> I'm a bit concerned that we are diverging the build systems and portability code between p4est and t8code.

I see your point. To be honest, IMHO the Right Way™️ would be to avoid using non-standard C functions altogether and instead use standardized equivalents or, if that's too hard, in the case of t8code just implement the relevant part in C++.

But, since @jmark is on the verge of creating a precompiled binary package of t8code for Julia, maybe it would be an option to use the current approach for now but to create an issue for it such that it can be properly cleaned up later. But of course, that's a main developer decision you folks have to make :-)

holke:
Thank you for the great contribution @sloede
Yes, in the long term this implementation should move the libsc.
I agree that we should merge now to get the Trixi package forward and then properly add it to sc and update the sc version later on.

jmark (approved):

Issue #496, opened 2023-03-27
Compile-dependency on libtool archives of p4est and sc


jmark:
The build script of `t8code` looks for `lib/libp4est.la` and `lib/libsc.la` files in order to properly build.

Even then, when given external paths like so: `configure --with-p4est=... --with-sc=...`.

However, this is, according to my sources, not standard (anymore), and many package distributors of libraries delete these
libtool archive files before bundling.

Thus, t8code should be able to compile without these files.

jmark:
@cburstedde I think this might is also be a problem upstream in the build system of p4est.

cburstedde:
We are using libtool in a pretty standard way. Has there been an update in how to invoke it from automake?
If we are doing it as the majority of gnu packages, we should be on the safe side.

If anything is to be done,  I think this should be updated in libsc, p4est and t8code in the same way, by the most recent documentation of libtool.

Btw, the p4est autoconf logic is ahead of t8code in a couple of small details. Would it make sense to align t8code with the most recent p4est automake/libtool setup first and then recheck what needs to be done?

cburstedde:
I've just make installed libsc, then removed the libsc.la file, and configured p4est --with-libsc= to that path. It built fine on my linux machine. How does it work for y'all with the latest prev3-releases of p4est (API 2:0:0)?

sloede:
> How does it work for y'all with the latest prev3-releases of p4est (API 2:0:0)?

I can give it a try, but I am not sure which version of p4est I should use: 2.8? 2.3.5? Or the prev3-develop branch with bootstrapping? I couldn't find a `prev3-releases` branch :-/

cburstedde:
I've made some major version bumps of everything to libtool API version 2:0:0.
The master branch is now current (previously it was tracking years-old backward compatibility code).

New PRs can go against develop for new features, against master for maintenance-related code/doc/meta.

sloede:
Great, thanks for letting us know! I saw in https://github.com/p4est/p4est.github.io/tree/master/release/ that there is now a v2.8.5 and a v2.8.5.5-... release tarball. Am I correct to assume that v2.8.5.5-... is the most up-to-date one?

sloede:
[...]

Issue #483, opened 2023-03-19
Use standard soname for library products


sloede:
Right now, when building t8code, the library products `libt8.so`, `libsc.so`, and `libp4est.so` have the full version and some hash encoded in the filename. For Linux, the soname for a library should be something like `libLIBRARYNAME.so.X.Y.Z`, where `X.Y.Z` are *not* the library's release number but rather the *API version*. That is, the version should only change if there are backwards incompatible changes to the library API (see e.g., https://tldp.org/HOWTO/Program-Library-HOWTO/shared-libraries.html).

Right now, t8code produces a soname with its version and some has encoded in it, e.g., `libt8-1.1.0.207-d6a74.so`, which is then linked symbolically to the generic `libt8.so` (similar for libp4est and libsc). This makes it somewhat inconvenient to use, as you can only link to a single library version at a time (see also the remark by @giordano at https://github.com/JuliaPackaging/Yggdrasil/pull/6389#issuecomment-1475283561). When building p4est itself, it actually does what is expected: it creates a file named `libp4est.so.0.0.0`, which is linked to `libp4est.so.0`, which is linked to `libp4est.so` (similarly for libsc).

Long story short, I think it would be beneficial to change the build system of t8code to reflect this Unix/Linux convention of how libraries are named. At the moment, though, I have not encountered any problems myself yet, so for me this is not an urgent issue (yet) 🙂

cc @jmark

cburstedde:
> Long story short, I think it would be beneficial to change the build system of t8code to reflect this Unix/Linux convention of how libraries are named. At the moment, though, I have not encountered any problems myself yet, so for me this is not an urgent issue (yet) 🙂

It is sufficient to add -version-info to LDFLAGS of the library in Makefile.am.

Please see how I just did this with the oldcompat branches of p4est and libsc.

The libtool manual has a whole section on versioning.

sloede:
Interesting. However, I do not understand why the current versions of t8code and p4est behave differently, if this flag has not been used before on the p4est side. That is, why does p4est 2.8 install "correct" library names, while t8code encodes the full version has. Do you have an idea how this could happen?

cburstedde:
I've added API version 2:0:0 to the latest p4est and libsc, which previously defaulted to 0.
I'm adding the same API version to t8code in #499 to start fresh with the numbering.

In t8code, in addition to -version-info there is -release $(VERSION), which may be needed or not (?).

[...]

PR #506, opened 2023-04-04
Make install: use libraries instead of .la and switch to standard *.so naming

cburstedde:

I've worked on the make-install build. This branch does not require any .la files in this mode.
Try configuring --with-sc= and --with-p4est=, where both have been make installed before.
I'm adding an rpath to the t8code executables for convenience so they run from anywhere.

The submodules are committed but not versioned at this point. Will do when approved.
The standard build without configure options should work unchanged.

Comments?


jmark:
@cburstedde When I compile this branch, with and without `--with-{sc,p4est}` then `libt8.so` does not point to `libsc.so` and `libp4est.so`. Is this on purpose?

cburstedde:
> @cburstedde When I compile this branch, with and without `--with-{sc,p4est}`, then `libt8.so` does not point to `libsc.so` and `libp4est.so`. Is this on purpose?

Yes, I decided to not give any -L and -l flags to building the library itself. What do you think is the better convention?
This mechanism does not work on half the existing platforms anyway according to the libtool documentation.

jmark:
For the Julia wrapper we run into problems, since the wrapper builder does some linking checks directly on `libt8.so`. They fail since `libp4est.so` and `libsc.so` are not loaded.

cburstedde:
So maybe it would make sense to add -L and -l of dependencies to the LIBADD of the higher-level build.
I can try this and see if that solves these problems.

jmark:
When this issue https://github.com/cburstedde/p4est/issues/204 is resolved, we can look at this PR again.

@cburstedde

jmark:
The current state of this PR fixes the issues raised in https://github.com/DLR-AMR/t8code/issues/483 and https://github.com/DLR-AMR/t8code/issues/496.

We should wait for the merge till https://github.com/cburstedde/libsc/pull/114 (libsc) and https://github.com/cburstedde/p4est/pull/208 (p4est) are merged into the main branches.

@cburstedde @sloede

src/Makefile.am
```diff
@@ -132,10 +132,14 @@ src_libt8_la_SOURCES = \
 src_libt8_la_CPPFLAGS = $(AM_CPPFLAGS) $(T8_CPPFLAGS)
 ## This is the official API versioning scheme of libtool.  Please see:
 ## Read https://www.gnu.org/software/libtool/manual/libtool.html#Versioning
-src_libt8_la_LDFLAGS = -release $(VERSION) -version-info 2:0:0
+src_libt8_la_LDFLAGS = -version-info 2:0:0
 src_libt8_la_LIBADD = @T8_P4EST_LDADD@ @T8_SC_LDADD@
```
cburstedde, line 136: This works by accident; with the latest commits on feature-make-install you can use T8_*_LIBADD.

jmark: Can you elaborate respectively propose a patch? I'm not savvy in autotools and coded by trial-and-error only.

sloede (commented):
I'm not an expert on autotools, but with the proposed change by Carsten this LGTM

jmark:
@cburstedde 

The t8code tests on Github CI fails with:
```
g++: error: T8_P4EST_LIBADD@: No such file or directory
g++: error: T8_SC_LIBADD@: No such file or directory
```

On my local machine and on [Yggdrasil ](https://github.com/JuliaPackaging/Yggdrasil/pull/6565)all compiles just fine.

Have you seen this behavior before? Any clue what the problem is? Incompatible or old autotools version on GH?

jmark:
Ok, found the problem. The submodules for `sc` and `p4est` were not pointing to https://github.com/DLR-AMR/t8code/issues/483 and https://github.com/DLR-AMR/t8code/issues/496.

@cburstedde

jmark:
With this PR I could finally register an official [t8code_jll](https://juliahub.com/ui/Packages/t8code_jll/6PmXJ/1.1.2+2) Julia package. But this is only a temporary solution, since this builds on pre-releases and branches of sc, p4est and t8code. If this PR will be merged in the near future, the package could be build upon official releases of the three libraries.

cburstedde:
Once the latest develop branches of p4est and sc are made to master, these will have to be updated here to work together with t8code.

jmark:
This PR got obsolete since the proposed changes have been included in other PRs already. Many thanks @cburstedde for the input. This PR was nevertheless vital in the understanding of the issue. Gonna close this PR now.

PR #518, opened 2023-04-13
Feature puma patches [3/x] add element ref coords implementation

sandro-elsweijer:
The implementation of the actual element ref coord functions. The function gets a set of reference coordinates inside an element $[0, 1]^d$ and maps these coordinates to the reference coordinates of the tree $[0, 1]^d$, regarding the anchor coordinate, level and orientation. This way, we can evaluate the geometry of arbitrary points inside an element. Before, we could only evaluate the vertices of the elements.

### DIFF
--- a/src/t8_schemes/t8_default/t8_default_hex/t8_default_hex_cxx.cxx
+++ b/src/t8_schemes/t8_default/t8_default_hex/t8_default_hex_cxx.cxx
@@ -666,7 +666,23 @@ t8_default_scheme_hex_c::t8_element_reference_coords (const t8_element_t
                                                       double *out_coords)
   const
 {
-  SC_ABORTF ("Not implemented\n");
+  T8_ASSERT (t8_element_is_valid (elem));
+  const p8est_quadrant_t *q1 = (const p8est_quadrant_t *) elem;
+
+  /* Get the length of the quadrant */
+  const int           len = P8EST_QUADRANT_LEN (q1->level);
+
+  /* Compute the x, y and z coordinates of the point depending on the
+   * reference coordinates */
+  out_coords[0] = q1->x + ref_coords[0] * len;
+  out_coords[1] = q1->y + ref_coords[1] * len;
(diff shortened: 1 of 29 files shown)


src/t8_schemes/t8_default/t8_default_hex/t8_default_hex_cxx.cxx
```diff
@@ -666,7 +666,23 @@ t8_default_scheme_hex_c::t8_element_reference_coords (const t8_element_t
                                                       double *out_coords)
   const
 {
-  SC_ABORTF ("Not implemented\n");
+  T8_ASSERT (t8_element_is_valid (elem));
```
jmark, line 669: Please move this implementation to a C function `t8_dhex_compute_reference_coords`. For this you will have to introduce new files `t8_dhex_bits.{h,c}`, which is fine since we want to have them in the near future anyway.

src/t8_schemes/t8_default/t8_default_quad/t8_default_quad_cxx.cxx
```diff
@@ -799,7 +799,18 @@ t8_default_scheme_quad_c::t8_element_reference_coords (const t8_element_t
                                                        double *out_coords)
   const
 {
-  SC_ABORTF ("Not implemented\n");
+  T8_ASSERT (t8_element_is_valid (elem));
```
jmark, line 802: Please move this implementation to a C function `t8_dquad_compute_reference_coords`. For this you will have to introduce new files `t8_dquad_bits.{h,c}`, which is fine since we want to have them in the near future anyway.

jmark (changes requested):

src/t8_schemes/t8_default/t8_default_quad/t8_default_quad_cxx.cxx
sandro-elsweijer: Done

jmark (approved):

PR #555, opened 2023-05-12
Feature shmem allgatherv

Davknapp:
Implement an AllgatherV for a shared memory array. 

#### General

#### Tests

#### Github action

#### Scripts and Wiki

#### Licence

### DIFF
--- a/src/t8_data/t8_shmem.c
+++ b/src/t8_data/t8_shmem.c
@@ -310,6 +310,7 @@ t8_shmem_array_allgatherv_common (void *sendbuf,
   t8_compute_recvcounts_displs (intra_recv_total, inter_recvcount,
                                 inter_displ, sizeof (sendtype),
                                 internode_comm);
+
   if (t8_shmem_array_start_writing (recvarray)) {
     mpiret =
       sc_MPI_Allgatherv (noderecvchar, intra_recv_total, sendtype,
@@ -317,9 +318,6 @@ t8_shmem_array_allgatherv_common (void *sendbuf,
                          recvtype, internode_comm);
     SC_CHECK_MPI (mpiret);
     T8_FREE (noderecvchar);
-  }
-  else {
-    t8_errorf ("Can not write shmem-array.\n");
   }
   t8_shmem_array_end_writing (recvarray);
 


src/t8_data/t8_shmem.c
```diff
@@ -218,6 +218,201 @@ t8_shmem_array_allgather (const void *sendbuf, int sendcount,
...
+
+  mpiret = sc_MPI_Allgather ((void *) &sendcount, 1, sc_MPI_INT, recvcounts,
+                             1, sc_MPI_INT, comm);
+  SC_CHECK_MPI (mpiret);
+
+  int                 recv_total = recvcounts[0];
+  for (int i = 1; i < mpisize; i++) {
+    displs[i] = displs[i - 1] + recvcounts[i - 1];
+    recv_total += recvcounts[i];
+  }
```
sandro-elsweijer, line 246: Please use a more describing variable name here. Also a comment would be nice

src/t8_data/t8_shmem.c
```diff
@@ -218,6 +218,201 @@ t8_shmem_array_allgather (const void *sendbuf, int sendcount,
...
+ * Allgathers sendbuffer of various sizes and writes them in the recvarray. 
+ * Can be used for SC_SHMEM_WINDOW, SC_SHMEM_WINDOW_PRESCAN
+ * SC_SHMEM_BGQ or SC_SHMEMBGQ_PRESCAN (sc_shmem.h) 
+ * 
+ * \param[in] sendbuf   The sendbuffer
+ * \param[in] sendcount The number of items send by this proc
+ * \param[in] sendtype The type of items to send
+ * \param[in, out] recvarray The destination
+ * \param[in] recvtype The type of items to receive
+ * \param[in] comm The mpicommunicator to use.
```
sandro-elsweijer, line 262: Not all arguments are documented

src/t8_data/t8_shmem.h
```diff
@@ -147,6 +147,32 @@ void                t8_shmem_array_allgather (const void *sendbuf,
                                               int recvcount,
                                               sc_MPI_Datatype recvtype);
 
+/**
+ * Fill a t8_shmem array with an Allgatherv
+ * Computes the recvcount-array and displacement-array for each rank of a node using the
+ * sendcount.
+ * The total number of items of each node is then used to compute the
+ * recvcount-array and displacement-array between nodes.
```
sandro-elsweijer, line 155: Since `MPI_Allgatherv` does not scale well in HPC (and t8code is especially written for the HPC context), we should add a disclaimer to use `MPI_Allgather` whenever possible.

src/t8_data/t8_shmem.c
```diff
@@ -218,6 +218,201 @@ t8_shmem_array_allgather (const void *sendbuf, int sendcount,
...
+ * \param[in] comm The mpicommunicator to use. 
+ */
+static void
+t8_shmem_array_allgatherv_common (void *sendbuf,
+                                  const int sendcount,
+                                  sc_MPI_Datatype sendtype,
+                                  t8_shmem_array_t recvarray,
+                                  sc_MPI_Datatype recvtype,
+                                  sc_MPI_Comm comm, sc_MPI_Comm intranode,
+                                  sc_MPI_Comm internode)
```
sandro-elsweijer, line 271: I would call these `intranode_comm` and `internode_comm`. While reading the function I am always forgetting, that `intranode` and `internode` are communicators

src/t8_data/t8_shmem.c
```diff
@@ -218,6 +218,201 @@ t8_shmem_array_allgather (const void *sendbuf, int sendcount,
...
+  mpiret = sc_MPI_Comm_size (internode, &intersize);
+  SC_CHECK_MPI (mpiret);
+
+  /* intranode-gatherv */
+  int                *intra_displ = T8_ALLOC_ZERO (int, intrasize);
+  int                *intra_recvcounts = T8_ALLOC_ZERO (int, intrasize);
+  int                 intra_recv_total =
+    t8_compute_recvcounts_displs (sendcount, intra_recvcounts, intra_displ,
+                                  sizeof (sendtype), intranode);
+  if (!intrarank) {
```
sandro-elsweijer, line 292:
```suggestion
  if (intrarank == 0) {
```
I know its the same, but it makes it more obvoius, that this condition is only true for the intranode main rank

src/t8_data/t8_shmem.c
```diff
@@ -218,6 +218,201 @@ t8_shmem_array_allgather (const void *sendbuf, int sendcount,
...
+    sc_MPI_Gatherv (sendbuf, sendcount, sendtype, noderecvchar,
+                    intra_recvcounts, intra_displ, recvtype, 0, intranode);
+  SC_CHECK_MPI (mpiret);
+
+/* internode-allgatherv */
+  int                *inter_displ = T8_ALLOC_ZERO (int, intersize);
+  int                *inter_recvcount = T8_ALLOC_ZERO (int, intersize);
+  t8_compute_recvcounts_displs (intra_recv_total, inter_recvcount,
+                                inter_displ, sizeof (sendtype), internode);
+  if (t8_shmem_array_start_writing (recvarray)) {
```
sandro-elsweijer, line 305: Shouldn't we throw an error, if we cannot write into the recvarray?

src/t8_data/t8_shmem.c
```diff
@@ -218,6 +218,201 @@ t8_shmem_array_allgather (const void *sendbuf, int sendcount,
...
+  SC_CHECK_MPI (mpiret);
+
+  /* intranode-gatherv */
+  int                *intra_displ = T8_ALLOC_ZERO (int, intrasize);
+  int                *intra_recvcounts = T8_ALLOC_ZERO (int, intrasize);
+  int                 intra_recv_total =
+    t8_compute_recvcounts_displs (sendcount, intra_recvcounts, intra_displ,
+                                  sizeof (sendtype), intranode);
+  if (!intrarank) {
+    noderecvchar = T8_ALLOC (char, intra_recv_total * typesize);
```
sandro-elsweijer, line 293: Why are you allocating `intra_recv_total * typesize` times a `char` and not `intra_recv_total` times a `type`? This makes this less readable. Is it, because you cannot convert a `sc_MPI_Datatype` to a normal datatype? If yes, a comment would be nice :)

src/t8_data/t8_shmem.c
```diff
@@ -218,6 +218,201 @@ t8_shmem_array_allgather (const void *sendbuf, int sendcount,
                       recvcount, recvtype, recvarray->comm);
 }
 
+/**
+ * Computes the recvcounter- and displacementarray for 
+ * an MPI_Gatherv
```
sandro-elsweijer, line 223:
```suggestion
 * Computes the recvcounter and displacement array for 
 * an MPI_Gatherv
```

test/t8_data/t8_gtest_shmem.cxx
```diff
@@ -165,6 +165,76 @@ TEST_P(shmem, test_sc_shmem_alloc){
   }
 }
 
+TEST_P(shmem, test_shmem_array_allgatherv){
+  const size_t           element_size = sizeof (t8_gloidx_t);
+  int                 mpirank, mpisize;
```
sandro-elsweijer, line 170: Coding guidelines

test/t8_data/t8_gtest_shmem.cxx
```diff
@@ -165,6 +165,76 @@ TEST_P(shmem, test_sc_shmem_alloc){
...
+  mpiret = sc_MPI_Comm_size (comm, &mpisize);
+  SC_CHECK_MPI (mpiret);
+
+  t8_debugf("[D] mpirank: %i\n", mpirank);
+  
+  const int                   base_size = 10;
+  const t8_gloidx_t           array_length = (mpirank+1) * base_size;
+  t8_gloidx_t                 *sendbuf = T8_ALLOC_ZERO(t8_gloidx_t, array_length);
+  const t8_gloidx_t           first_array_value = base_size * mpirank * (mpirank + 1) / 2;
+  const int                   total_size = base_size * mpisize * (mpisize + 1) / 2;
```
sandro-elsweijer, line 184: Please add some comments here

src/t8_data/t8_shmem.c
sandro-elsweijer, line 293: Why are you allocating `intra_recv_total * typesize` times a `char` and not `intra_recv_total` times a `type`? This makes this less readable. Is it, because you cannot convert a `sc_MPI_Datatype` to a normal datatype? If yes, a comment would be nice :)

sandro-elsweijer (changes requested):
I haven't reviewed the test fully yet, since I have some questions about the implementation. We should have a call about that 😉

src/t8_data/t8_shmem.h
Davknapp: I don't agree with that. The name of the function is similar to the MPI-name and the user should be aware of the scaling behaviour. You're right, All-to-all communication does not scale well, but sometime we can't avoid it. Nevertheless we don't warn the user about it.

src/t8_data/t8_shmem.c
Davknapp: `sc_MPI_Datatype` is defined as `MPI_Datatype` which is a typedef of an `int`. But we need the actual size of the elements.

Davknapp: See comment above.

src/t8_data/t8_shmem.c
sandro-elsweijer: `sc_MPI_Comm intranode_comm` and `sc_MPI_Comm internode_comm` are still undocumented. Please only resolve conversations, that are really resolved ;)

src/t8_data/t8_shmem.c
sandro-elsweijer: Is an error message without throwing an error enough? This can easily be overseen, if the console output is long

sandro-elsweijer (changes requested):

src/t8_data/t8_shmem.c
Davknapp: All of the functions are void functions, so we can't pass integer codes down. Furthermore there isn't something like an  try-throw error handling in t8code. We could either improve the error handling overall in the shmem functions, or leave it as it is. But an improvement should be done outside of this PR.

src/t8_data/t8_shmem.h
Davknapp: The negative consequence would be to comment on all functions of t8code how they scale. We don't warn the user at the function `t8_shmem_array_allgather' about the scaling and I don't think we should here either.

sandro-elsweijer (commented):
Just minor comments on your comments.

PR #596, opened 2023-06-27
Fix issue 142 - t8_face_leaf_neighbors

### LINKED ISSUE #142: Split up t8_forest_leaf_face_neighbors
t8_forest_leaf_face_neighbors provides a lot of functionality.
Sometimes not all of this is required but the caller still needs to provide allocated memory. There should be subfunctions for

- only get the number of face neighbors
- set output vars to NULL lets the function ignore them

jfussbro:

Before, t8_forest_leaf_face_neighbors dynamically allocated and freed memory for the leaf_face_neighbors for every face. Now, the caller of the function is responsible for the allocation of the memory and passes pointers to the allocated arrays into the function.
If the caller passes null-pointers it is considert, that information about leaf_face_neighbors is not needed and they will not be calculated.

Closes #142 

#### General

#### Tests

#### Github action

#### Scripts and Wiki

#### Licence


Davknapp:
All tests are failing, please have another look at it ;)

jmark:
@jfussbro Would be please revive this PR?

Please merge the latest `main`.

Furthermore, please create a wrapper function which emulates the old signature. That is the wrapper allocates the arrays and passes them to your modified function. We want to keep backwards compatibility for now.

I suggest to call the wrapper function like the old version and the modified (without array allocations) gets another name. E.g.,
`t8_forest_leaf_face_neighbors_ext`.

We will add the query for the `orientation` in the `*_ext` version, too. But one step at the time.

holke:
Thanks for this change.

Some remarks:

If the caller is responsible for allocating the memory, we should provide a function that does so.
i.e. `t8_leaf_face_neighbor_alloc/free` or similar.
Also the documentation must be very clear here since letting the user manage the memory can lead to trip falls.
I.e user allocates memory for an element with 4 face neighbors and uses the same memory for the next element but it has 6 neighbors but user does not remember to allocate new memory.

The user must be aware that this can happen and what they can do to prevent.

Additionally, #142 suggest splitting up the function into subfunction. One of them calculating the number of neighbors.

jmark:
> Thanks for this change.
> 
> Some remarks:
> 
> If the caller is responsible for allocating the memory, we should provide a function that does so. i.e. `t8_leaf_face_neighbor_alloc/free` or similar. Also the documentation must be very clear here since letting the user manage the memory can lead to trip falls. I.e user allocates memory for an element with 4 face neighbors and uses the same memory for the next element but it has 6 neighbors but user does not remember to allocate new memory.
> 
> The user must be aware that this can happen and what they can do to prevent.
> 
> Additionally, #142 suggest splitting up the function into subfunction. One of them calculating the number of neighbors.

Thanks for the thoughts. I actually thought that a user calling this routine can even use static arrays being part of the function stack. The arrays just have to be large enough. I suggest to simply use constants ala `T8_MAX_NUM_FACE_CHILDREN` or similar.

That way the user has full control and exactly knows where and when memory is allocated.

jmark:
@jfussbro As we discussed I close this PR.

PR #633, opened 2023-07-19
Disable non-applicable workflows on forks

sloede:

Right now, all scheduled workflows run unconditionally in all forks. While this is likely neither required nor intended, it is also annoying because most workflows seem to fail when executed on a fork, e.g., [this one](https://github.com/sloede/t8code/actions/runs/5594638751/jobs/10229692683), and thus cause a number of failure notifications from GitHub each night.

This PR adds a guard to all workflows with a `schedule` trigger that ensures that `schedule` only triggers a workflow on the main repository.

For good measure, I also disabled the Mattermost workflow on forks, since it also does not seem to be applicable, always fails, and is triggered by things like syncing your fork with the upstream repo.

#### General

#### Tests

#### Github action

#### Scripts and Wiki

#### Licence

### DIFF
--- a/.github/workflows/tests_t8code_parallel.yml
+++ b/.github/workflows/tests_t8code_parallel.yml
@@ -272,3 +272,4 @@ jobs:
       with:
         name: test-suite_release_MPI_cxx.log
         path: build_release_cpp/test-suite.log
+
--- a/.github/workflows/tests_t8code_parallel.yml
+++ b/.github/workflows/tests_t8code_parallel.yml
@@ -49,6 +49,7 @@ on:
 jobs:
   build:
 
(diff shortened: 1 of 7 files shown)


.github/workflows/tests_t8code_parallel.yml
```diff
@@ -270,4 +271,4 @@ jobs:
       uses: actions/upload-artifact@v3
       with:
         name: test-suite_release_MPI_cxx.log
-        path: build_release_cpp/test-suite.log
```
sandro-elsweijer, line 273: Can you please add the line break at the end?

sandro-elsweijer (changes requested):
Apart from the one comment: Lgtm

.github/workflows/tests_t8code_parallel.yml
sloede: Fixed in 3c580c97.

PR #635, opened 2023-07-19
Move t8_cmesh_set_tree_vertices

sandro-elsweijer:
The `t8_cmesh_set_tree_vertices` function was in the `t8_cmesh_vtk_writer.h` header and it made no sense. Based on a Mattermost discussion the according changes were made. There is no new code in this PR, I only moved things around.

Since the vertices of a tree are only needed, if a `t8_geometry_w_vertices` geometry is used, the function is now part of this header. Therefore, it is automatically loaded when a `t8_geometry_w_vertices` or derived geometry is used. This should also enshure, that all external applications which link against t8code do not have to be adapted (apart from that in some cases the `t8_cmesh_vtk_writer.h` header does not have to be included anymore).

#### General

#### Tests

#### Github action

#### Scripts and Wiki

#### Licence


jmark (commented):
@sandro-elsweijer What is the meaning of "_w_" in all these names? Is it an abreviation for "_with_" or anything else? I think it wouldn't hurt to spell it out. This makes the intent of the code much clearer.

sandro-elsweijer:
> @sandro-elsweijer What is the meaning of `_w_` in all these names? Is it an abreviation for `_with_` or anything else? I think it wouldn't hurt to spell it out. This makes the intent of the code much clearer.

Yes you are right, this can be confusing. Changed that

jmark (approved):

PR #733, opened 2023-09-01
Feature batch processing t8 forest element point inside

Davknapp:
Implemented a batched version of the function t8_forest_element_point_inside. 
With the updated implementation we avoid computations that prepare the actuall inside-check for each point check. 
instead this preparing step is done once and then all points are checked. 

#### General

#### Tests

#### Github action

#### Scripts and Wiki

#### Licence

### DIFF
--- a/src/t8_forest/t8_forest_cxx.cxx
+++ b/src/t8_forest/t8_forest_cxx.cxx
@@ -1145,7 +1145,7 @@ t8_line_point_inside (const double *p_0, const double *vec, const double *point,
  * 
  * \param[in] p_0         The first vertex of a triangle
  * \param[in] v           The vector from p_0 to p_1 (second vertex in the triangle)
- * \param[in] w           The vector from p_0 to p_2 (second vertex in the triangle)
+ * \param[in] w           The vector from p_0 to p_2 (third vertex in the triangle)
  * \param[in] point       The coordinates of the point to check
  * \param[in] tolerance   A double > 0 defining the tolerance
  * \return                0 if the point is outside, 1 otherwise.  
--- a/src/t8_forest/t8_forest_cxx.cxx
+++ b/src/t8_forest/t8_forest_cxx.cxx
@@ -1209,7 +1209,7 @@ t8_triangle_point_inside (const double p_0[3], const double v[3], const double w
  * \param[in] point_on_face   A point on the plane
  * \param[in] face_normal     The normal of the face
  * \param[in] point           The point to check
- * \return                0 if the point is outside, 1 otherwise.                   
+ * \return                    0 if the point is outside, 1 otherwise.                   
  */
 static int
 t8_plane_point_inside (const double point_on_face[3], const double face_normal[3], const double point[3])
@@ -1226,9 +1226,9 @@ t8_plane_point_inside (const double point_on_face[3], const double face_normal[3
   return 1;
 }
 
-int
-t8_forest_element_point_inside (t8_forest_t forest, t8_locidx_t ltreeid, const t8_element_t *element,
-                                const double *points, const double tolerance)
+void
+t8_forest_element_point_batch_inside (t8_forest_t forest, t8_locidx_t ltreeid, const t8_element_t *element,
+                                      const double *points, int num_points, int *is_inside, const double tolerance)
 {
   const t8_eclass_t tree_class = t8_forest_get_tree_class (forest, ltreeid);
   t8_eclass_scheme_c *ts = t8_forest_get_eclass_scheme (forest, tree_class);
@@ -1251,7 +1251,10 @@ t8_forest_element_point_inside (t8_forest_t forest, t8_locidx_t ltreeid, const t
     t8_forest_element_coordinate (forest, ltreeid, element, 0, vertex_coords);
     /* Check whether the point and the vertex are within tolerance distance
        * to each other */
-    return t8_vertex_point_inside (vertex_coords, points, tolerance);
+    for (int ipoint = 0; ipoint < num_points; ipoint++) {
+      is_inside[ipoint] = t8_vertex_point_inside (vertex_coords, &points[ipoint * 3], tolerance);
+    }
+    return;
   }
   case T8_ECLASS_LINE: {
     /* A point p is inside a line that is defined by the edge nodes
@@ -1268,13 +1271,14 @@ t8_forest_element_point_inside (t8_forest_t forest, t8_locidx_t ltreeid, const t
     t8_forest_element_coordinate (forest, ltreeid, element, 1, v);
     /* v = p_1 - p_0 */
     t8_vec_axpy (p_0, v, -1);
-
-    return t8_line_point_inside (p_0, v, points, tolerance);
+    for (int ipoint = 0; ipoint < num_points; ipoint++) {
+      is_inside[ipoint] = t8_line_point_inside (p_0, v, &points[ipoint * 3], tolerance);
+    }
+    return;
   }
   case T8_ECLASS_QUAD: {
     /* We divide the quad in two triangles and use the triangle check. */
     double p_0[3], p_1[3], p_2[3], p_3[3];
-    int point_inside = 0;
     /* Compute the vertex coordinates of the quad */
     t8_forest_element_coordinate (forest, ltreeid, element, 0, p_0);
     t8_forest_element_coordinate (forest, ltreeid, element, 1, p_1);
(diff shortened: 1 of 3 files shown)


src/t8_forest/t8_forest_cxx.cxx
```diff
@@ -1065,14 +1065,95 @@ t8_forest_element_face_normal (t8_forest_t forest, t8_locidx_t ltreeid, const t8
...
+  }
+  /* The point is on the line. */
+  return 1;
+}
+
+/**
+ * Check if a point is inside of a triangle described by a point \a p_0 and two vectors \a v and \a w. 
+ * 
+ * \param[in] p_0         The first vertex of a triangle
+ * \param[in] v           The vector from p_0 to p_1 (second vertex in the triangle)
```
sandro-elsweijer, line 1147: Do we count from zero or one here?
```suggestion
 * \param[in] v           The vector from p_0 to p_1 (first vertex in the triangle)
```

src/t8_forest/t8_forest_cxx.cxx
```diff
@@ -1065,14 +1065,95 @@ t8_forest_element_face_normal (t8_forest_t forest, t8_locidx_t ltreeid, const t8
...
+ * \param[in] point       The coordinates of the point to check
+ * \param[in] tolerance   A double > 0 defining the tolerance
+ * \return                0 if the point is outside, 1 otherwise.  
+ */
+static int
+t8_triangle_point_inside (const double p_0[3], const double v[3], const double w[3], const double point[3],
+                          const double tolerance)
+{
   /* A point p is inside the triangle that is spanned
    * by the vectors p_0 p_1 p_2 if and only if the linear system
```
sandro-elsweijer, line 1158:
```suggestion
   * by the point p_0 and vectors v and w if and only if the linear system
```

src/t8_forest/t8_forest_cxx.cxx
```diff
@@ -1065,14 +1065,95 @@ t8_forest_element_face_normal (t8_forest_t forest, t8_locidx_t ltreeid, const t8
...
    * (p_1 - p_0)x + (p_2 - p_0)y = p - p_0
```
sandro-elsweijer, line 1159:
```suggestion
   * vx + wy = point - p_0
```

src/t8_forest/t8_forest_cxx.cxx
```diff
@@ -1083,15 +1164,9 @@ t8_triangle_point_inside (const double p_0[3], const double p_1[3], const double
...
-  double b[3], v[3], w[3];
-  double x, y, z;
-
-  /* v = v - p_0 = p_1 - p_0 */
-  t8_vec_axpyz (p_0, p_1, v, -1);
-  /* w = w - p_0 = p_2 - p_0 */
-  t8_vec_axpyz (p_0, p_2, w, -1);
+  T8_ASSERT (tolerance > 0); /* negative values and zero are not allowed */
+  double b[3];
   /* b = p - p_0 */
```
sandro-elsweijer, line 1170:
```suggestion
  /* b = point - p_0 */
```

src/t8_forest/t8_forest_cxx.cxx
```diff
@@ -1131,19 +1204,35 @@ t8_triangle_point_inside (const double p_0[3], const double p_1[3], const double
   return 1;
 }
 
-int
-t8_forest_element_point_inside (t8_forest_t forest, t8_locidx_t ltreeid, const t8_element_t *element,
-                                const double point[3], const double tolerance)
+/** Check if a point lays on the inner side of a plane of bilinearly interpolated volume element. 
+ * the plane is described by a point an the normal of the face.
```
sandro-elsweijer, line 1208:
```suggestion
/** Check if a point lays on the inner side of a plane of a bilinearly interpolated volume element. 
 * the plane is described by a point and the normal of the face. 
```

src/t8_forest/t8_forest_cxx.cxx
```diff
@@ -1131,19 +1204,35 @@ t8_triangle_point_inside (const double p_0[3], const double p_1[3], const double
...
+static int
+t8_plane_point_inside (const double point_on_face[3], const double face_normal[3], const double point[3])
+{
+  /* Set x = x - p */
+  double pof[3] = { point_on_face[0], point_on_face[1], point_on_face[2] };
+  t8_vec_axpy (point, pof, -1);
+  /* Compute <x-p,n> */
+  const double dot_product = t8_vec_dot (pof, face_normal);
+  if (dot_product < 0) {
+    /* The point is outside of the plane */
```
sandro-elsweijer, line 1223:
```suggestion
    /* The point is on the wrong side of the plane */
```

src/t8_forest/t8_forest_cxx.cxx
```diff
@@ -1278,29 +1352,43 @@ t8_forest_element_point_inside (t8_forest_t forest, t8_locidx_t ltreeid, const t
      *  <x - p, n> >= 0
```
sandro-elsweijer, line 1352: I cannot comment on line 1346 directly, but the formatting is weird in that comment

src/t8_forest/t8_forest_general.h
```diff
@@ -761,6 +761,28 @@ int
...
+ * \param [in]      points      3-dimensional coordinates of the points to check
+ * \param [in]      num_points  The number of points to check
+ * \param [in, out] is_inside   An array of length num_points, filled with 0/1 on output. True (non-zero) if \a point 
+ *                              lies within \a element, false otherwise. The return value is also true if the point 
+ *                              lies on the element boundary. Thus, this function may return true for different leaf 
+ *                              elements, if they are neighbors and the point lies on the common boundary.
+ * \param [in]      tolerance   tolerance that we allow the point to not exactly match the element.
+ *                              If this value is larger we detect more points.
+ *                              If it is zero we probably do not detect points even if they are inside
+ *                              due to rounding errors.
```
sandro-elsweijer, line 780: `\a` will not put an "a" into the documentation. "a" has to be set individually
```suggestion
/** Query whether a batch of points lies inside an element. For bilinearly interpolated elements.
 * \note For 2D quadrilateral elements this function is only an approximation. It is correct
 *  if the four vertices lie in the same plane, but it may produce only approximate results if 
 *  the vertices do not lie in the same plane.
 * \param [in]      forest      The forest.
 * \param [in]      ltree_id    The forest local id of the tree in which the element is.
 * \param [in]      element     The element.
 * \param [in]      points      3-dimensional coordinates of the points to check
 * \param [in]      num_points  The number of points to check
 * \param [in, out] is_inside   An array of length \a num_points, filled with 0/1 on output. True (non-zero) if a \a point 
 *                              lies within an \a element, false otherwise. The return value is also true if the point 
 *                              lies on the element boundary. Thus, this function may return true for different leaf 
 *                              elements, if they are neighbors and the point lies on the common boundary.
 * \param [in]      tolerance   Tolerance that we allow the point to not exactly match the element.
 *                              If this value is larger we detect more points.
 *                              If it is zero we probably do not detect points even if they are inside
 *                              due to rounding errors.
```

sandro-elsweijer (changes requested):

src/t8_forest/t8_forest_cxx.cxx
```diff
@@ -1065,14 +1065,95 @@ t8_forest_element_face_normal (t8_forest_t forest, t8_locidx_t ltreeid, const t8
...
+  * if any v[i] is not 0.
+  *
+  * Otherwise the line is degenerated (which should not happen).
+  */
+  for (i = 0; i < 3; ++i) {
+    if (vec[i] != 0) {
+      x = b[i] / vec[i];
+      break; /* found a non-zero coordinate. We can stop now. */
+    }
+  }
```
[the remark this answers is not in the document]
Davknapp: But that wouldn't hurt. If vec is zero the computation divides by zero, if b is zero a zero is divided by something which is ok. Or did  I miss something?

src/t8_forest/t8_forest_cxx.cxx
Davknapp: No, p_0 is the first vertex in the triangle, p_1 the second and p_2 the third.

src/t8_forest/t8_forest_cxx.cxx
Davknapp: Updated the format.

src/t8_forest/t8_forest_cxx.cxx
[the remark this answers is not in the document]
sandro-elsweijer: Oh yes you are right

src/t8_forest/t8_forest_cxx.cxx
```diff
@@ -1065,34 +1065,109 @@ t8_forest_element_face_normal (t8_forest_t forest, t8_locidx_t ltreeid, const t8
...
+  /* The point is on the line. */
+  return 1;
+}
+
+/**
+ * Check if a point is inside of a triangle described by a point \a p_0 and two vectors \a v and \a w. 
+ * 
+ * \param[in] p_0         The first vertex of a triangle
+ * \param[in] v           The vector from p_0 to p_1 (second vertex in the triangle)
+ * \param[in] w           The vector from p_0 to p_2 (second vertex in the triangle)
```
sandro-elsweijer, line 1148:
```suggestion
 * \param[in] w           The vector from p_0 to p_2 (third vertex in the triangle)
```

sandro-elsweijer (changes requested):

src/t8_forest/t8_forest_cxx.cxx
Davknapp: Ahh, copy-pasta^^

sandro-elsweijer:
Converted this to a draft to prevent accidental merging, before the base PR is merged

sandro-elsweijer (approved):

Issue #33, opened 2020-05-20
Implement partition for coarsening


holke:
If the appropriate flag is set in forest partitioning, families should not get seperated by process boundaries, thus ensuring that coarsening one level is a process local operation.

This needs changing the first and last elements on a process such that only whole families land on a single process after the partition operation.

lukasdreyer:
@holke 
Do you have an idea how to compute the new offsets after computing the standard offsets, without sending the elements first?
All my ideas boil down to essentially this (with different possibilities on how to do the adjustements): 

Do the computation of the standard offsets, send the elements, then adjust the number of elements on a process slightly (considering edgecases like empty ranks), compute new offsets and finally send some elements again according to these new offsets.

holke:
Basically you only need to ensure that the first element of each process has child Id 0.
This should be computable without communication.

PR #772, opened 2023-09-26
Feature Partition For Coarsening

Niklas997:

Implemented the "Partition For Coarsening" feature. This feature is controlled by a flag during the function call of `t8_forest_set_partition(...)`. Previously, the value of the flag was ignored.

During a "default" set_partition-call (with the flag `set_for_coarsening = 0`), the partition is chosen, such that each rank will hold the same amount of elements (+/-1). However, element-families are not taken into account and may end up on different ranks due to this strict load-balancing. This prevents the family from being coarsened (since the family is not passed as a whole to the adapt-callback). If the flag is set to "true" (`set_for_coarsening = 1`), the default partition-boundaries are adjusted such that no family is ripped apart. In order to achieve this, each partition-boundary is shifted by at most `+/- (#Number_of_Siblings / 2)`  elements resulting in a maximum load imbalance of `#Number_of_Siblings + 1` between different ranks.

An example has been implemented displaying the effect of this partitioning feature in comparison to the "default" partitioning scheme. Tests for this core-utlility will be supplied as a follow-up PR if this one will be accepted.

#### General

#### Tests

#### Github action

#### Scripts and Wiki

#### Licence

### DIFF
--- a/example/forest/Makefile.am
+++ b/example/forest/Makefile.am
@@ -6,9 +6,11 @@ bin_PROGRAMS += \
 	example/forest/t8_face_neighbor \
 	example/forest/t8_test_ghost \
 	example/forest/t8_test_face_iterate \
-	example/forest/t8_test_ghost_large_level_diff
+	example/forest/t8_test_ghost_large_level_diff \
+	example/forest/t8_forest_partition_for_coarsening
 
 example_forest_t8_face_neighbor_SOURCES = example/forest/t8_face_neighbor.cxx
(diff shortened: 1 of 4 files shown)


lukasdreyer:
Maybe test this with meshes with holes

lukasdreyer:
closes #33

jmark:
@Niklas997 @lukasdreyer 
What is the current status of this PR? Any new ideas or is it feature-complete and polished now? Who will do the final review?

PR #803, opened 2023-10-25
Test for element ref coords

### LINKED ISSUE #777: Write a test for t8_forest_element_centroid

sandro-elsweijer:
Adds a test for the element ref coord computation and fixes several bugs. Closes #777 
Reimplements the new centroid computation, which was removed due to a bug.

#### General

#### Tests

#### Github action

#### Scripts and Wiki

#### Licence


sandro-elsweijer:
This is ready for review, but #805 applies

test/t8_schemes/t8_gtest_element_ref_coords.cxx
```diff
@@ -0,0 +1,184 @@
...
+  buffer.setf (std::ios::fixed);
+  buffer << message;
+  for (int i_dim = 0; i_dim < dim; ++i_dim) {
+    buffer << " " << array[i_dim];
+  }
+  t8_debugf ("%s\n", buffer.str ().c_str ());
+}
+
+/* To save an iterate, we will test the ref coords and centroid right in the adapt callback itself. 
+ * Refines as long as the MAXLEVEL is not reached. */
```
Davknapp, line 95: I think this optimization is a little bit to much. If we ever encounter a bug in the ref_coords computation again it will make debugging more complicated. The purpose of the tests are not to run the most optimal way, but to ensure all functionalities of t8code and to make our lifes easier, when we hunt for bugs. 
Seperating adaptation and testing will make it easier.

src/t8_element.c
```diff
@@ -55,10 +55,10 @@ const double
         { 1, 1, 1 } },
       { /* T8_ECLASS_PRISM */
         { 0, 0, 0 },
```
Davknapp, line 57: This look-up-table is undocumented. Please add documentation

src/t8_element.c
```diff
@@ -74,7 +74,7 @@ const double t8_element_centroid_ref_coords[T8_ECLASS_COUNT][3] = {
   { 2. / 3., 1. / 3., 0 },   /* T8_ECLASS_TRIANGLE */
   { 0.5, 0.5, 0.5 },         /* T8_ECLASS_HEX */
   { 0.75, 0.25, 0.5 },       /* T8_ECLASS_TET */
```
Davknapp, line 76: This look-up-table is undocumented. Please add documentation

src/t8_schemes/t8_default/t8_default_prism/t8_dprism_bits.c
```diff
@@ -538,7 +538,7 @@ t8_dprism_vertex_coords (const t8_dprism_t *elem, const int vertex, int coords[3
   T8_ASSERT (elem->line.level == elem->tri.level);
   /*Compute x and y coordinate */
   t8_dtri_compute_coords (&elem->tri, vertex % 3, coords);
-  /*Compute z coordinatecoords[0] *= T8_DPRISM_ROOT_BY_DTRI_ROOT; */
+  /*Compute z coordinate coords[0] *= T8_DPRISM_ROOT_BY_DTRI_ROOT; */
```
Davknapp, line 541:
```suggestion
  /* Compute z coordinate coords[0] *= T8_DPRISM_ROOT_BY_DTRI_ROOT; */
```

test/t8_schemes/t8_gtest_element_ref_coords.cxx
```diff
@@ -0,0 +1,184 @@
...
+  const int elem_dim = t8_eclass_to_dimension[shape];
+  t8_get_batch_coords_for_element_type (shape, batch_coords);
+  t8_debugf ("Testing for shape %s\n", t8_eclass_to_string[shape]);
+  t8_debugf ("with num_vertices %i\n", num_vertices);
+  t8_debugf ("and elem_dim %i\n", elem_dim);
+
+  ts->t8_element_reference_coords (elements[0], batch_coords, num_vertices, tree_ref_coords_by_element_ref_coords);
+  for (int i_vertex = 0; i_vertex < num_vertices; ++i_vertex) {
+    ts->t8_element_vertex_reference_coords (elements[0], i_vertex, tree_ref_coords_by_vertex);
+    t8_debugf ("Checking vertex %i\n", i_vertex);
```
Davknapp, line 126: Can you add this message via `<<` to the `EXPECT_NEAR`. That way this is printed only for failing tests.

Davknapp (changes requested):
Thanks for the test!

test/t8_schemes/t8_gtest_element_ref_coords.cxx
sandro-elsweijer: Yes, but since the adapt callback just returns 1, there will be no bugs in the adaptation itself. I see no advantage in using an adapt callback which just returns 1 and another iterate, which would itself add more complexity.

test/t8_schemes/t8_gtest_element_ref_coords.cxx
sandro-elsweijer: I could, but the messages print all coordinates and the `EXPECT_NEAR` only checks one dimension at a time. If we append the message, and multiple coords fail, the same message gets printed multiple times. Moreover, the logs of passing tests is not included in the overall log file, so it should not overcrowd the log files

src/t8_element.c
sandro-elsweijer: done

src/t8_element.c
sandro-elsweijer: done

test/t8_schemes/t8_gtest_element_ref_coords.cxx
Davknapp: We  just don't know the error that will make it crash yet. 
It easiert to debug a function that does only one thing, than a function that does two things. 
What kind of complexity do you mean? The runtime complexity, or the ccomplexity to understand the code?

test/t8_schemes/t8_gtest_element_ref_coords.cxx
Davknapp: Yes, you're right.

Davknapp (changes requested):
Thanks for the update, but please seperate the testing from the adaptation-callback.

test/t8_schemes/t8_gtest_element_ref_coords.cxx
Davknapp: If your adaptation returns 1, you actually compute a uniform forest on each level. For that purpose it is better to use t8_forest_new_uniform (as you don't have any data that is passed from the original to the refined forest).

sandro-elsweijer: Changed that :)

test/t8_schemes/t8_gtest_element_ref_coords.cxx
```diff
@@ -0,0 +1,189 @@
...
+  t8_eclass_t eclass;
+  t8_cmesh_t cmesh;
+  t8_forest_t forest, forest_partition;
+};
+
+TEST_P (class_ref_coords, t8_check_elem_ref_coords)
+{
+  t8_locidx_t itree, ielement;
+  /* Generate a uniform forest for each level up to MAXLEVEL and partition it */
+  for (int level = 0; level <= MAXLEVEL; ++level) {
```
Davknapp, line 167: From 0 to MaxLevel is actually a big range. How much time does the test need?

Can you add level as a parameter to the testsuite? Then you can move the construction of the forest also into the constructor of the class.

Davknapp (changes requested):
Please use the level as a parameter for the testsuite. That way you can also move more code into the constructor of the class.

test/t8_schemes/t8_gtest_element_ref_coords.cxx
Davknapp: Ah, you can ignore my comment. I thaught this was the maximal allowed level for that element. Can you rename it to something like `MAX_LEVEL_REF_COORD_TEST`

sandro-elsweijer: Done

Davknapp: That was ambiguous. I still think making the level a parameter of the testsuite is a good idea ;)

Sorry for not being clear about it.

sandro-elsweijer: Done

test/t8_schemes/t8_gtest_element_ref_coords.cxx
```diff
@@ -34,11 +34,41 @@ along with t8code; if not, write to the Free Software Foundation, Inc.,
...
 #if T8_ENABLE_LESS_TESTS
-#define MAXLEVEL 4
+#define MAX_LEVEL_REF_COORD_TEST 3
 #else
-#define MAXLEVEL 5
+#define MAX_LEVEL_REF_COORD_TEST 4
 #endif
 
+/**
+ * Writes a the contents of an array into a string.
```
Davknapp, line 43:
```suggestion
 * Writes the contents of an array into a string.
```

Davknapp (approved):

PR #831, opened 2023-11-09
Enforce c++ 17

lukasdreyer:
Enforce C++17

Uses https://www.gnu.org/software/autoconf-archive/ax_cxx_compile_stdcxx.html

Also limits the number of processes for the make check command in the github actions workflow to 8, because leaving it unlimited lead to weird racing conditions during compiling in the docker container. This could not be replicated when manually trying to repeat the build process in the docker container.

#### General

#### Tests

#### Github action

#### Scripts and Wiki

#### Licence

### DIFF
--- a/.github/workflows/tests_t8code_linkage_parallel_debug.yml
+++ b/.github/workflows/tests_t8code_linkage_parallel_debug.yml
@@ -30,6 +30,8 @@ name: t8code tests linkage parallel debug
 #       in the "key:" entries of the sc and p4est steps or set the variables
 #       SC_IGNORE_CACHE and P4EST_IGNORE_CACHE to 1 in the respective steps.
 
+env:
+  MAKEFLAGS: "-j2 V=0"
 
 on:
   push:
@@ -110,7 +112,7 @@ jobs:
       run: cd $SC_DEBUG && ../configure --enable-mpi --enable-debug --prefix=$PWD/install
       if: ${{ steps.sc_cache.outputs.cache-hit != 'true' || env.SC_IGNORE_CACHE == 1 }}
     - name: sc build debug
-      run: cd $SC_DEBUG && make -j V=0 && make -j V=0 install
+      run: cd $SC_DEBUG && make $MAKEFLAGS && make $MAKEFLAGS install
       if: ${{ steps.sc_cache.outputs.cache-hit != 'true' || env.SC_IGNORE_CACHE == 1 }}
 #
 # P4EST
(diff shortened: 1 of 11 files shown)


config/t8_include.m4
```diff
@@ -50,8 +50,9 @@ AC_DEFUN([T8_CHECK_LIBRARIES],
 T8_CHECK_NETCDF([$1])
 T8_CHECK_VTK([$1])
 T8_CHECK_OCC([$1])
-T8_CHECK_CPPSTD([$1])
+T8_CHECK_CPPSTDLIB([$1])
 ])
+AC_DEFUN([T8_CHECK_CPPSTD],[AX_CXX_COMPILE_STDCXX([17],[ext],[mandatory])])
```
lukasdreyer, line 55: I am not sure, if we really want ext here

jmark: I also wouldn't allow for `ext` feature set. Please set to `noext` and see if it the test passes.

.github/workflows/tests_t8code_linkage_parallel_debug.yml
```diff
@@ -192,7 +192,7 @@ jobs:
     - name: make install
       run: cd build_netcdf && make install -j
     - name: make check
-      run: cd build_netcdf && make check -j V=0 CFLAGS="$CFLAGS_var -Werror" CXXFLAGS="$CXXFLAGS_var -Werror"
+      run: cd build_netcdf && make check -j8 V=0 CFLAGS="$CFLAGS_var -Werror" CXXFLAGS="$CXXFLAGS_var -Werror"
```
jmark, line 195: Instead of hard-coding the number of jobs in each make call I suggest to set the environment variable `MAKEFLAGS` globally for all workflows.
```
  env:
    MAKEFLAGS: "-j2"
```

There are multiple hacky ways to get the number of avaible cores:
https://github.com/SimenB/github-actions-cpu-cores
https://github.com/orgs/community/discussions/25057


But it seems that only two virtual cores are available anyway. So restriciting to two jobs should be the optimum. See this for more info: https://docs.github.com/de/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources

jmark: You can set `env` variables at the top level of a runner file:
https://snyk.io/de/blog/how-to-use-github-actions-environment-variables/

jmark (changes requested):

jmark (approved):

PR #871, opened 2023-11-29
Enhance cmesh testing

Davknapp:
This is a draft PR to discuss further needed changes of this new feature. 

This branch provides a way to iterate over all of our cmeshes using googletest functionality. 
It basically consists of two classes. 
- The base class `cmesh_creator`, a class to create cmeshes of similar type (for example all cmeshes that only need a communicator). It also holds the cmesh. 
- The class `cmesh_generator`,  iterates over all cmesh_creators and all of the parameter-sets usable by each cmesh_creator. Overloads the `+`and `<` to be usable by testing::Range(start, end, step)` of the GoogleTestSuite. 

I implemented two examples of a `cmesh_creator`, one for cmesh-examples that only get a communicator and one for examples that get a communicator and a number of trees. 

Goal of this feature is to provide a way to test cmeshes, that is easily extendable by new testcases. 
In the current setup one would either:
a) Add the new cmesh function to one of the existing cmesh_creators or
b) Implement a new 'cmesh_creator' and add it to the 'cmesh_generator'

As soon as this PR is merged, we would also be able to provide a MAKRO to iterate over all cmeshes and a customized assertion to check the equality of two cmeshes. 

I will be happy about your opinion on this new feature!

#### General

#### Tests

#### Github action

#### Scripts and Wiki

#### Licence


test/t8_cmesh/t8_gtest_cmesh_copy.cxx
```diff
@@ -25,19 +25,19 @@
...
+class cmesh_copy_equality: public testing::TestWithParam<cmesh_generator> {
  protected:
   void
   SetUp () override
   {
-    cmesh_id = GetParam ();
-
-    cmesh_original = t8_test_create_cmesh (cmesh_id);
+    cmesh_gen = GetParam ();
+    cmesh_gen.create_cmesh ();
```
jmark, line 39: Why not just also return the cmesh object here and store it in `cmesh_original`? Then you could save the line after it.

test/t8_cmesh/t8_gtest_cmesh_copy.cxx
```diff
@@ -74,5 +74,10 @@ TEST_P (cmesh_copy_equality, check_equality_of_copied_cmesh_with_original)
...
 }
 
+/* TODO: I am not happy about this. Can we use something else than a global variable? */
+cmesh_generator first (0);
+cmesh_generator last ((first.get_last ()));
+cmesh_generator step (0);
+
 /* Test all cmeshes over all different inputs we get through their id */
-INSTANTIATE_TEST_SUITE_P (t8_gtest_cmesh_copy, cmesh_copy_equality, AllCmeshs);
+INSTANTIATE_TEST_SUITE_P (t8_gtest_cmesh_copy, cmesh_copy_equality, ::testing::Range (::first, ::last, ::step));
```
jmark, line 83: Instead of creating an integer range resp. iterator I suggest to directly pass a container/iterator which gives you all cmeshes one by one.

Check out the manual: http://google.github.io/googletest/reference/testing.html#INSTANTIATE_TEST_SUITE_P
There is the `ValuesIn(container)` paramter generator. The container could implement the C++ iterator interface.
Maybe just pushing to a std::vector<CmeshGenerator> does the job or implementing a container your own as you already started to do.

test/t8_cmesh/t8_gtest_cmesh_copy.cxx
```diff
@@ -74,5 +74,10 @@ TEST_P (cmesh_copy_equality, check_equality_of_copied_cmesh_with_original)
   EXPECT_TRUE (t8_cmesh_is_equal (cmesh_original, cmesh_copy));
 }
 
+/* TODO: I am not happy about this. Can we use something else than a global variable? */
```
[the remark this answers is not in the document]
jmark: However, considering the comment further doen below these three lines can be eliminated.

jmark (changes requested):
Thanks for the interesting contribution @Davknapp !

I like the idea very much and we should definitely consider to migrate to this testing scheme soon!

However, in my view creating three C++ classes for this rather simple logic is shooting birds with guns. I suggest to consider a refactor.

As also noted in the line attached comments I suggest to simply create **one** C++ iterator class which spits out one cmesh after the other. Very simple, easy to comprehend and quick to extend with lots of cmesh variants.

Here some links. Probably just implement an **output iterator**. One proably does not need to implement all operators. We do not need to apply sort or random access or such fancy stuff.

https://www.internalpointers.com/post/writing-custom-iterators-modern-cpp
https://www.programiz.com/cpp-programming/iterators

test/t8_cmesh/t8_gtest_cmesh_copy.cxx
Davknapp: Because the Range would create all cmeshes at once, leading to a huge memory consumption. That way we only create the cmesh when we need it.

Davknapp:
> Thanks for the interesting contribution @Davknapp !
> 
> I like the idea very much and we should definitely consider to migrate to this testing scheme soon!
> 
> However, in my view creating three C++ classes for this rather simple logic is shooting birds with guns. I suggest to consider a refactor.
> 
> As also noted in the line attached comments I suggest to simply create **one** C++ iterator class which spits out one cmesh after the other. Very simple, easy to comprehend and quick to extend with lots of cmesh variants.
> 
> Here some links. Probably just implement an **output iterator**. One proably does not need to implement all operators. We do not need to apply sort or random access or such fancy stuff.
> 
> https://www.internalpointers.com/post/writing-custom-iterators-modern-cpp https://www.programiz.com/cpp-programming/iterators

The goal is to be able to run tests over a sum of cmeshes. I agree, for a single set of cmeshes (for example all cmeshes that only need a communicator, or all cmeshes that get a communicator and a number of trees to generate) it will be better to implement an iterator. 
We want to iterate over varying number of parameters (not all functions creating a cmesh have the same number of arguments), this we way, we are able to specify how we iterate over a subset. Furthermore not all functions create valid cmeshes for all parameters, this way we are able to prevent such cmeshes. Additionally we can use the step to skip entire subsets of cmeshes. Maybe it is a good idea to have a call about this subject ;)

test/t8_cmesh/t8_gtest_cmesh_copy.cxx
jmark, line 1: Strange. Besides deleting some lines nothing changes in this file. Shouldn't be here an example using the new feature?

Would you please provide a gtest example which uses the new cmesh generator?

test/t8_cmesh_generator/t8_gtest_cmesh_cartestian_product.hxx
```diff
@@ -0,0 +1,223 @@
...
+/**
+ * A base class for cmesh examples.
+ * An example must be comparable to another via == and !=. 
+ * 
+ * For pretty debug output a function to translate the parameters of the example to a string should be provided. 
+ * 
+ * The
+ * 
+ */
+class base_example {
```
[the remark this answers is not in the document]
jmark: Do we actually need this base class here? I do not see that much functionality.

test/t8_cmesh_generator/t8_gtest_cmesh_cartestian_product.hxx
```diff
@@ -0,0 +1,223 @@
...
+   */
+  base_example (std::string name): name (name) {};
+
+  /**
+   * A function to create a cmesh. The class should not own the cmesh.
+   * 
+   * \return t8_cmesh_t 
+   */
+  virtual t8_cmesh_t
+  cmesh_create () const
```
jmark, line 57: If this base class is supposed to be also used for other use cases than this member function looks quite specific for cmesh generators, though.

test/t8_cmesh_generator/t8_cmesh_example_sets.hxx
```diff
@@ -0,0 +1,60 @@
...
+ */
+auto pretty_print_base_example = [] (const testing::TestParamInfo<base_example *> &info) {
+  std::string name;
+  info.param->param_to_string (name);
+  return name;
+};
+
+namespace cmesh_list
+{
+std::vector<example_parameter_combinator *> cart_prod_vec
```
jmark, line 48: Why do I need to create a Cartesian product of these three cmesh examples? What Cartesian product are we producing here? Refinement levels vs element classes, for example?

jmark (changes requested):
Let's have a direct telco about this PR. It is pretty tedious to get an overview over the interdependence of classes you use here and how and where they are used.

test/t8_cmesh/t8_gtest_cmesh_copy.cxx
Davknapp: Following our talk about the new feature in mattermost I created a seperate testfile that tests the new feature and also demonstrates it usage. You can find it t8_cmesh_generator_test. cxx

test/t8_cmesh_generator/t8_cmesh_example_sets.hxx
Davknapp: I guuess the name is misleading here. Should be parameter_combinations instead of combinator. This vector holds three sets of cmeshes together with the Cartesian product of their parameters. If you have a look at the files where the sets are constructed you will see the Cartesian product

test/t8_cmesh_generator/t8_gtest_cmesh_cartestian_product.hxx
Davknapp: This base class is necessary so the sum of sets can be produced. Since the examples have different types I can not put them directly in one vector. But they all have the same base class. Therefore I can use a pointer of this base class to address all of them via a vector

test/t8_cmesh_generator/t8_gtest_cmesh_cartestian_product.hxx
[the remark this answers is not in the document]
Davknapp: This base class is necessary so the sum of sets can be produced. Since the examples have different types I can not put them directly in one vector. But they all have the same base class. Therefore I can use a pointer of this base class to address all of them via a vector

test/t8_cmesh_generator/t8_gtest_cmesh_cartestian_product.hxx
Davknapp: Cmesh_example_base might be a more fitting name then

jmark (approved):

PR #872, opened 2023-11-29
Moved t8_forest_element_from_ref_coords implementation into cxx file.

jmark:

Title says it all. Only some compiler versions make `inline [...]` routines in header files also visible as symbols in shared libraries which application then can use in their code. Otherwise the routine is statically inlined in the t8code library and cannot be called by external applications.

#### General

#### Tests

#### Github action

#### Scripts and Wiki

#### Licence

### DIFF
--- a/src/t8_forest/t8_forest_cxx.cxx
+++ b/src/t8_forest/t8_forest_cxx.cxx
@@ -423,6 +423,13 @@ t8_forest_element_from_ref_coords_ext (t8_forest_t forest, t8_locidx_t ltreeid, 
   t8_geometry_evaluate (cmesh, gtreeid, tree_ref_coords, num_coords, coords_out);
 }
 
+void
+t8_forest_element_from_ref_coords (t8_forest_t forest, t8_locidx_t ltreeid, const t8_element_t *element,
+                                   const double *ref_coords, const size_t num_coords, double *coords_out)
+{
+  t8_forest_element_from_ref_coords_ext (forest, ltreeid, element, ref_coords, num_coords, coords_out, NULL);
+}
+
 /* Compute the diameter of an element. */
 double
 t8_forest_element_diam (t8_forest_t forest, t8_locidx_t ltreeid, const t8_element_t *element)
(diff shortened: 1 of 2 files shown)


sandro-elsweijer:
With which compiler do you have the problem? Does it maybe work with the cpp17 branch?
`inline` is by far not a new feature and I am worried, that we will never program in a modern way, if we block "new" concepts because of ancient compilers

jmark:
> @jmark With which compiler do you have the problem? Does it maybe work with the cpp17 branch? `inline` is by far not a new feature and I am worried, that we will never program in a modern way, if we block "new" concepts because of ancient compilers

@sandro-elsweijer I totally get your point and sympathize with your view.

On my workstation I compiled t8code with `g++ (Ubuntu 9.4.0-1ubuntu1~20.04.2) 9.4.0` and even tried `-std=c++17`. It does not create a symbol for `t8_forest_element_from_ref_coords` in the shared library `libt8.so`. This compiler version series is not that old. It started from 2019 and last revision was released in 2022: https://gcc.gnu.org/gcc-9/

`g++ (Spack GCC) 12.1.0` on the other hand creates a symbol entry in `libt8.so`.

This raises a general question on how we want to handle compatibility to many different kinds of compilers and platforms.

sandro-elsweijer:
> > @jmark With which compiler do you have the problem? Does it maybe work with the cpp17 branch? `inline` is by far not a new feature and I am worried, that we will never program in a modern way, if we block "new" concepts because of ancient compilers
> 
> @sandro-elsweijer I totally get your point and sympathize with your view.
> 
> On my workstation I compiled t8code with `g++ (Ubuntu 9.4.0-1ubuntu1~20.04.2) 9.4.0` and even tried `-std=c++17`. It does not create a symbol for `t8_forest_element_from_ref_coords` in the shared library `libt8.so`. This compiler version series is not that old. It started from 2019 and last revision was released in 2022: https://gcc.gnu.org/gcc-9/
> 
> `g++ (Spack GCC) 12.1.0` on the other hand creates a symbol entry in `libt8.so`.
> 
> This raises a general question on how we want to handle compatibility to many different kinds of compilers and platforms.

Does this her maybe help? https://gcc.gnu.org/onlinedocs/gcc/Alternate-Keywords.html

PR #892, opened 2024-01-11
Feature: C-API for Geometry Analytic

jmark:

Also added an 11th example to geometry examples to test the C API.

grafik

#### General

#### Tests

#### Github action

#### Scripts and Wiki

#### Licence


example/geometry/t8_example_geometries.cxx
```diff
@@ -1063,7 +1091,8 @@ main (int argc, char **argv)
                       "\t\t7 - A curved triangle with an occ curve.\n"
                       "\t\t8 - A cube with two occ curves as edges.\n"
                       "\t\t9 - Two cubes with one occ surface as face.\n"
-                      "\t\t10 - A hollow cylinder with a occ surface on the in- and outside.\n");
+                      "\t\t10 - A hollow cylinder with a occ surface on the in- and outside.\n"
+                      "\t\t11 - A quad morphed into a sphere.\n");
```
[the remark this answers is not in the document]
jmark: Moved the example to slot 7 as you suggested.

example/geometry/t8_example_geometries.cxx
```diff
@@ -589,6 +607,15 @@ t8_analytic_geom (int level, t8_example_geom_type geom_type)
     t8_cmesh_set_tree_class (cmesh, 0, T8_ECLASS_QUAD);
     snprintf (vtuname, BUFSIZ, "forest_moving_lvl_%i", level);
     break;
+  case T8_GEOM_ANALYTIC_QUAD_TO_SPHERE:
```
jfussbro, line 610: Please also adjust the position of the new geometry in the switch statement for to keep everything in sync

jfussbro (changes requested):
One last small change :)

example/geometry/t8_example_geometries.cxx
jmark: It already is.

sandro-elsweijer (changes requested):
Why move all the functions from the hxx header to the new h header? All the other geometries just provide a `new` and `destroy` function in their c interface. I think we should keep everything in the hxx header, since we want to move to cpp anyway.

jmark:
> Why move all the functions from the hxx header to the new h header? All the other geometries just provide a `new` and `destroy` function in their c interface. I think we should keep everything in the hxx header, since we want to move to cpp anyway.

@sandro-elsweijer `.hpp` header files cannot be included in `.c` files. In order to use the geometry analytical interface from C the signatures for the callbacks must be present in the `.h` file.

sandro-elsweijer:
> > Why move all the functions from the hxx header to the new h header? All the other geometries just provide a `new` and `destroy` function in their c interface. I think we should keep everything in the hxx header, since we want to move to cpp anyway.
> 
> @sandro-elsweijer `.hpp` header files cannot be included in `.c` files. In order to use the geometry analytical interface from C the signatures for the callbacks must be present in the `.h` file.

@jmark I understand your problem here. But it bugs me, that the cpp interface would be removed even though we want to have an explicit cpp interface. Maybe we can discuss defining these functions twice in the next developer meeting

sandro-elsweijer:
> > @jmark I understand your problem here. But it bugs me, that the cpp interface would be removed even though we want to have an explicit cpp interface. Maybe we can discuss defining these functions twice in the next developer meeting
> 
> If you want a real C++ API for this then I suggest to offer a geometry analytic base class from which you derive and implement the analytical evaluate and Jacobian member functions. The actual evaluate and Jacobian function (wrapping around the analytical callbacks) should be finalized in the analytical geometry base class, of course.
> 
> BTW: The C++ interface is not removed. The `.h` header file is included in the `.hpp` header. Technically, nothing changed from the C++ perspective.

Then let's keep this in mind for our cpp interface t8code week. Do you mind writing a issue for that so that we do not forget it?

jmark:
> > > @jmark I understand your problem here. But it bugs me, that the cpp interface would be removed even though we want to have an explicit cpp interface. Maybe we can discuss defining these functions twice in the next developer meeting
> > 
> > 
> > If you want a real C++ API for this then I suggest to offer a geometry analytic base class from which you derive and implement the analytical evaluate and Jacobian member functions. The actual evaluate and Jacobian function (wrapping around the analytical callbacks) should be finalized in the analytical geometry base class, of course.
> > BTW: The C++ interface is not removed. The `.h` header file is included in the `.hpp` header. Technically, nothing changed from the C++ perspective.
> 
> @jmark Then let's keep this in mind for our cpp interface t8code week. Do you mind writing an issue for that so that we do not forget it?

Done. https://github.com/DLR-AMR/t8code/issues/904

@jfussbro @sandro-elsweijer Anything else to do?

jfussbro (approved):
Approved! :)

jmark:
@sandro-elsweijer ?

jmark:
@sandro-elsweijer ??

example/geometry/t8_example_geometries.cxx
```diff
@@ -26,7 +26,7 @@
 #include <t8_forest/t8_forest_general.h>
 #include <t8_forest/t8_forest_io.h>
 #include <t8_geometry/t8_geometry_base.hxx>
-#include <t8_geometry/t8_geometry_implementations/t8_geometry_analytic.hxx>
```
sandro-elsweijer, line 29: In anticipation for the coming changes regarding the differentiation between the c and cpp interface, it would be nice to use the cpp interface here

src/t8_geometry/t8_geometry_implementations/t8_geometry_analytic.cxx
```diff
@@ -66,10 +66,32 @@ t8_geometry_analytic::t8_geom_load_tree_data (t8_cmesh_t cmesh, t8_gloidx_t gtre
...
+  return (t8_geometry_c *) geom;
+}
+
 void
 t8_geom_load_tree_data_vertices (t8_cmesh_t cmesh, t8_gloidx_t gtreeid, const void **vertices_out)
 {
   T8_ASSERT (t8_cmesh_is_committed (cmesh));
   t8_locidx_t ltreeid = t8_cmesh_get_local_id (cmesh, gtreeid);
   *vertices_out = t8_cmesh_get_tree_vertices (cmesh, ltreeid);
 }
```
sandro-elsweijer, line 95: It seems like something got mixed up here. The git blame tells me, that this error is rather old, but we should fix it. The analytic geometry does not inherit from the geometry with vertices, and therefore has no vertices. But it has user data, which is retrieved here. Rephrasing the specific word vertices to user data would IMHO solve the confusion.

src/t8_geometry/t8_geometry_implementations/t8_geometry_analytic.h
```diff
@@ -0,0 +1,96 @@
...
+                          const void *user_data);
+
+/**
+ * Load vertex data from given tree. 
+ * \param [in]  cmesh       The cmesh.
+ * \param [in]  gtreeid     The global tree id (in the cmesh).
+ * \param [out] vertex_out  The load tree vertices.
+ */
+void
+t8_geom_load_tree_data_vertices (t8_cmesh_t cmesh, t8_gloidx_t gtreeid, const void **vertices_out);
```
sandro-elsweijer, line 92: Same comment about the "vertices"

sandro-elsweijer (changes requested):
Just minor comments

src/t8_geometry/t8_geometry_implementations/t8_geometry_analytic.cxx
jmark: Done!

src/t8_geometry/t8_geometry_implementations/t8_geometry_analytic.h
jmark: Done.

example/geometry/t8_example_geometries.cxx
jmark: I switched to the C++ interface.

jmark:
> Just minor comments

Thank you very much! Refactored as you suggested.

PR #751, opened 2023-09-14
Feature point inside check for axis aligned geom

Davknapp:
Implemented a point inside check that is specialized for axis-aligned geometries. 

Also enhanced the gtest to test this feature. 

**_All these boxes must be checked by the reviewers before merging the pull request:_**

As a reviewer please read through all the code lines and make sure that the code is fully understood, bug free, well-documented and well-structured.

#### General
- [x] The reviewer executed the new code features at least once and checked the results manually

- [x] The code follows the [t8code coding guidelines](https://github.com/holke/t8code/wiki/Coding-Guideline)
- [x] New source/header files are properly added to the Makefiles
- [x] The code is well documented
- [x] All function declarations, structs/classes and their members have a proper doxygen documentation
- [x] All new algorithms and data structures are sufficiently optimal in terms of memory and runtime (If this should be merged, but there is still potential for optimization, create a new issue)

#### Tests
- [x] The code is covered in an existing or new test case using Google Test

#### Github action

- [x] The code compiles without warning in debugging and release mode, with and without MPI (this should be executed automatically in a github action)
- [x] All tests pass (in various configurations, this should be executed automatically in a github action)

  If the Pull request introduces code that is not covered by the github action (for example coupling with a new library):
  - [ ] Should this use case be added to the github action?
  - [ ] If not, does the specific use case compile and all tests pass (check manually)

#### Scripts and Wiki

- [x] If a new directory with source-files is added, it must be covered by the `script/find_all_source_files.scp` to check the indentation of these files.
- [x] If this PR introduces a new feature, it must be covered in an example/tutorial and a Wiki article.

#### Licence

- [x] The author added a BSD statement to `doc/` (or already has one)

PR #897, opened 2024-01-19
Moving the negative volume check into the geometry module

Davknapp:
As @jmark mentioned in #751 we should also move the no-negative-volume check into the geometry module. 

#### General

#### Tests

#### Github action

#### Scripts and Wiki

#### Licence


sandro-elsweijer (changes requested):
We need a test for that

sandro-elsweijer:
Did you forget to push the t8_geometry_zero.h?

src/t8_geometry/t8_geometry_base.hxx
```diff
@@ -120,29 +120,20 @@ struct t8_geometry
...
+  t8_geom_tree_negative_volume (const t8_cmesh_t cmesh) const
   {
-    int is_inside = 0;
-    t8_geom_point_batch_inside_element (forest, ltreeid, element, points, 1, &is_inside, tolerance);
-    return is_inside;
-  }
+    SC_ABORTF ("Function not implemented yet");
+    /* To compress compiler warnings. */
+    return 0;
+  };
```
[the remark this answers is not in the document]
Davknapp: I think we already had this discussion once: 
https://github.com/DLR-AMR/t8code/pull/751#discussion_r1487672396

Davknapp:
> Did you forget to push the t8_geometry_zero.h?

I don't think so? Is there anything missing?

src/t8_geometry/t8_geometry_implementations/t8_geometry_linear_axis_aligned.cxx
```diff
@@ -82,6 +82,13 @@ t8_geometry_linear_axis_aligned::t8_geom_point_batch_inside_element (t8_forest_t
   return;
 }
 
+bool
+t8_geometry_linear_axis_aligned::t8_geom_tree_negative_volume (const t8_cmesh_t cmesh) const
+{
+  return active_tree_vertices[3] < active_tree_vertices[0] || active_tree_vertices[4] < active_tree_vertices[1]
+         || active_tree_vertices[5] < active_tree_vertices[2];
```
lukasdreyer, line 89: I am still not sure that this is right. If exactly two axis are switched, I think the volume should be positive. Also, we should probably have a test for that

lukasdreyer (changes requested):

src/t8_geometry/t8_geometry_implementations/t8_geometry_linear_axis_aligned.cxx
Davknapp: In general you are right, but we require this condition for our axis-aligned geometry anyway, as it simplifies other computations, too. 
You're right about the test, but currently testing would be difficult, as we would need to commit a cmesh with a negative volume. @sandro-elsweijer currently works on a branch that circumvents it and as soon as it is merged I will implement a geometry-wise test for this function

Davknapp: With this condition I mean, that the x-coordinate of the first point is smaller than the x-coordinate of the second (same for y and z)

lukasdreyer: If you already require the harder condition that all coordinates are sorted, this function could just return false.

lukasdreyer: Which other computations are simplified by disallowing switching two of the 3 axes? My intuition is that the formulas should also work for negative lengths.

Davknapp: For example the point-inside check

lukasdreyer: Ah, that makes sense

lukasdreyer: I like that!

src/t8_geometry/t8_geometry_implementations/t8_geometry_linear_axis_aligned.cxx
```diff
@@ -82,6 +104,13 @@ t8_geometry_linear_axis_aligned::t8_geom_point_batch_inside_element (t8_forest_t
   return;
 }
 
+bool
+t8_geometry_linear_axis_aligned::t8_geom_tree_negative_volume (const t8_cmesh_t cmesh) const
+{
+  T8_ASSERT (correct_point_order (active_tree_vertices));
```
sandro-elsweijer, line 110: Shouldn't we just return true, if the vertices are not in the correct order? Because this would lead to a negative volume, right?

sandro-elsweijer (approved):
Just one small thing

src/t8_geometry/t8_geometry_implementations/t8_geometry_linear_axis_aligned.cxx
Davknapp: No, because it is a requirement for this geometry that the vertices are ordered correctly. Especially for the is_inside check. Therefore, we use an assertion instead of an if/else.

PR #929, opened 2024-02-05
Initial CMake support

dutkalex:
This PR is a MVP candidate for CMake build support in t8code. Currently supports building the main library, the tests, the examples and the tutorials.

#### General

#### Tests

#### Github action

#### Scripts and Wiki

#### Licence


CMakeLists.txt
```diff
@@ -0,0 +1,36 @@
+cmake_minimum_required( VERSION 3.16 )
+project( T8CODE DESCRIPTION "Parallel algorithms and data structures for tree-based AMR with arbitrary element shapes." LANGUAGES C CXX )
+include( CTest )
+
+option( T8CODE_ENABLE_MPI "" ON )
+option( T8CODE_BUILD_TESTS "" ON )
+option( T8CODE_BUILD_TUTORIALS "" ON )
```
jmark, line 7: How about also adding the `example` directory and build all examples?

jmark (changes requested):
Really nice work! Thank you very much! Looks very promising so far!

Would you please substitute all `leafs` with `leaves` and also merge in the latest `main`? We just made this global change in t8code. Some files got renamed by this.

CMakeLists.txt
dutkalex: Yes and I realized this morning that I have not ported all tutorials too. It's on my plate for today 😉

dutkalex:
When building the 116 unit tests, I noticed that not all of them pass. From my limited understanding of the code base, I think that most of them are inherited from p4est and sc, but 2 that I ported myself fail. However, the pipeline passes as it uses the autotools BS...

CMakeLists.txt
dutkalex: The examples are now ported

CMakeLists.txt
```diff
@@ -0,0 +1,43 @@
...
+option( T8CODE_BUILD_EXAMPLES "" ON )
+
+if( NOT CMAKE_CXX_STANDARD )
+    set( CMAKE_CXX_STANDARD 17 )
+endif()
+
+if( NOT CMAKE_INSTALL_PREFIX )
+    set( CMAKE_INSTALL_PREFIX ${CMAKE_CURRENT_LIST_DIR}/build )
```
jmark, line 15: In the `ccmake` configure GUI I get `[build_directory]/p4est/local`.

CMakeLists.txt
```diff
@@ -0,0 +1,43 @@
...
+
+if ( T8CODE_BUILD_TUTORIALS )
+    add_subdirectory( ${CMAKE_CURRENT_LIST_DIR}/tutorials )
+endif()
+
+if ( T8CODE_BUILD_EXAMPLES )
+    add_subdirectory( ${CMAKE_CURRENT_LIST_DIR}/example )
+endif()
+
+install( TARGETS T8 DESTINATION ${CMAKE_INSTALL_PREFIX}/lib )
```
jmark, line 43: The result library is named `libT8.so`. We usually store it as `libt8.so`. Is this doable?

dutkalex:
The zlib part is managed by the existing p4est & sc CMake build system which does strange things at times on my machine too. Maybe I will try to fix this in the future but it is outside of the scope of this PR IMO. I will investigate the 2 t8_gtest_*** which fail and this mesh file thing. Thx for the feedback! 🙃

example/CMakeLists.txt
```diff
@@ -0,0 +1,53 @@
+# Conditionally adding new sources to the main T8 target as in the original BS looked wrong...
+add_library( T8ex )
```
jmark, line 2: Is it possible to link the example binaries with the shared library in the INSTALL_PREFIX? Alternatively, compile in t8code statically?

jmark (changes requested):
Thanks for you effort! Very educational reading you cmake scripts!

It would be nice if we could replicate the folder structure in the INSTALL_PREFIX as done with the old autotools build system. For example, we have bindings generators for e.g. Julia which assume this structure.

Let me know if you need more info!

example/CMakeLists.txt
dutkalex: Hmmm yes, I think this should be easy to do. Do you want both at the same time or is some kind of `T8CODE_ENABLE_DYNAMIC_LINKAGE` flag acceptable ?

jmark: `8CODE_ENABLE_DYNAMIC_LINKAGE` sounds like a viable option.

CMakeLists.txt
dutkalex: Done

example/CMakeLists.txt
dutkalex: Done

dutkalex:
> It would be nice if we could replicate the folder structure in the INSTALL_PREFIX as done with the old autotools build system. For example, we have bindings generators for e.g. Julia which assume this structure.

What install directory structure is needed exactly? As of this comment it just contains `lib/libt8.*` and `include/some_zlib_stuff`

CMakeLists.txt
dutkalex: Very strange, maybe p4est does some stuff with this variable when not specified by the user. However, it does the right thing when typing `cmake .. `DCMAKE_INSTALL_PREFIX=...`

dutkalex:
Also, what is your policy regarding documentation? Where do you want me to document how to use the CMake BS? In a new section in the `INSTALL` file maybe?

dutkalex:
As I can't make a clean PR for the wiki, here is the documentation for the CMake installation.
[Installation-CMake.md](https://github.com/DLR-AMR/t8code/files/14194585/Installation-CMake.md)

I would also replace the first paragraph in the Installation.md file with something like:
```
Here, we discuss how to install t8code from the github repository on a linux machine using 
the autotools build system. Alternatively, t8code can also be installed using CMake (then 
see [this page](https://github.com/holke/t8code/wiki/Installation-CMake) instead).
```

@holke @jmark Anything more you need to merge this?

example/CMakeLists.txt
```diff
@@ -0,0 +1,49 @@
...
```
jmark, line 2:
```suggestion
add_library( t8example )
```

Can you rename this to `libt8example.so`? Than it is much clearer what this library is for.

example/CMakeLists.txt
```diff
@@ -0,0 +1,49 @@
+# Conditionally adding new sources to the main T8 target as in the original BS looked wrong...
```
jmark, line 1: I agree with that one.

src/CMakeLists.txt
```diff
@@ -0,0 +1,172 @@
...
+
+target_include_directories( T8 PUBLIC ${CMAKE_CURRENT_LIST_DIR} )
+target_link_libraries( T8 PUBLIC P4EST::P4EST SC::SC )
+
+if ( T8CODE_ENABLE_MPI )
+    target_compile_definitions( T8 PUBLIC T8_ENABLE_MPI )
+    target_compile_definitions( T8 PUBLIC T8_ENABLE_MPIIO )
+    target_link_libraries( T8 PUBLIC MPI::MPI_C )
+endif()
+
```
jmark, line 33: Is it very tedious to also add options to link t8code against VTK?

A typical `configure` call to link against VTK looks like this:
```
  ../t8code/configure \
    --prefix="${T8CODE_INSTALL_PREFIX}" \
    --enable-mpi \
    --with-vtk \
    --with-vtk_version_number=9.1 \
    CFLAGS="-O3" CXXFLAGS="-O3" \
    CC=mpicc CXX=mpicxx \
    LDFLAGS="-L${VTK_INSTALL_PREFIX}/lib" \
    CPPFLAGS="-I${VTK_INSTALL_PREFIX}/include/vtk-9.1"
```

example/CMakeLists.txt
jmark, line 1: Can you also add the examples to the `CMAKE_INSTALL_PREFIX`?

tutorials/CMakeLists.txt
jmark, line 1: Can you also add the tutorial binaries to the `CMAKE_INSTALL_PREFIX`?

CMakeLists.txt
jmark, line 1: When I call `ccmake` the `BUILD_SHARED_LIBS` options defaults to `OFF`. I have to enable this setting under any configuration to make it compile. For example, I tried disabling `T8CODE_ENABLE_DYNAMIC_LINKAGE`  under the assumption that shared libs are then not needed. Still the building fails.

Do you observe the same behavior on your machine?

jmark (changes requested):

CMakeLists.txt
dutkalex: No, I can't reproduce the build failure. On my machine all 4 combinations of `BUILD_SHARED_LIBS ` and `T8CODE_ENABLE_DYNAMIC_LINKAGE ` work :/

src/CMakeLists.txt
dutkalex: I do not have VTK installed on my machine, but according to VTK's doc it should work now. Can you test on your side?

jmark (changes requested):
Thank you very much for you efforts! I tried out the build system and it works quite well! However, I still found some issues. We should fix them before merging into the official repo.

* First, I still have to manually enable `BUILD_SHARED_LIBS`. Could you make this the default?
grafik

* Then, I got the following warning. Is it possible to properly set the version numbers during configuration?
grafik

* I prefer to create a build directory not as a sub-directory of the source codes but somewhere else and call `ccmake /path/to/t8code/sources`. When I do this and enable VTK I get the following error when configuring:
grafik
When doing this in a sub-folder of t8code, e.g. `build`, it works just fine. Please support other build folder pathes.

* The default `CMAKE_INSTALL_PREFIX` is still set by `p4est`.
grafik

src/CMakeLists.txt
```diff
@@ -0,0 +1,184 @@
...
+set_target_properties( T8 PROPERTIES OUTPUT_NAME t8 )
+
+target_compile_definitions( T8 PUBLIC T8_CMAKE_BUILD )
+target_compile_definitions( T8 PUBLIC T8_CC="${CMAKE_C_COMPILER}" )
+target_compile_definitions( T8 PUBLIC T8_CFLAGS="${CMAKE_C_FLAGS}" )
+target_compile_definitions( T8 PUBLIC T8_CPP="${CMAKE_CXX_COMPILER}" )
+target_compile_definitions( T8 PUBLIC T8_CPPFLAGS="${CMAKE_CXX_FLAGS}" )
+target_compile_definitions( T8 PUBLIC T8_LDFLAGS="${CMAKE_SHARED_LINKER_FLAGS}" )
+target_compile_definitions( T8 PUBLIC T8_LIBS="???" ) # Not sure what this flag is supposed to be...
```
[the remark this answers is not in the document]
dutkalex: This flag is only used en `t8.c` line 158. Removing it breaks the build:
`[ 42%] Building C object src/CMakeFiles/T8.dir/t8.c.o
/Users/adutka/t8code/src/t8.c:158:50: error: use of undeclared identifier 'T8_LIBS'
  t8_global_productionf ("%-*s %s\n", w, "LIBS", T8_LIBS);`
 I don't quite undersand what was the purpose of this line, but I find it quite useless (at least in the context of a CMake build)

dutkalex:
> Thank you very much for you efforts! I tried out the build system and it works quite well! However, I still found some issues. We should fix them before merging into the official repo.
> 
> * First, I still have to manually enable `BUILD_SHARED_LIBS`. Could you make this the default?
>   grafik

I can't change that as it is defined in the `sc` build system, but I think we should consider opening issues/PRs to fix this in the `sc` repo

> * Then, I got the following warning. Is it possible to properly set the version numbers during configuration?
>   grafik

I can't reproduce it on my machine but I think this one this one might be a consequence of your 3rd remark, because CMake queries git to define the `T8_VERSION_XXX` constants. It should work now. Can you confirm?

> * I prefer to create a build directory not as a sub-directory of the source codes but somewhere else and call `ccmake /path/to/t8code/sources`. When I do this and enable VTK I get the following error when configuring:
>   grafik
>   When doing this in a sub-folder of t8code, e.g. `build`, it works just fine. Please support other build folder pathes.

This has nothing to do with VTK as I was able to reproduce it on my machine (without VTK). This should be fixed now.

> * The default `CMAKE_INSTALL_PREFIX` is still set by `p4est`.
>   grafik

Same answer as for your first remark: this is outside of the scope of this PR as the problem comes from the `p4est` repo

src/CMakeLists.txt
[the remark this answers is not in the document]
jmark: It is defined here: https://github.com/DLR-AMR/t8code/blob/bd82ee6ee4e87a610abd8f76a45a3c0cfbbeb422/Makefile.t8.pre.in#L27

Can we reproduce this with cmake?

jmark:
> > * First, I still have to manually enable `BUILD_SHARED_LIBS`. Could you make this the default?
> >   grafik
> 
> I can't change that as it is defined in the `sc` build system, but I think we should consider opening issues/PRs to fix this in the `sc` repo

How about we set the option to true before including `sc` sub-project? See here for more info:
https://stackoverflow.com/questions/14061605/override-option-in-cmake-subproject


> 
> > * Then, I got the following warning. Is it possible to properly set the version numbers during configuration?
> >   grafik
> 
> I can't reproduce it on my machine but I think this one this one might be a consequence of your 3rd remark, because CMake queries git to define the `T8_VERSION_XXX` constants. It should work now. Can you confirm?

Can confirm! Thanks!


> > * I prefer to create a build directory not as a sub-directory of the source codes but somewhere else and call `ccmake /path/to/t8code/sources`. When I do this and enable VTK I get the following error when configuring:
> >   grafik
> >   When doing this in a sub-folder of t8code, e.g. `build`, it works just fine. Please support other build folder pathes.
> 
> This has nothing to do with VTK as I was able to reproduce it on my machine (without VTK). This should be fixed now.

Confirmed! Thanks!

> 
> > * The default `CMAKE_INSTALL_PREFIX` is still set by `p4est`.
> >   grafik
> 
> Same answer as for your first remark: this is outside of the scope of this PR as the problem comes from the `p4est` repo

How about we copy the following into t8codes cmake build scripts as well?
https://github.com/cburstedde/p4est/blob/aee0bf1ccccfbcaeba53beb38688b1d4fc3da8ab/cmake/options.cmake#L25

dutkalex:
> > > * First, I still have to manually enable `BUILD_SHARED_LIBS`. Could you make this the default?
> > >   grafik
> > 
> > 
> > I can't change that as it is defined in the `sc` build system, but I think we should consider opening issues/PRs to fix this in the `sc` repo
> 
> How about we set the option to true before including `sc` sub-project? See here for more info: https://stackoverflow.com/questions/14061605/override-option-in-cmake-subproject

Ok, I was not aware of this trick. I will add this.

> > > * The default `CMAKE_INSTALL_PREFIX` is still set by `p4est`.
> > >   grafik
> > 
> > 
> > Same answer as for your first remark: this is outside of the scope of this PR as the problem comes from the `p4est` repo
> 
> How about we copy the following into t8codes cmake build scripts as well? https://github.com/cburstedde/p4est/blob/aee0bf1ccccfbcaeba53beb38688b1d4fc3da8ab/cmake/options.cmake#L25

Why not I guess. I often say that `set( ... FORCE )` is a code smell and I still think that p4est should not be doing that, but in this case it would be better than nothing... But what should be the default then?

src/CMakeLists.txt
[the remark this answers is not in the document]
dutkalex: Looks like this would not be trivial and a little bit hacky (see https://stackoverflow.com/questions/34165365/retrieve-all-link-flags-in-cmake for example)

jmark:
> I would argue that `set( ... FORCE )` is a code smell and I still think that p4est should not be doing that, but in this case it would be better than nothing I guess... But what should be the default then? I want to emphasize that I think that this is something that **must** be fixed (among other things) in the p4est repo, because this quick fix is just a mean of kicking the can further down the road, and make it someone else's problem...

Fair enough. You have a point there.

src/CMakeLists.txt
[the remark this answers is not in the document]
dutkalex: Closing this thread. Feel free to reopen if you disagree with me on this matter.

dutkalex:
@jmark Anything else that blocks integration? As I noticed you guys are very active on the main branch, it's only a matter of time before a merge conflict arises...

CMakeLists.txt
```diff
@@ -0,0 +1,60 @@
...
+list( APPEND CMAKE_CXX_FLAGS "-Wall" )
+
+
+if( T8CODE_ENABLE_MPI )
+    find_package( MPI COMPONENTS C REQUIRED )
+    if( NOT MPIEXEC_EXECUTABLE )
+        message( FATAL_ERROR "MPIEXEC was not found" )
+    endif()
+    set( SC_ENABLE_MPI ON )
+    set( mpi ON ) # This is very dirty, we should consider fixing this in the p4est repo
```
jmark, line 36:
```suggestion
    set( mpi ON ) # This is very dirty, we should consider fixing this in the p4est repo
    add_compile_definitions(T8_ENABLE_MPI=1)
```

CMakeLists.txt
```diff
@@ -0,0 +1,60 @@
...
+endif()
+
+if( T8CODE_ENABLE_VTK )
+    find_package( VTK REQUIRED )
```
jmark, line 40:
```suggestion
    find_package( VTK REQUIRED )

    # Following definition bypasses the 'VTK used version' check in
    # `t8_gtest_vtk_linkage.cxx` since we do not explicitly set the VTK version via cmake.
    add_compile_definitions(T8_VTK_VERSION_USED="${VTK_MAJOR_VERSION}.${VTK_MINOR_VERSION}")

    add_compile_definitions(T8_WITH_VTK=1)
```

src/CMakeLists.txt
```diff
@@ -0,0 +1,197 @@
...
+
+if ( T8CODE_ENABLE_MPI )
+    target_compile_definitions( T8 PUBLIC T8_ENABLE_MPI )
+    target_compile_definitions( T8 PUBLIC T8_ENABLE_MPIIO )
+    target_link_libraries( T8 PUBLIC MPI::MPI_C )
+endif()
+
+if( T8CODE_ENABLE_VTK )
+    target_include_directories( T8 PUBLIC ${VTK_INCLUDE_DIRS} )
+    target_link_libraries( T8 PUBLIC ${VTK_LIBRARY_DIRS} )
```
jmark, line 56:
```suggestion
    target_link_libraries( T8 PUBLIC ${VTK_LIBRARIES} )
```

jmark (changes requested):
I took a closer look if t8code is actually linked against VTK when enabled. It wasn't. So I added my modifications as review comments to make it work.

Also made other suggestions here and there. Please take a look at them, too.

jmark:
> @jmark Anything that blocks integration? As I noticed you guys are very active on the main branch, it's only a matter of time before a merge conflict arises...

There are still some issues to solve and some things to do. What is missing is a proper test of the cmake build system via Github CI where also all tests are passing.

As of now, the cmake build system is considered highly experimental and far from feature complete. I have to consult with the other core developers if we want to merge it with `main` nevertheless but with the clear note that this feature is experimental.

Migrating to a new build system takes time. Nevertheless, awesome work!

CMakeLists.txt
```diff
@@ -0,0 +1,60 @@
...
+endif()
+
+# Override default for this libsc option
+set( BUILD_SHARED_LIBS ON CACHE BOOL "Build libsc as a shared library" )
+
```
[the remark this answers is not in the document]
dutkalex: This is a much nicer solution!

dutkalex:
> There are still some issues to solve and some things to do. For example, what is missing is a proper test of the cmake build system via Github CI where also all tests are passing.

This looks to me like a "chicken and the egg" situation: if we want to integrate into `main`, we need to have test coverage first, but in order for `main` to continue passing the tests, we need to integrate the code first. I think that since this is purely new stuff that can not break anything else that is already in production, having temporary uncovered (but obviously flagged as experimental) code in `main` is a better alternative that to break the pipeline for everyone else...

> As of now, the cmake build system is considered highly experimental and far from feature complete. I have to consult with the other core developers if we want to merge it with `main` nevertheless but with the clear note that this feature is experimental.
> 
> Migrating to a new build system takes time. Nevertheless, awesome work!

You plan to replace the autotools build system with this one? I though of this work more as an alternative community-driven build system, as in `p4est`... This indeed increases the completeness and robustness requirements for this PR!

> Yes, t8code is moving very fast and falling behind `main` is a common issue. So we have to keep up with it.

Yes and I think it is a good thing! I was pointing this out because the PR is already starting to get big (400+ LoC), and I am already pretty sure that new unit tests have been added since I did the porting! I think it would be better to merge this more or less as is, while being explicit about the fact that this in not yet production-ready, let people add their tests in both the autotools and cmake build systems, figure out together the edge cases as people start to use it, and improve it incrementally  with additional PRs until we are confident enough to remove the experimental label...

CMakeLists.txt
```diff
@@ -0,0 +1,63 @@
...
+# Prevent `libsc` and `p4est` from overwriting the default install prefix.
+set(CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT FALSE)
+
```
jmark, line 48:
```suggestion

# Rpath options necessary for shared library install to work correctly in user projects.
set(CMAKE_INSTALL_NAME_DIR ${CMAKE_INSTALL_PREFIX}/lib)
set(CMAKE_INSTALL_RPATH ${CMAKE_INSTALL_PREFIX}/lib)
set(CMAKE_INSTALL_RPATH_USE_LINK_PATH true)

```

example/CMakeLists.txt
```diff
@@ -0,0 +1,51 @@
...
+add_library( t8example )
+target_sources( t8example PRIVATE common/t8_example_common.cxx common/t8_example_common_functions.cxx )
+target_include_directories( t8example PRIVATE ${CMAKE_CURRENT_LIST_DIR}/.. )
+target_link_libraries( t8example PRIVATE T8 )
+
```
jmark, line 6:
```suggestion
install( TARGETS t8example DESTINATION ${CMAKE_INSTALL_PREFIX}/lib )

```

test/CMakeLists.txt
```diff
@@ -0,0 +1,74 @@
...
+    set( options "" )
+    set( oneValueArgs "NAME" )  
+    set( multiValueArgs "SOURCES" )
+    cmake_parse_arguments( ADD_T8_TEST "${options}" "${oneValueArgs}" "${multiValueArgs}" ${ARGN} )
+
+    add_executable( ${ADD_T8_TEST_NAME} ${ADD_T8_TEST_SOURCES} )
+    target_link_libraries( ${ADD_T8_TEST_NAME} PRIVATE T8 gtest )
+    add_test( NAME ${ADD_T8_TEST_NAME} COMMAND ./${ADD_T8_TEST_NAME} )
+endfunction()
+
```
jmark, line 14:
```suggestion

# Copy test files to build folder so that the t8_test programs can find them.
function( copy_test_file TEST_FILE_NAME )
    configure_file(${CMAKE_CURRENT_LIST_DIR}/testfiles/${TEST_FILE_NAME} ${CMAKE_CURRENT_BINARY_DIR}/test/testfiles/${TEST_FILE_NAME} COPYONLY)
endfunction()

```

test/CMakeLists.txt
```diff
@@ -0,0 +1,74 @@
...
+add_t8_test( NAME t8_gtest_element_count_leaves  SOURCES t8_gtest_main.cxx t8_schemes/t8_gtest_element_count_leaves.cxx )
+add_t8_test( NAME t8_gtest_element_ref_coords    SOURCES t8_gtest_main.cxx t8_schemes/t8_gtest_element_ref_coords.cxx )
+add_t8_test( NAME t8_gtest_descendant            SOURCES t8_gtest_main.cxx t8_schemes/t8_gtest_descendant.cxx )
+add_t8_test( NAME t8_gtest_find_parent           SOURCES t8_gtest_main.cxx t8_schemes/t8_gtest_find_parent.cxx )
+add_t8_test( NAME t8_gtest_equal                 SOURCES t8_gtest_main.cxx t8_schemes/t8_gtest_equal.cxx )
+add_t8_test( NAME t8_gtest_successor             SOURCES t8_gtest_main.cxx t8_schemes/t8_gtest_successor.cxx )
+add_t8_test( NAME t8_gtest_boundary_extrude      SOURCES t8_gtest_main.cxx t8_schemes/t8_gtest_boundary_extrude.cxx )
+add_t8_test( NAME t8_gtest_face_descendant       SOURCES t8_gtest_main.cxx t8_schemes/t8_gtest_face_descendant.cxx )
+add_t8_test( NAME t8_gtest_default               SOURCES t8_gtest_main.cxx t8_schemes/t8_gtest_default.cxx )
+add_t8_test( NAME t8_gtest_child_parent_face     SOURCES t8_gtest_main.cxx t8_schemes/t8_gtest_child_parent_face.cxx )
```
jmark, line 74:
```suggestion
add_t8_test( NAME t8_gtest_child_parent_face     SOURCES t8_gtest_main.cxx t8_schemes/t8_gtest_child_parent_face.cxx )

copy_test_file( test_cube_unstructured_1.inp )
copy_test_file( test_cube_unstructured_2.inp )
copy_test_file( test_vtk_tri.vtu )
copy_test_file( test_vtk_cube.vtp )
copy_test_file( test_parallel_file.pvtu )
copy_test_file( test_polydata.pvtp )

```

jmark (changes requested):
@dutkalex Please consider these code changes. Other than that LGTM so far!

jmark:
> This looks to me like a "chicken and the egg" situation: if we want to integrate into `main`, we need to have test coverage first, but in order for `main` to continue passing the tests, we need to integrate the code first. I think that since this is purely new stuff that can not break anything else that is already in production, having temporarily uncovered (but obviously flagged as experimental) code in `main` is a better alternative than to break the pipeline for everyone else...

Agree, the changes in this PR are non-intrusive.

> > Migrating to a new build system takes time. Nevertheless, awesome work!
> > Yes, t8code is moving very fast and falling behind `main` is a common issue. So we have to keep up with it.
> 
> Yes and I think it is a good thing! I was pointing this out because the PR is already starting to get big (400+ LoC), and I am already pretty sure that new unit tests have been added since I did the porting! I think it would be better to merge this more or less as is, while being explicit about the fact that this in not yet production-ready, let people add their tests in both the autotools and cmake build systems, figure out together the edge cases as people start to use it, and improve it incrementally with additional PRs until we are confident enough to remove the experimental label...

So, in accordance with the other core developers, here is the plan:

We leave the PR now as it is. A lot of stuff is already working and it is pretty much usable. We mark it as experimental and let it evolve slowly by incrementally adding new features one by one. At some point we can make the final switch and finally get rid of autotools for good.

Thanks @dutkalex for getting this project rolling!

dutkalex:
Good news @jmark ! I have included verbatim your proposed changes as everything looked fine. Don't hesitate to tag me in future PRs about the build system, I'd be happy to continue helping if I can 🙃

src/CMakeLists.txt
```diff
@@ -0,0 +1,199 @@
...
+target_sources( T8 PRIVATE 
+    t8.c 
+    t8_eclass.c 
+    t8_mesh.c 
+    t8_element.c 
+    t8_element_cxx.cxx 
+    t8_element_c_interface.cxx 
+    t8_refcount.c 
+    t8_cmesh/t8_cmesh.c 
+    t8_cmesh/t8_cmesh_occ.cxx
```
jmark, line 70:
```suggestion
    t8_cmesh/t8_cmesh_cad.cxx 
```

src/CMakeLists.txt
```diff
@@ -0,0 +1,199 @@
...
+    t8_cmesh/t8_cmesh_offset.c 
+    t8_cmesh/t8_cmesh_readmshfile.cxx 
+    t8_forest/t8_forest.c 
+    t8_forest/t8_forest_adapt.cxx 
+    t8_geometry/t8_geometry.cxx 
+    t8_geometry/t8_geometry_helpers.c 
+    t8_geometry/t8_geometry_base.cxx 
+    t8_geometry/t8_geometry_with_vertices.cxx 
+    t8_geometry/t8_geometry_implementations/t8_geometry_analytic.cxx 
+    t8_geometry/t8_geometry_implementations/t8_geometry_occ.cxx
```
jmark, line 96:
```suggestion
    t8_geometry/t8_geometry_implementations/t8_geometry_cad.cxx 
```

src/CMakeLists.txt
```diff
@@ -0,0 +1,199 @@
...
+    t8_cmesh_netcdf.h 
+    t8_forest_netcdf.h 
+    t8_element_shape.h 
+    t8_netcdf.h
+    t8_cmesh/t8_cmesh_testcases.h
+    t8_cmesh/t8_cmesh_save.h
+    t8_cmesh/t8_cmesh_examples.h 
+    t8_cmesh/t8_cmesh_geometry.h 
+    t8_cmesh/t8_cmesh_helpers.h 
+    t8_cmesh/t8_cmesh_occ.hxx
```
jmark, line 165:
```suggestion
    t8_cmesh/t8_cmesh_cad.hxx
```

src/CMakeLists.txt
```diff
@@ -0,0 +1,199 @@
...
+    t8_geometry/t8_geometry_base.hxx 
+    t8_geometry/t8_geometry_base.h 
+    t8_geometry/t8_geometry_with_vertices.hxx 
+    t8_geometry/t8_geometry_with_vertices.h 
+    t8_geometry/t8_geometry_helpers.h
+    t8_geometry/t8_geometry_implementations/t8_geometry_linear.h
+    t8_geometry/t8_geometry_implementations/t8_geometry_linear_axis_aligned.h
+    t8_geometry/t8_geometry_implementations/t8_geometry_analytic.hxx
+    t8_geometry/t8_geometry_implementations/t8_geometry_examples.h
+    t8_geometry/t8_geometry_implementations/t8_geometry_occ.h
```
jmark, line 188:
```suggestion
    t8_geometry/t8_geometry_implementations/t8_geometry_cad.h
```

src/CMakeLists.txt
```diff
@@ -0,0 +1,199 @@
...
+    t8_geometry/t8_geometry_implementations/t8_geometry_occ.hxx
```
jmark, line 189:
```suggestion
    t8_geometry/t8_geometry_implementations/t8_geometry_cad.hxx
```

test/CMakeLists.txt
```diff
@@ -0,0 +1,86 @@
...
+function( copy_test_file TEST_FILE_NAME )
+    configure_file(${CMAKE_CURRENT_LIST_DIR}/testfiles/${TEST_FILE_NAME} ${CMAKE_CURRENT_BINARY_DIR}/test/testfiles/${TEST_FILE_NAME} COPYONLY)
+endfunction()
+
+add_t8_test( NAME t8_gtest_cmesh_bcast       SOURCES t8_gtest_main.cxx t8_cmesh/t8_gtest_bcast.cxx )
+add_t8_test( NAME t8_gtest_eclass            SOURCES t8_gtest_main.cxx t8_gtest_eclass.cxx )
+add_t8_test( NAME t8_gtest_vec               SOURCES t8_gtest_main.cxx t8_gtest_vec.cxx )
+add_t8_test( NAME t8_gtest_mat               SOURCES t8_gtest_main.cxx t8_gtest_mat.cxx )
+add_t8_test( NAME t8_gtest_refcount          SOURCES t8_gtest_main.cxx t8_gtest_refcount.cxx )
+add_t8_test( NAME t8_gtest_occ_linkage       SOURCES t8_gtest_main.cxx t8_gtest_occ_linkage.cxx )
```
jmark, line 25:
```suggestion
add_t8_test( NAME t8_gtest_cad_linkage       SOURCES t8_gtest_main.cxx t8_gtest_cad_linkage.cxx )
```

test/CMakeLists.txt
```diff
@@ -0,0 +1,86 @@
...
+add_t8_test( NAME t8_gtest_ghost_delete              SOURCES t8_gtest_main.cxx t8_forest/t8_gtest_ghost_delete.cxx )
+add_t8_test( NAME t8_gtest_ghost_and_owner           SOURCES t8_gtest_main.cxx t8_forest/t8_gtest_ghost_and_owner.cxx )
+
+add_t8_test( NAME t8_gtest_permute_hole      SOURCES t8_gtest_main.cxx t8_forest_incomplete/t8_gtest_permute_hole.cxx )
+add_t8_test( NAME t8_gtest_recursive         SOURCES t8_gtest_main.cxx t8_forest_incomplete/t8_gtest_recursive.cxx )
+add_t8_test( NAME t8_gtest_iterate_replace   SOURCES t8_gtest_main.cxx t8_forest_incomplete/t8_gtest_iterate_replace.cxx )
+add_t8_test( NAME t8_gtest_empty_local_tree  SOURCES t8_gtest_main.cxx t8_forest_incomplete/t8_gtest_empty_local_tree.cxx )
+add_t8_test( NAME t8_gtest_empty_global_tree SOURCES t8_gtest_main.cxx t8_forest_incomplete/t8_gtest_empty_global_tree.cxx )
+
+add_t8_test( NAME t8_gtest_geometry_occ  SOURCES t8_gtest_main.cxx t8_geometry/t8_geometry_implementations/t8_gtest_geometry_occ.cxx )
```
jmark, line 59:
```suggestion
add_t8_test( NAME t8_gtest_geometry_cad  SOURCES t8_gtest_main.cxx t8_geometry/t8_geometry_implementations/t8_gtest_geometry_cad.cxx )
```

jmark (changes requested):
@dutkalex One final edit to do. There were some file renamings just yesterday. Good timing! Sorry. But with this we can finally merge.

dutkalex:
Done. @jmark

jmark:
Just for the record: https://github.com/DLR-AMR/t8code/issues/954

jmark (approved):
Let's do it!

PR #944, opened 2024-02-16
CAD instead of OCC

jfussbro:

With this PR we switch our extern geometry naming from occ to cad, in order to keep it more general.

#### General

#### Tests

#### Github action

#### Scripts and Wiki

#### Licence


.github/workflows/tests_sc_p4est.yml
```diff
@@ -103,7 +103,7 @@ jobs:
       run: echo No cache found or cache will be ignored. SC_IGNORE_CACHE=$SC_IGNORE_CACHE
     - name: if ignore cache, delete folders
       if: ${{ env.SC_IGNORE_CACHE == 1 }}
-      # The true at the end is to ignore errors that i.e. occur when the folders do not exist
+      # The true at the end is to ignore errors that i.e. cadur when the folders do not exist
```
jmark, line 106: @jfussbro This is not good ... please check for this before making a PR. Most text substitution tools have a parameter/switch where you tell it to only replace stand-alone words. What tool are you using?

```suggestion
      # The true at the end is to ignore errors that i.e. occur when the folders do not exist
```

jmark (changes requested):

.github/workflows/tests_sc_p4est.yml
jfussbro: I'm sorry. I didn't think about that. I used find and replace in VS code.

.github/workflows/tests_t8code_linkage_parallel_debug.yml
```diff
@@ -205,25 +205,25 @@ jobs:
     - name: check OpenCASCADE
       run: echo "Checking OpenCASCADE with MPI in debugging mode"
     - name: configure MPI OpenCASCADE debug
-      run: mkdir build_occ && cd build_occ && ../configure $CONFIG_DEBUG --with-occ CFLAGS="$CFLAGS_var -I/usr/include/opencascade" CXXFLAGS="$CXXFLAGS_var -I/usr/include/opencascade"
+      run: mkdir build_cad && cd build_cad && ../configure $CONFIG_DEBUG --with-cad CFLAGS="$CFLAGS_var -I/usr/include/opencascade" CXXFLAGS="$CXXFLAGS_var -I/usr/include/opencascade"
```
jmark, line 208: How can this even work when the `configure` script does not accept `--with-cad`?

https://github.com/DLR-AMR/t8code/blob/44d96ab2dd2224c9a4cd935c47ef4eb108cbe0a9/config/t8_occ.m4#L15

.github/workflows/tests_t8code_linkage_parallel_debug.yml
jmark, line 1: Is it necessary to rename to `cad`? We are linking against `occ` here, right?

jmark (changes requested):

jmark (approved):

PR #951, opened 2024-02-20
Skip vtk reader test for MPI sizes other than 2.

jmark:

This PR fixes a silently failing test by crashing for mpisizes other than two. The test actually assumes two ranks by design.
Autotools ignores the crash while cmake marks this as a failed test. This is the error message when calling with one rank:
```
Abort(469312775) on node 0 (rank 0 in comm 0): Fatal error in internal_Bcast: Invalid root, error stack:
internal_Bcast(107): MPI_Bcast(buffer=0x7ffc3584882c, count=1, MPI_INT, 1, MPI_COMM_WORLD) failed
internal_Bcast(61).: Invalid root (value given was 1)
```

Proposed solution: Just skip it for mpisizes other than two. Note, two MPI ranks is the default within the Github CI. So this test is properly covered there.

#### General

#### Tests

#### Github action

#### Scripts and Wiki

#### Licence


test/t8_IO/t8_gtest_vtk_reader.cxx
```diff
@@ -44,7 +44,7 @@ class vtk_reader: public testing::TestWithParam<std::tuple<int, int, int>> {
     SC_CHECK_MPI (mpiret);
     mpiret = sc_MPI_Comm_rank (sc_MPI_COMM_WORLD, &mpirank);
     SC_CHECK_MPI (mpiret);
-    if (mpisize > T8_VTK_TEST_NUM_PROCS) {
+    if (mpisize != T8_VTK_TEST_NUM_PROCS) {
```
lukasdreyer, line 47: Could you add a comment on why only mpisize = T8_VTK_TEST_NUM_PROCS is allowed?
Something along the lines of
`/** mpisize must fit to the number of files that are read in */`

lukasdreyer (changes requested):

test/t8_IO/t8_gtest_vtk_reader.cxx
jmark: Done!

lukasdreyer (approved):

PR #952, opened 2024-02-21
Feature scheme mempool

lukasdreyer:
Currently, all of the base functionality for the different schemes is collected in `t8_default_scheme_common_c`.
Since I want to introduce new schemes that would benefit from having the mempool functionality available, this PR seperates this functionality into its own base class, so that other classes can inherit from it.

Additionally, the scheme base until now had a void* ts_context, that was only ever used by the mempool. Therefore, it was moved to the derived class and renamed.

#### General

#### Tests

#### Github action

#### Scripts and Wiki

#### Licence

### DIFF
--- a/src/t8_schemes/t8_scheme_mempool_cxx.hxx
+++ b/src/t8_schemes/t8_scheme_mempool_cxx.hxx
@@ -36,28 +36,30 @@
 #define T8_SCHEME_IS_TYPE(VAR, TYPE) ((dynamic_cast<TYPE> (VAR)) != NULL)
 
 class t8_scheme_mempool_c: public t8_eclass_scheme_c {
+ private:
+  sc_mempool_t *mempool;
+
  public:
   /** Destructor for all default schemes */
   virtual ~t8_scheme_mempool_c ()
   {
-    T8_ASSERT (ts_context != NULL);
-    SC_ASSERT (((sc_mempool_t *) ts_context)->elem_count == 0);
-    sc_mempool_destroy ((sc_mempool_t *) ts_context);
+    T8_ASSERT (mempool != NULL);
+    SC_ASSERT (mempool->elem_count == 0);
+    sc_mempool_destroy (mempool);
   }
   t8_scheme_mempool_c (t8_eclass_t eclass_in, int elem_size)
   {
     element_size = elem_size;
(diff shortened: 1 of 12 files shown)


src/t8_schemes/t8_scheme_mempool_cxx.hxx
```diff
@@ -0,0 +1,100 @@
...
+
+#include <t8_element_cxx.hxx>
+#include <t8_element.h>
+#include <sc_functions.h>
+
+/* Macro to check whether a pointer (VAR) to a base class, comes from an
+ * implementation of a child class (TYPE). */
+#define T8_SCHEME_IS_TYPE(VAR, TYPE) ((dynamic_cast<TYPE> (VAR)) != NULL)
+
+class t8_scheme_mempool_c: public t8_eclass_scheme_c {
```
jmark, line 38:
```suggestion
class t8_eclass_scheme_with_mempool: public t8_eclass_scheme {
```
This is just a suggestion. Let me know if you have a different opinion on this.

jmark: What is the meaning of the suffix `_c`? "common" or does it have something to do with the language "C"?

src/t8_schemes/t8_scheme_mempool_cxx.hxx
```diff
@@ -0,0 +1,100 @@
...
+#ifndef T8_SCHEME_COMMON_CXX
+#define T8_SCHEME_COMMON_CXX
+
+#include <t8_element_cxx.hxx>
+#include <t8_element.h>
+#include <sc_functions.h>
+
+/* Macro to check whether a pointer (VAR) to a base class, comes from an
+ * implementation of a child class (TYPE). */
+#define T8_SCHEME_IS_TYPE(VAR, TYPE) ((dynamic_cast<TYPE> (VAR)) != NULL)
```
jmark, line 36: Where is this macro used? Is this specific to the mempool class?

jmark (commented):
@lukasdreyer As we discussed yesterday. I like the general idea! I just have some minor suggestions. 

I think it is a good idea to have a second opinion on such a kind of code change. Keep me in the loop about any discussion about this PR and the related ones.

@Davknapp

PR #977, opened 2024-03-06
update number of the corner of the second vertex

Davknapp:
The corner number was used wrongly in this example. We were using the vertex number of the vertices given to the tree instead of the vertices of the element. 

#### General

#### Tests

#### Github action

#### Scripts and Wiki

#### Licence


lukasdreyer (changes requested):
I'd love to see a test that would have caught this

Davknapp:
> I'd love to see a test that would have caught this

I've extended the point inside check, such that it would've failed.
During debugging I also found another bug, which lead to a failed vtk-output for axis aligned geometries. To catch this bug I opend up the issue #980 , because I think that it requires more detailed testing which would be out of the scope of this PR.

test/t8_geometry/t8_geometry_implementations/t8_gtest_geometry_linear.cxx
```diff
@@ -25,88 +25,142 @@
...
+  void
+  SetUp () override
+  {
+    const int geom_int = std::get<0> (GetParam ());
+    eclass = std::get<1> (GetParam ());
+    t8_cmesh_init (&cmesh);
+    if (geom_int == T8_GEOMETRY_TYPE_LINEAR_AXIS_ALIGNED
+        && !(eclass == T8_ECLASS_LINE || eclass == T8_ECLASS_QUAD || eclass == T8_ECLASS_HEX)) {
+      GTEST_SKIP ();
+    }
```
lukasdreyer, line 57: I would rather just initialize the testsuite only with the line, quad and hex eclass

test/t8_geometry/t8_geometry_implementations/t8_gtest_geometry_linear.cxx
```diff
@@ -25,88 +25,142 @@
...
+      vertices[3] = vertices[3 * (num_vertices - 1)];
+      vertices[4] = vertices[3 * (num_vertices - 1) + 1];
+      vertices[5] = vertices[3 * (num_vertices - 1) + 2];
+
+      t8_cmesh_register_geometry<t8_geometry_linear_axis_aligned> (cmesh, t8_eclass_to_dimension[eclass]);
+      break;
+    default:
+      break;
+    }
+    t8_cmesh_set_tree_vertices (cmesh, 0, vertices, t8_eclass_num_vertices[eclass]);
```
lukasdreyer, line 84: For the axis aligned geometry, only two vertices should be set

test/t8_geometry/t8_gtest_point_inside.cxx
```diff
@@ -301,6 +303,12 @@ TEST_P (geometry_point_inside, test_point_inside)
 
         num_in += point_is_inside ? 1 : 0;
       }
+      for (int side = 0; side < min_points_outside; ++side) {
+        test_point[3 * (num_points + side)] = (side == 0) ? -2.0 : (side == 1) ? 2.0 : -0.1;
+        test_point[3 * (num_points + side) + 1] = (side == 2) ? -2.0 : (side == 3) ? 2.0 : 0.3;
+        test_point[3 * (num_points + side) + 2] = (side == 4) ? -2.0 : (side == 5) ? 2.0 : 0.15;
+        point_is_inside[num_points + side] = false;
+      }
```
lukasdreyer, line 311: What is happening here? Please document

lukasdreyer (changes requested):

test/t8_geometry/t8_geometry_implementations/t8_gtest_geometry_linear.cxx
Davknapp: But for the linear geometry we want to test all eclasses.

lukasdreyer (approved):

PR #1071, opened 2024-05-21
Catch and fix negative volumes of triangles/quads in 2D Gmsh files.

jmark:

This PR checks for negative volumes of triangles and quads and corrects this when reading Gmsh files. "Positive" 2D shapes are defined by the cross product of the two sides of the triangle/quad pointing towards the same direction as the positive z-direction. Of course, this only works when the quad/triangle lies in the x-y plane.

You can use this Gmsh files with negative quads for testing: https://gist.githubusercontent.com/jmark/bfe0d45f8e369298d6cc637733819013/raw/cecf86edecc736e8b3e06e354c494b2052d41f7a/rectangle_with_negative_volumes.msh

~~This PR has one problem though: It breaks the Lagrange-Geometry handler for Hexes since some faces of a hex (which are quads) are created with negative volumes. I hope this can be easily fixed. This has probably something to do with the fact that the faces resp. quads do not lie in the x-y-plane anymore.~~

#### General

#### Tests

#### Github action

#### Scripts and Wiki

#### Licence

### DIFF
--- a/src/t8_cmesh/t8_cmesh.cxx
+++ b/src/t8_cmesh/t8_cmesh.cxx
@@ -426,12 +426,21 @@ t8_cmesh_get_partition_table (t8_cmesh_t cmesh)
 }
 
 void
-t8_cmesh_set_dimension (t8_cmesh_t cmesh, int dim)
+t8_cmesh_set_dimension (t8_cmesh_t cmesh, const int dim)
 {
   T8_ASSERT (t8_cmesh_is_initialized (cmesh));
   T8_ASSERT (0 <= dim && dim <= T8_ECLASS_MAX_DIM);
 
   cmesh->dimension = dim;
+}
+
+int
+t8_cmesh_get_dimension (const t8_cmesh_t cmesh)
+{
+  T8_ASSERT (t8_cmesh_is_committed (cmesh));
+  T8_ASSERT (0 <= cmesh->dimension && cmesh->dimension <= T8_ECLASS_MAX_DIM);
+
+  return cmesh->dimension;
 }
 
 void
--- a/src/t8_cmesh/t8_cmesh.cxx
+++ b/src/t8_cmesh/t8_cmesh.cxx
@@ -472,32 +472,30 @@ t8_cmesh_tree_vertices_negative_volume (const t8_eclass_t eclass, const double *
 {
   T8_ASSERT (num_vertices == t8_eclass_num_vertices[eclass]);
 
-  /* Points and lines do not have a volume. */
+  /* Points and lines do not have a volume orientation. */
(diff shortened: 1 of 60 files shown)


jmark:
@sandro-elsweijer @CsatiZoltan Would be please take a look at this PR. Is there a fix for the Lagrange geometry handler?

jmark:
> @sandro-elsweijer @CsatiZoltan Would be please take a look at this PR? Is there a fix for the Lagrange geometry handler?

Nevermind. I just got an idea. Let me try that out first.

CsatiZoltan:
@jmark The Lagrange geometry elements were designed to have positive face orientations, i.e. with outward unit vector.

jmark:
> @jmark The Lagrange geometry elements were designed to have positive face orientations, i.e. with outward unit vector.
> 
> > This has probably something to do with the fact that the faces resp. quads do not lie in the x-y-plane anymore.
> 
> I think so.
> 
> > This PR checks for negative volumes of triangles and quads. That is, the cross product of the two sides of the triangle/quad must point towards the same direction as the positive z-direction.
> 
> What you test for is negative _area_, no? If your goal is to test positive orientation for tri/quad, then you want to work in 2D. In that case, in my opinion, you do not have to test the faces of 3D elements. Those should rather be subject to a positivity test for volumes (in which case, you could use the triple product of three element edges). Do I miss something?

@CsatiZoltan You got it right. The negative volume test for 2D shapes makes only really sense in the x-y plane. Anyway,
I removed the check in the validation routines. So no code breaking anymore.

Instead I only check for "negative" 2D elements when reading 2D Gmsh files. This is the new purpose of this PR.

src/t8_cmesh/t8_cmesh.cxx
```diff
@@ -479,22 +479,35 @@ t8_cmesh_set_tree_class (t8_cmesh_t cmesh, t8_gloidx_t gtree_id, t8_eclass_t tre
...
+   *      z             For 2D meshes we enforce the right-hand-rule in terms
+   *      |             of node ordering. The volume is defined by the parallelepiped
+   *      | 2- - -(3)   spanned by the vectors between nodes 0:1 and 0:2 as well as the
+   *      |/____ /      unit vector in z-direction. This definition works for both triangles and quads.
+   *      0     1
+   *
+   *      6 ______  7   For Hexes and pyramids, if the vertex 4 is below the 0-1-2-3 plane,
    *       /|     /     the volume is negative. This is the case if and only if
    *    4 /_____5/|     the scalar product of v_4 with the cross product of v_1 and v_2 is
    *      | | _ |_|     smaller 0:
```
Prasanna-Ponnusamy, line 513: smaller than 0

Prasanna-Ponnusamy (approved):

PR #1115, opened 2024-07-09
Simplified C interface for registering geometries in cmeshes.

jmark:

I basically just switched from a pointer-pointer to just a pointer. I see no reason why an application should manage a reference to the pointer of a geometry. Besides, I'm having troubles with the garbage collector in Julia wiping the reference to the geometry away.

#### General

#### Tests

#### Github action

#### Scripts and Wiki

#### Licence

### DIFF
--- a/src/t8_cmesh.h
+++ b/src/t8_cmesh.h
@@ -69,6 +69,12 @@ T8_EXTERN_C_BEGIN ();
 void
 t8_cmesh_init (t8_cmesh_t *pcmesh);
 
+/** Allocate a new un-commmited cmesh.
+ * \return                     A pointer to an un-committed t8_cmesh structure.
+ */
+t8_cmesh_t
+t8_cmesh_new ();
+
 /** Check whether a cmesh is not NULL, initialized and not committed.
  * In addition, it asserts that the cmesh is consistent as much as possible.
  * \param [in] cmesh            This cmesh is examined.  May be NULL.
@@ -404,7 +410,7 @@ t8_cmesh_reorder (t8_cmesh_t cmesh, sc_MPI_Comm comm);
  *       registered \a after the bcast operation, not before.
  */
 void
-t8_cmesh_register_geometry (t8_cmesh_t cmesh, t8_geometry_c **geometry);
+t8_cmesh_register_geometry (t8_cmesh_t cmesh, t8_geometry_c *geometry);
 
 /** Set the geometry for a tree, thus specify which geometry to use for this tree.
  * \param [in] cmesh     A non-committed cmesh.
--- a/src/t8_cmesh.h
+++ b/src/t8_cmesh.h
@@ -69,7 +69,7 @@ T8_EXTERN_C_BEGIN ();
 void
 t8_cmesh_init (t8_cmesh_t *pcmesh);
 
-/** Allocate a new un-commmited cmesh.
+/** Allocate a new un-committed cmesh.
  * \return                     A pointer to an un-committed t8_cmesh structure.
  */
 t8_cmesh_t
--- a/src/t8_cmesh/t8_cmesh.cxx
+++ b/src/t8_cmesh/t8_cmesh.cxx
@@ -179,6 +179,14 @@ t8_cmesh_init (t8_cmesh_t *pcmesh)
   T8_ASSERT (t8_cmesh_is_initialized (cmesh));
 }
 
+t8_cmesh_t
+t8_cmesh_new ()
+{
+  t8_cmesh_t cmesh;
+  t8_cmesh_init (&cmesh);
+  return cmesh;
+}
+
 void
 t8_cmesh_set_derive (t8_cmesh_t cmesh, t8_cmesh_t set_from)
 {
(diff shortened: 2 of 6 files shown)


sandro-elsweijer:
It was implemented this way, because the geometry handler takes ownership of the geometry and puts it inside a unique ptr. Depending on the unique pointer, the location of the geometry could change and this way the reference provided by the user would point into nothingness. This is unexpected behavior and really hard to debug.
This could lead to troubles if the user wants to use some functionality of the geometry after assigning it to the cmesh. This for example happens with curved geometries, when the geometry is asked for topology information.

jmark:
> It was implemented this way, because the geometry handler takes ownership of the geometry and puts it inside a unique ptr. Depending on the unique pointer, the location of the geometry could change and this way the reference provided by the user would point into nothingness. This is unexpected behavior and really hard to debug. This could lead to troubles if the user wants to use some functionality of the geometry after assigning it to the cmesh. This for example happens with curved geometries, when the geometry is asked for topology information.

@sandro-elsweijer Thank you for your explanation! Appreciated!

The reasoning why you did it this way in general is already clear to me. However, in my opinion, your approach s not well thought through by breaking well established software design patterns.

The already existing documentation is crystal clear about the expected behavior:
```
/** Register a geometry in the cmesh. The cmesh takes ownership of the geometry.
```
The user cannot expect to properly use the geometry anymore, once it has given up ownership to t8code.

If you really want to provide use cases where the user and t8code share the geometry then we have to think about different approaches. For example, having a two copies of the geometry. If the geometry is expensive than the copies of the geometry should share the expensive routines/data.

sandro-elsweijer:
> > It was implemented this way, because the geometry handler takes ownership of the geometry and puts it inside a unique ptr. Depending on the unique pointer, the location of the geometry could change and this way the reference provided by the user would point into nothingness. This is unexpected behavior and really hard to debug. This could lead to troubles if the user wants to use some functionality of the geometry after assigning it to the cmesh. This for example happens with curved geometries, when the geometry is asked for topology information.
> 
> @sandro-elsweijer Thank you for your explanation! Appreciated!
> 
> The reasoning why you did it this way in general is already clear to me. However, in my opinion, your approach is not well thought through by breaking well established software design patterns.
> 
> The already existing documentation is crystal clear about the expected behavior:
> 
> ```
> /** Register a geometry in the cmesh. The cmesh takes ownership of the geometry.
> ```
> 
> The user cannot expect to properly use the geometry anymore, once it has given up ownership to t8code.
> 
> If you really want to provide use cases where the user and t8code share the geometry then we have to think about different approaches. For example, having two copies of the geometry. If the geometry is expensive than the copies of the geometry should share the expensive routines/data.

The two copy option is especially with OCCs memory management really hard.
But since this is just the C interface and the geometry classes members cannot be accessed in C anyway, we can do it this way. Users who want to use the geometry afterwards can just use the CPP interface.
But please adapt the [register_geometry](https://github.com/DLR-AMR/t8code/blob/b64c7eb021f1cda7d5753b70d516a1798c010f6e/src/t8_geometry/t8_geometry_handler.cxx#L38-L43) member function of the handler as well.

sandro-elsweijer (changes requested):
> > It was implemented this way, because the geometry handler takes ownership of the geometry and puts it inside a unique ptr. Depending on the unique pointer, the location of the geometry could change and this way the reference provided by the user would point into nothingness. This is unexpected behavior and really hard to debug. This could lead to troubles if the user wants to use some functionality of the geometry after assigning it to the cmesh. This for example happens with curved geometries, when the geometry is asked for topology information.
> 
> @sandro-elsweijer Thank you for your explanation! Appreciated!
> 
> The reasoning why you did it this way in general is already clear to me. However, in my opinion, your approach is not well thought through by breaking well established software design patterns.
> 
> The already existing documentation is crystal clear about the expected behavior:
> 
> ```
> /** Register a geometry in the cmesh. The cmesh takes ownership of the geometry.
> ```
> 
> The user cannot expect to properly use the geometry anymore, once it has given up ownership to t8code.
> 
> If you really want to provide use cases where the user and t8code share the geometry then we have to think about different approaches. For example, having two copies of the geometry. If the geometry is expensive than the copies of the geometry should share the expensive routines/data.

The two copy option is especially with OCCs memory management really hard.
But since this is just the C interface and the geometry classes members cannot be accessed in C anyway, we can do it this way. Users who want to use the geometry afterwards can just use the CPP interface.
But please adapt the [register_geometry](https://github.com/DLR-AMR/t8code/blob/b64c7eb021f1cda7d5753b70d516a1798c010f6e/src/t8_geometry/t8_geometry_handler.cxx#L38-L43) member function of the handler as well.

jmark:
> The two copy option is especially with OCCs memory management really hard. But since this is just the C interface and the geometry classes members cannot be accessed in C anyway, we can do it this way. Users who want to use the geometry afterwards can just use the CPP interface. But please adapt the [register_geometry](https://github.com/DLR-AMR/t8code/blob/b64c7eb021f1cda7d5753b70d516a1798c010f6e/src/t8_geometry/t8_geometry_handler.cxx#L38-L43) member function of the handler as well.

Fair enough! Adapted the `register_geometry` member function as you said.

sandro-elsweijer (approved):
lgtm!

PR #1162, opened 2024-07-18
Enhancement #1089: Replace the implemented binary search by a binary search from the standard library

### LINKED ISSUE #1089: Improvement: Replace t8_forest_bin_search_lower with std::lower_bound
The title more or less says it all.
Check whether the functionality provided by t8_forest_bin_search_lower can be replaced by a call to std::lower_bound and if yes, do so.

Niklas997:

A bidirectional iterator that wraps the serialized content of a `t8_element_array_t` has been implemented as well as functions providing a `begin()` and an `end()` iterator to a `t8_element_array_t.`
This allows a convenient usage of algorithms from the standard library like std::lower_bound, etc. on `t8_element_array_t`'s.

The implemented binary search in `t8_forest_bin_lower` has been replaced by a std::upper_bound search utilizing the before mentioned iterators providing the same output as the initial implementation.

The "discussion" and "help wanted" labels result from the fact, that this is more of a pseudo-iterator since the original data is serialized to char-bytes in the internal sc_array_t and I am not sure if this is the best solution. A complete makover of the t8_element_array_t class would be of course a nicer but more complex solution than the presented approch for closing the issue.

Closes #1089 

#### General

#### Tests

#### Github action

#### Scripts and Wiki

#### Licence

### DIFF
--- a/src/t8_data/t8_element_array_iterator.hxx
+++ b/src/t8_data/t8_element_array_iterator.hxx
@@ -1,0 +1,191 @@
+/*
+  This file is part of t8code.
+  t8code is a C library to manage a collection (a forest) of multiple
+  connected adaptive space-trees of general element classes in parallel.
+
+  Copyright (C) 2015 the developers
+
+  t8code is free software; you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation; either version 2 of the License, or
+  (at your option) any later version.
+
+  t8code is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with t8code; if not, write to the Free Software Foundation, Inc.,
+  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+*/
+
+/** \file t8_element_array_iterator.hxx
+ * We define a pseudo-bidirectional-iterator for the t8_element_array that stores elements of a given
+ * eclass scheme.
+ */
+
+#ifndef T8_ELEMENT_ARRAY_ITERATOR_HXX
+#define T8_ELEMENT_ARRAY_ITERATOR_HXX
+
+#include <t8.h>
+#include <t8_element.h>
+#include <t8_data/t8_containers.h>
+
+#include <cstddef>
+#include <iterator>
+
+T8_EXTERN_C_BEGIN ();
+
+/**
+ * @brief This struct resembles a bidirectional-iterator wrapper for the content of an \a t8_element_array_t.
+ * The iterators can be dereferenced in order to receive an \a t8_element_t pointer to the corresponding index.
+ * These iterators may be used in algorithms of the standard library (e.g. std::lower_bound,...) or to iterate
+ * through an \a t8_element_array_t.
+ * Since the actual data is serialized to char-bytes in the underlying array, we have to store an element pointer
+ *  \a dref_element_ resembling the actual (serialized) element withint the iterator class and cannot return a real
+ *  reference of the element pointer residing in the underlying array. Therefore, read-only operations are possible.
+ */
+class t8_element_array_iterator {
+ public:
+  using iterator_category = std::bidirectional_iterator_tag;
+  using difference_type = std::ptrdiff_t;
+  using pointer = t8_element_t**;
+  using value_type = t8_element_t*;
+  using reference = t8_element_t*&;
+
+  /* Constructors */
+  t8_element_array_iterator () = delete;
+  t8_element_array_iterator (const t8_element_array_t* element_array, const t8_locidx_t position)
+    : scheme_ { t8_element_array_get_scheme (element_array) }, array_ { t8_element_array_get_array (element_array) },
+      index_ { position } {};
+
+  /* Copy/Move Constructors/Assignment-Operators */
+  t8_element_array_iterator (const t8_element_array_iterator& other) = default;
+  t8_element_array_iterator&
+  operator= (const t8_element_array_iterator& other)
+    = default;
+  t8_element_array_iterator (t8_element_array_iterator&& other) = default;
+  t8_element_array_iterator&
+  operator= (t8_element_array_iterator&& other)
+    = default;
+
+  /* Destructor */
+  ~t8_element_array_iterator () = default;
+
+  /* Dereferencing operator of the iterator wrapper, returning a reference to const of the t8_element_t-pointer to the serialized bytes representing the actual serialized t8_element_t pointer */
+  const reference
+  operator* ()
+  {
+    T8_ASSERT (index_ >= 0 && static_cast<size_t> (index_) < array_->elem_count);
+    dref_element_ = static_cast<t8_element_t*> (t8_sc_array_index_locidx (array_, index_));
+    return dref_element_;
+  };
+
+  /* Pre- and Postfix increment */
+  t8_element_array_iterator&
+  operator++ ()
+  {
+    ++index_;
+    return *this;
+  };
+  t8_element_array_iterator
+  operator++ (int)
+  {
+    t8_element_array_iterator tmp_iterator (*this);
+    ++(*this);
+    return tmp_iterator;
+  };
+
+  /* Pre- and Postfix decrement */
+  t8_element_array_iterator&
+  operator-- ()
+  {
+    --index_;
+    return *this;
+  };
+  t8_element_array_iterator
+  operator-- (int)
+  {
+    t8_element_array_iterator tmp_iterator (*this);
+    --(*this);
+    return tmp_iterator;
+  };
+
+  /* Return the index within the array the iterator currently points to [0,...,size] */
+  t8_locidx_t
+  GetArrayIndex () const
+  {
+    return index_;
+  };
+
+  /* Compute the linear id at a given level for the element the iterator points to */
+  t8_linearidx_t
+  GetLinearIDAtLevel (const int level)
+  {
+    T8_ASSERT (index_ >= 0 && static_cast<size_t> (index_) < array_->elem_count);
+    dref_element_ = static_cast<t8_element_t*> (t8_sc_array_index_locidx (array_, index_));
+    return scheme_->t8_element_get_linear_id (dref_element_, level);
+  };
+
+  /* Comparison operators */
+  friend bool
+  operator== (const t8_element_array_iterator& iter1, const t8_element_array_iterator& iter2)
+  {
+    return (iter1.array_->array == iter2.array_->array && iter1.index_ == iter2.index_);
+  };
+  friend bool
+  operator!= (const t8_element_array_iterator& iter1, const t8_element_array_iterator& iter2)
+  {
+    return (iter1.array_->array != iter2.array_->array || iter1.index_ != iter2.index_);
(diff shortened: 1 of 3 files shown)


src/t8_data/t8_element_array_iterator.hxx
```diff
@@ -0,0 +1,191 @@
...
+    return index_;
+  };
+
+  /* Compute the linear id at a given level for the element the iterator points to. */
+  t8_linearidx_t
+  GetLinearIDAtLevel (const int level)
+  {
+    T8_ASSERT (index_ >= 0 && static_cast<size_t> (index_) < array_->elem_count);
+    dref_element_ = static_cast<t8_element_t*> (t8_sc_array_index_locidx (array_, index_));
+    return scheme_->t8_element_get_linear_id (dref_element_, level);
```
jmark, line 128: Does this work?
```suggestion
    return scheme_->t8_element_get_linear_id (*(*this), level);
```

jmark (changes requested):
@Niklas997 Always a treat reading your code! I'm learning new quirks of C++ every time from you. I posted some comments/ideas. Let me know what you think.

src/t8_data/t8_element_array_iterator.hxx
```diff
@@ -0,0 +1,191 @@
...
+  t8_element_array_iterator&
+  operator-= (const difference_type n)
+  {
+    index_ -= n;
+    T8_ASSERT (index_ < 0);
+    return *this;
+  }
+
+ private:
+  const t8_eclass_scheme_c* scheme_; /*!< The scheme of the elements residing within the array. */
```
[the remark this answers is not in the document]
Niklas997: Yes, I like the style of adding an underscore as a suffix to class members. I think it helps distinguishing between class members and function parameters and local variables in member functions. This convention is for example stated in Google's C++ naming conventions https://google.github.io/styleguide/cppguide.html#General_Naming_Rules and in many cases I like to stick with them.
But I have committed your proposed changes to the naming of the variables. I think it fits better to the current style used in t8code.

src/t8_data/t8_element_array_iterator.hxx
```diff
@@ -0,0 +1,191 @@
...
+  const sc_array_t* array_;          /*!< A pointer to the actual serialized array of element pointers. */
+  t8_locidx_t index_ { 0 };          /*!< The index the iterator currently points to. */
+  t8_element_t* dref_element_; /*!< A helper variable for de-serializing the actual element pointers in the array. */
```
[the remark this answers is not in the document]
Niklas997: Yes, this was a point I did not like in the implementation but somehow I wanted to return a (const) reference from the dereferencing operator and this was not possible, because due to casting of the underlying bytes we created a temporary object to which we cannot return a reference, therefore, there was the need to store  a t8_element_t pointer as a member of the class. However, now the return type of the dereferencing operator changed to a value_type and therefore, we could get rid of the intermediate process and the class member holding the t8_element_t pointer. I think now it is a nice solution.

Niklas997:
> @Niklas997 Always a treat reading your code! I'm learning new quirks of C++ every time from you. I posted some comments/ideas. Let me know what you think.

Thank you! That is very nice to hear! And thanks for the input. Now it seems to be a nice substitution for the previously implemented binary search in t8_forest_bin_lower

src/t8_data/t8_element_array_iterator.hxx
Niklas997: Yes, this would work and with the updated dereference operator it works nicely together

src/t8_forest/t8_forest.cxx
```diff
@@ -1438,50 +1441,41 @@ t8_forest_copy_trees (t8_forest_t forest, t8_forest_t from, int copy_elements)
...
-  return low;
+  /* A typedef for the value type of the t8_element_array_t iterator. */
+  using element_ptr_t = t8_element_t *;
+
+  /* We search for the first element in the array that is greater than the given element id. */
+  auto elem_iter
+    = std::upper_bound (t8_element_array_begin (elements), t8_element_array_end (elements), element_id,
+                        [&maxlevel, &ts] (const t8_linearidx_t element_id_, const element_ptr_t &elem_ptr) {
+                          const t8_linearidx_t elem_id = ts->t8_element_get_linear_id (elem_ptr, maxlevel);
+                          return (element_id_ < elem_id);
```
jmark, line 1470:
```suggestion
                        [&maxlevel, &ts] (const t8_linearidx_t element_id, const t8_element_array_iterator::value_type elem_ptr) {
                          return element_id < ts->t8_element_get_linear_id (elem_ptr, maxlevel);
```

src/t8_forest/t8_forest.cxx
```diff
@@ -1438,50 +1441,41 @@ t8_forest_copy_trees (t8_forest_t forest, t8_forest_t from, int copy_elements)
...
+                        });
+
+  /* After we found the element with an id greater than the given one, we are able to jump one index back.
+   * This guarantees us that the element at (index - 1) is smaller or equal to the given element id.
+   * In case we do not find an element that is greater than the given element_id, the binary search returns
+   * the end-iterator of the element array. In that case, we want to return the last index from the element
+   * array. */
+  return elem_iter.GetArrayIndex () - 1;
```
jmark, line 1478:
```suggestion
  return elem_iter.GetCurrentIndex () - 1;
```

src/t8_data/t8_element_array_iterator.hxx
```diff
@@ -0,0 +1,190 @@
...
+    current_index += n;
+    return *this;
+  }
+  t8_element_array_iterator&
+  operator-= (const difference_type n)
+  {
+    current_index -= n;
+    T8_ASSERT (current_index < 0);
+    return *this;
+  }
```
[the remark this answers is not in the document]
jmark: As it is right now, you do not actually need these operations. But switching to `std::random_access_iterator_tag` you do however.

src/t8_data/t8_element_array_iterator.hxx
```diff
@@ -0,0 +1,190 @@
...
+ */
+class t8_element_array_iterator {
+
+ private:
+  const t8_eclass_scheme_c* scheme; /*!< The scheme of the elements residing within the array. */
+  const sc_array_t* elements;       /*!< A pointer to the actual serialized array of element pointers. */
+  t8_locidx_t current_index { 0 };  /*!< The index the iterator currently points to. */
+
+ public:
+  using iterator_category = std::bidirectional_iterator_tag;
```
jmark, line 57: For proper binary search support you need to change to this.
```suggestion
  using iterator_category = std::random_access_iterator_tag;
```

https://en.cppreference.com/w/cpp/algorithm/binary_search states that
```
However, if ForwardIt is not a [LegacyRandomAccessIterator](https://en.cppreference.com/w/cpp/named_req/RandomAccessIterator), the number of iterator increments is linear in N.
```

jmark: Then more operators are needed.

jmark (changes requested):
I got into the mood for code deletions. Let me know what you think about them.

Moreover, it seems that currently `std::upper_bound` falls back to linear search with given iterator.

src/t8_data/t8_element_array_iterator.hxx
```diff
@@ -0,0 +1,190 @@
...
+  using value_type = t8_element_t*;
+  using reference = t8_element_t*&;
+
+  /* Constructors */
+  t8_element_array_iterator () = delete;
+  t8_element_array_iterator (const t8_element_array_t* element_array, const t8_locidx_t position)
+    : scheme { t8_element_array_get_scheme (element_array) }, elements { t8_element_array_get_array (element_array) },
+      current_index { position } {};
+
+  /* Copy/Move Constructors/Assignment-Operators */
```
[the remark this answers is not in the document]
Niklas997: Yes, those would be implicitly declared nevertheless and are not necessarily needed. However, dependig on your class it might be possible that a class is for example not copyable but movable and it might not always be clear from the class's public interface or that some of those operations are implicitly deleted based on the design of the class. By explcitly declaring/defaulting or deleting those copy and move operations in the public section, the user can directly see what operations are permitted with the class. and therefore I would suggest to keep these lines.

src/t8_data/t8_element_array_iterator.hxx
Niklas997: Yes some of the current operators are needed in order to meet the requirements of a LegacyRandomAccessIterator, but apparently I have not implemented all of the needed comparison operators. However, this does not enforce us to switch to a std::random_access_iterator_tag in order to not have a linear complexity in the iterator increments. However, we can try to change the iterator from a bidirectional to random access iterator nonetheless.

jmark: I suggest to do so. Otherwise we would be slower than the old code, right?

Niklas997: Yes, this could be the case. By switching to a std::random_iterator_tag we can be sure that it has a logarithmic complexity in the iterator increments.

src/t8_data/t8_element_array_iterator.hxx
```diff
@@ -0,0 +1,190 @@
...
+  GetArrayIndex () const
+  {
+    return current_index;
+  };
+
+  /* Compute the linear id at a given level for the element the iterator points to. */
+  t8_linearidx_t
+  GetLinearIDAtLevel (const int level)
+  {
+    T8_ASSERT (current_index >= 0 && static_cast<size_t> (current_index) < elements->elem_count);
```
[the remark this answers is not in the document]
Niklas997: Personally, I would like to keep the assertion, since it is only activated in Debug-Mode and it helps to directly identify that the iterator access is "invalid". Otherwise, the assertion would happen deeper in the callback hierachy and it is "harder" to trace back.

src/t8_forest/t8_forest.cxx
Niklas997: Personally, I like to define a const variable beforehand, but we can change this of course.

src/t8_forest/t8_forest.cxx
```diff
@@ -1438,50 +1441,41 @@ t8_forest_copy_trees (t8_forest_t forest, t8_forest_t from, int copy_elements)
...
-    }
-    else {
-      /* look further right, but keep guess in the search range */
-      low = guess;
-    }
-  }
-  T8_ASSERT (low == high);
-  return low;
+  /* A typedef for the value type of the t8_element_array_t iterator. */
+  using element_ptr_t = t8_element_t *;
```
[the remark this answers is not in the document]
jmark: I think you can save this line by simply using `t8_element_array_iterator::value_type`. See comment below.

src/t8_forest/t8_forest.cxx
```diff
@@ -1438,50 +1441,40 @@ t8_forest_copy_trees (t8_forest_t forest, t8_forest_t from, int copy_elements)
...
+
+  /* We search for the first element in the array that is greater than the given element id. */
+  auto elem_iter
+    = std::upper_bound (t8_element_array_begin (elements), t8_element_array_end (elements), element_id,
+                        [&maxlevel, &ts] (const t8_linearidx_t element_id_, const element_ptr_t &elem_ptr) {
```
jmark, line 1468:
```suggestion
                        [&maxlevel, &ts] (const t8_linearidx_t element_id_, const t8_element_array_iterator::value_type &elem_ptr) {
```

jmark (changes requested):

jmark (approved):

PR #1208, opened 2024-08-12
Improve search callback

### LINKED ISSUE #1203: Change the search callback function to better catch usage errors.
# Feature request

The `t8_forest_search_query_fn` (https://github.com/DLR-AMR/t8code/blob/ad7d4d8bd25a8fcc9969d91fa1a8a9776a498461/src/t8_forest/t8_forest_iterate.h#L57C15-L57C40) should be changed in several minor ways to drastically improve user experience:

When switching from the old to the new search interface users are not guided enough. Neither the docu nor the parameter design prevent hard to track conversion errors.

1. The `void *query` argument is actually a `sc_array_t *queries` and should be changed to that. Since the last interface change, if users 
do not change their code it still compiles since they needed to cast the `void *` into a query by hand. However they mistakenly convert an `sc_array_t *` and the code does not work. (happened to me).

1.a. Bonus points for templatizing the search over the query data type to allow C++ users to get away of the casting.

2. The docstring "\param[in] queries             if not NULL, a query that is passed through from the search function" must be adapted to the new interface.

3. The return value is not used in the search function anymore and thus `t8_forest_search_query_fn` should be `void`. This also will avoid user errors when switching to the new interface. The old return value is now handled in "query_matches".

4. The documentation of "query_matches" should state that the user has to set all entries (and thus handle all queries) when the element is not a leaf. When the element is a leaf the function can return before handling all queries to optimize performance.

5. The documentation of "query_indices" should clearly state the data type of the query_indices (size_t)

6. Use a different callback function for `search_fn` than for `query_fn` in the actual search. First these serve different purposes and have in fact different parameters (we currently just set some to NULL or 0). Also the `search_fn` actually has a meaningfull return value while `query_fn` has not.

**Estimated priority**

"Priority: medium"  Should be solved within half a year 

But could also be high for next release.

**Additional context**

Davknapp:
Closes #1203 

Split the search_query_callback function into two parts, one to use for searching, one to use for queries. 
updated the documentation. 

#### General

#### Tests

#### Github action

#### Scripts and Wiki

#### Licence


src/t8_forest/t8_forest_iterate.h
```diff
@@ -44,20 +44,30 @@ typedef int (*t8_forest_iterate_face_fn) (t8_forest_t forest, t8_locidx_t ltreei
  * \param[in] leaf_elements       the leaf elements in \a forest that are descendants of \a element (or the element 
  *                                itself if \a is_leaf is true)
  * \param[in] tree_leaf_index     the local index of the first leaf in \a leaf_elements
```
holke, line 46: When we are at it we should improve this comment.
There is no info at all about that typedef before the parameters.

I suggest writing two lines like "this is the callback function used by search to..." etc.

src/t8_forest/t8_forest_iterate.h
```diff
@@ -44,20 +44,30 @@ typedef int (*t8_forest_iterate_face_fn) (t8_forest_t forest, t8_locidx_t ltreei
...
- * \param[in] queries             if not NULL, a query that is passed through from the search function
```
holke, line 47: The element parameter comment says "the element for which the query is executed" which should be corrected.

src/t8_forest/t8_forest_iterate.h
```diff
@@ -44,20 +44,30 @@ typedef int (*t8_forest_iterate_face_fn) (t8_forest_t forest, t8_locidx_t ltreei
...
- * \param[in] query_indices       if \a query is not NULL the indices of \a query in the \a queries array from \ref 
- *                                t8_forest_search
- * \param[in, out] query_matches  if \a query is not NULL: true at the i-th index if and only if the element 'matches' 
- *                                the query of the i-th query index. 
- *                                if \a query is NULL: true if and only if the search should continue with the
- *                                children of \a element and the queries should be performed for this element.
```
holke, line 53: We should also document the return value of this callback.

src/t8_forest/t8_forest_iterate.h
```diff
@@ -44,20 +44,30 @@ typedef int (*t8_forest_iterate_face_fn) (t8_forest_t forest, t8_locidx_t ltreei
...
+ */
+typedef int (*t8_forest_search_fn) (t8_forest_t forest, const t8_locidx_t ltreeid, const t8_element_t *element,
+                                    const int is_leaf, const t8_element_array_t *leaf_elements,
+                                    const t8_locidx_t tree_leaf_index);
+
+/*
```
holke, line 52: The same as above: a few introductory lines about what this callback does.

holke: Also this should be a doxygen comment which it does not seem to be.
(Same above)

src/t8_forest/t8_forest_iterate.h
```diff
@@ -44,20 +44,30 @@ typedef int (*t8_forest_iterate_face_fn) (t8_forest_t forest, t8_locidx_t ltreei
...
+ * \param[in] forest              the forest
+ * \param[in] ltreeid             the local tree id of the current tree
+ * \param[in] element             the element for which the query is executed
```
holke, line 55:
```suggestion
 * \param[in] element             the element for which the queries are checked
```

src/t8_forest/t8_forest_iterate.h
```diff
@@ -44,20 +44,30 @@ typedef int (*t8_forest_iterate_face_fn) (t8_forest_t forest, t8_locidx_t ltreei
...
+ * \param[in] is_leaf             true if and only if \a element is a leaf element
+ * \param[in] leaf_elements       the leaf elements in \a forest that are descendants of \a element (or the element 
+ *                                itself if \a is_leaf is true)
+ * \param[in] tree_leaf_index     the local index of the first leaf in \a leaf_elements
+ * \param[in] queries             An array of queries that are checked by the function
+ * \param[in] query_indices       An array of size_t entries, where each entry is an index of a query in \q queries.
+ * \param[in, out] query_matches  An array of length \a num_active_queries. True at the i-th index if and only if the element 'matches' 
+ *                                the query of the i-th query index. When the element is a leaf we can return without handling all queries for optimization.
```
holke, line 63:
```suggestion
 * \param[in, out] query_matches  An array of length \a num_active_queries. 
 *                                                  If the element is not a leave must be set to true or false at the i-th index for each query, specifying whether the element 'matches' 
 *                                                  the query of the i-th query index or not. 
 *                                                  When the element is a leaf we can return before all entries are set. 
```

src/t8_forest/t8_forest_iterate.h
```diff
@@ -44,20 +44,30 @@ typedef int (*t8_forest_iterate_face_fn) (t8_forest_t forest, t8_locidx_t ltreei
...
+ * \param[in] element             the element for which the query is executed
+ * \param[in] is_leaf             true if and only if \a element is a leaf element
+ * \param[in] leaf_elements       the leaf elements in \a forest that are descendants of \a element (or the element 
+ *                                itself if \a is_leaf is true)
+ * \param[in] tree_leaf_index     the local index of the first leaf in \a leaf_elements
+ * \param[in] queries             An array of queries that are checked by the function
+ * \param[in] query_indices       An array of size_t entries, where each entry is an index of a query in \q queries.
+ * \param[in, out] query_matches  An array of length \a num_active_queries. True at the i-th index if and only if the element 'matches' 
+ *                                the query of the i-th query index. When the element is a leaf we can return without handling all queries for optimization. 
  * \param[in] num_active_queries  The number of currently active queries. Does not have to be equal to query->elem_count,
```
holke, line 64:
```suggestion
 * \param[in] num_active_queries  The number of currently active queries (equals the number of entries of \a query_matches and entries of \a query_indices). 
 *                                                 Does not have to be equal to queries->elem_count,
```

src/t8_forest/t8_forest_iterate.h
```diff
@@ -44,20 +44,30 @@ typedef int (*t8_forest_iterate_face_fn) (t8_forest_t forest, t8_locidx_t ltreei
...
  *                                since some queries might have been deactivated from previous calls
  */
-typedef int (*t8_forest_search_query_fn) (t8_forest_t forest, const t8_locidx_t ltreeid, const t8_element_t *element,
-                                          const int is_leaf, const t8_element_array_t *leaf_elements,
-                                          const t8_locidx_t tree_leaf_index, void *query, sc_array_t *query_indices,
-                                          int *query_matches, const size_t num_active_queries);
+typedef void (*t8_forest_query_fn) (t8_forest_t forest, const t8_locidx_t ltreeid, const t8_element_t *element,
+                                    const int is_leaf, const t8_element_array_t *leaf_elements,
+                                    const t8_locidx_t tree_leaf_index, sc_array_t *query, sc_array_t *query_indices,
```
holke, line 69:
```suggestion
                                    const t8_locidx_t tree_leaf_index, sc_array_t *queries, sc_array_t *query_indices,
```

tutorials/general/t8_tutorial_search.cxx
```diff
@@ -171,10 +169,10 @@ t8_tutorial_search_callback (t8_forest_t forest, const t8_locidx_t ltreeid, cons
  * a counter for this element by one.
  * These counters are provided in an sc_array as user data of the input forest.
  */
-static int
+static void
 t8_tutorial_search_query_callback (t8_forest_t forest, const t8_locidx_t ltreeid, const t8_element_t *element,
                                    const int is_leaf, const t8_element_array_t *leaf_elements,
-                                   const t8_locidx_t tree_leaf_index, void *query, sc_array_t *query_indices,
+                                   const t8_locidx_t tree_leaf_index, sc_array_t *query, sc_array_t *query_indices,
```
holke, line 175:
```suggestion
                                   const t8_locidx_t tree_leaf_index, sc_array_t *queries, sc_array_t *query_indices,
```

holke: Needs to be changed in the body of the function as well, but github doesnt let me do that here in review mode.

holke (changes requested):
Very good and thanks for adressing it so quickly.

Only minor changes/comments.

I think it is much clearer now and much easier to understand.

src/t8_forest/t8_forest_iterate.h
Davknapp: Thanks for the reminder! I updated the comment.

holke (approved):

PR #1230, opened 2024-08-22
Add option to export cmake compile commands

sandro-elsweijer:
Adds a CMaek option to export the `compile_commands.json` file. This file can be imported by IDEs to fetch the compiler macros (like T8_WITH_VTK) which were in the t8_config.h from autotools. This also exports the include files of external libraries to the IDE, so that docstrings and IntelliSense become available for external libraries.

#### General

#### Tests

#### Github action

#### Scripts and Wiki

#### Licence


src/CMakeLists.txt
```diff
@@ -25,6 +25,10 @@ if( CMAKE_BUILD_TYPE STREQUAL "Debug" )
     target_compile_definitions( T8 PUBLIC T8_ENABLE_DEBUG )
 endif()
 
+if( T8CODE_EXPORT_COMPILE_COMMANDS )
+  target_compile_definitions(T8 PRIVATE CMAKE_EXPORT_COMPILE_COMMANDS=1)
```
maelk3, line 29:
```suggestion
  set_target_properties( T8 PROPERTIES EXPORT_COMPILE_COMMANDS ON )
```
The `target_compile_definitions` did not work on my side, however this seem to work. I would also suggest adding the same property to every test/benchmark as well or advise users in the wiki to globally set the variable `CMAKE_EXPORT_COMPILE_COMMANDS` in the build command (e.g. `cmake path/to/t8code -DCMAKE_EXPORT_COMPILE_COMMANDS=ON`) to export compile commands for all targets automatically.

CMakeLists.txt
```diff
@@ -10,6 +10,7 @@ include( CTest )
 
 option( T8CODE_BUILD_AS_SHARED_LIBRARY "Whether t8code should be built as a shared or a static library" ON )
 option( T8CODE_BUILD_PEDANTIC "Compile t8code with `-Wall -pedantic -Werror` as done in the Github CI." OFF )
+option( T8CODE_EXPORT_COMPILE_COMMANDS "Export the compile commands as json. Can be used in code editors for intellisense" OFF )
```
maelk3, line 13:
```suggestion
option( T8CODE_EXPORT_COMPILE_COMMANDS "Export the compile commands as json. Can be used by IDEs for code completion (e.g. intellisense, clangd)" OFF )
```

maelk3 (changes requested):

src/CMakeLists.txt
sandro-elsweijer: Done :)

maelk3 (approved):

PR #1274, opened 2024-10-18
add placeholder for internal data handling

ole-alb:
Internal data is processed in the background without the user noticing or being restricted.

#### General

#### Tests

#### Github action

#### Scripts and Wiki

#### License

### DIFF
--- a/src/t8_data/t8_data_handler.hxx
+++ b/src/t8_data/t8_data_handler.hxx
@@ -248,6 +248,16 @@ class t8_data_handler: public t8_abstract_data_handler {
     return single_handler.type ();
   }
 
+  bool
+  create_handle_for_internal_data (auto &ihandler, type)
+  {
+    switch (type) {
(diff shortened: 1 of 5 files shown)


src/t8_data/t8_data_handler.hxx
```diff
@@ -248,6 +248,16 @@ class t8_data_handler: public t8_abstract_data_handler {
     return single_handler.type ();
   }
 
+  bool
+  create_handle_for_internal_data (auto &ihandler, type)
```
Davknapp, line 252:
```suggestion
  create_handle_for_internal_data (t8_abstract_data_handler * handler, const int type)
```

Davknapp: Please add this function to the base-class so we can call it from there.

Davknapp (changes requested):

PR #1304, opened 2024-11-15
add assertion with error message

### LINKED ISSUE #1213: Assertion with an optional message
# Feature request

I would like to test for preconditions and postconditions, but only when compiled in debug mode. Printing a message why something fails is useful. On the other hand, `T8_ASSERT`, similarly to the built-in `assert`, only accepts an expression, no optional message.

A convenience macro, e.g. `T8_ASSERTF` would be nice.

To achieve the same effect, we now need to use the more verbose
```cpp
#if T8_ENABLE_DEBUG
   if (!condition)
      SC_ABORTF("Error message");
#endif
```
Alternatively, Boost has `BOOST_ASSERT_MSG`. There are other tricks, such as
```cpp
assert(("Error message", condition));
```
See [StackOverflow](https://stackoverflow.com/questions/3692954/add-custom-messages-in-assert) for more details.

**Estimated priority**

"Priority: low" Should be solved eventually

**Additional context**
This feature request concerns assertions, not logging. For that, functions such as `t8_global_productionf` already exist.

ole-alb:
Add the macro `T8_ASSERTF` to allow outputting an error message alongside the assertion in debug mode.

Closes #1213 

#### General

#### Tests

#### Github action

#### Scripts and Wiki

#### License

### DIFF
--- a/src/t8.h
+++ b/src/t8.h
@@ -73,6 +73,25 @@ T8_EXTERN_C_BEGIN ();
 #define T8_ASSERT(c) SC_NOOP ()
 #endif
 
+/** Extended assertion with detailed error message (including the condition, file, function, 
+ * and line number) along with a custom user-specified message. Only active in debug-mode.
+ * \param cond     The condition to check.
+ * \param msg      A custom message to display if the assertion fails.
+ * \param ...      Optional additional arguments to format the message.
+  */
+#ifdef T8_ENABLE_DEBUG
+#define T8_ASSERTF(cond, msg, ...) \
+  do { \
+    if (!(cond)) { \
+      fprintf (stderr, "Assertion failed: (%s), function %s, file %s, line %d.\nMessage: " msg "\n", #cond, \
+               __FUNCTION__, __FILE__, __LINE__, ##__VA_ARGS__); \
+      abort (); \
+    } \
+  } while (0)
+#else
+#define T8_ASSERTF(cond, msg, ...) SC_NOOP ()
+#endif
+
 /** Allocate a \a t-array with \a n elements. */
 #define T8_ALLOC(t, n) (t *) sc_malloc (t8_get_package_id (), (n) * sizeof (t))
 


src/t8.h
```diff
@@ -73,6 +73,25 @@ T8_EXTERN_C_BEGIN ();
...
+  */
+#ifdef T8_ENABLE_DEBUG
+#define T8_ASSERTF(cond, msg, ...) \
+  do { \
+    if (!(cond)) { \
+      fprintf (stderr, "Assertion failed: (%s), function %s, file %s, line %d.\nMessage: " msg "\n", #cond, \
+               __FUNCTION__, __FILE__, __LINE__, ##__VA_ARGS__); \
+      abort (); \
+    } \
+  } while (0)
```
Davknapp, line 90: I do not like the while(0) approach here. Why do we need it? Can we remove it?

Davknapp (changes requested):

src/t8.h
ole-alb: As I understand it, this ensures that the macro always behaves as intended, for example, when used with external `if` statements. I got the explanation from [here](https://hownot2code.wordpress.com/2016/12/05/do-while-0-in-macros/).

Davknapp: What about a t8code wrapper around `SC_CHECK_ABORT` ? It will provide the desired feature and also take care that t8code stops on all processes, not only the one that just failed.

ole-alb: You are correct. The `SC_CHECK_ABORT`, which is currently used in `T8_ASSERT`, already has the capability to display an error message. Therefore, I have now added the option to include the error message directly in `T8_ASSERT`. As a result, there won't be an additional `T8_ASSERTF`, instead, you can now add the message directly in `T8_ASSERT`.

ole-alb: Never mind, I added `T8_ASSERTF`, which displays a custom error message. `T8_ASSERT` itself stays untouched.

PR #1326, opened 2024-12-13
Configure with -O1 on debug-builds

Davknapp:
All Debug builds should be configured with -O1 to increase runtime. 

#### General

#### Tests

#### Github action

#### Scripts and Wiki

#### License

#### Tag Label


.github/workflows/build_cmake_tarball.yml
```diff
@@ -79,6 +79,9 @@ jobs:
       if: ${{ inputs.LESS_TESTS }}
       run: export LESS_TEST_OPTION="-DT8CODE_ENABLE_LESS_TESTS=ON"
           && echo LESS_TEST_OPTION="$LESS_TEST_OPTION" >> $GITHUB_ENV
+    - name: compiler-optimization
+      if: ${{ inputs.BUILD_TYPE == 'Debug' }}
+      run: export -DCMAKE_C_FLAGS_DEBUG="-O1 -g" -DCMAKE_CXX_FLAGS_DEBUG="-O1 -g"
```
sandro-elsweijer, line 84: Do you maybe want to define a new input which gets passed to `CONFIG_OPTIONS`? Then we can control this in the central CI script. Something like

Input: addtitional_config_options

.github/workflows/build_cmake_tarball.yml
```diff
@@ -79,6 +79,9 @@ jobs:
...
     - name: build config variables
       run: export CONFIG_OPTIONS="${LESS_TEST_OPTION} -GNinja -DT8CODE_USE_SYSTEM_SC=OFF -DT8CODE_USE_SYSTEM_P4EST=OFF -DT8CODE_BUILD_PEDANTIC=ON -DT8CODE_BUILD_WALL=ON -DT8CODE_BUILD_WERROR=ON -DT8CODE_ENABLE_MPI=$MPI -DCMAKE_BUILD_TYPE=$BUILD_TYPE"
```
sandro-elsweijer, line 86: just pseudocode
```suggestion
      run: export CONFIG_OPTIONS="${LESS_TEST_OPTION} -GNinja -DT8CODE_USE_SYSTEM_SC=OFF -DT8CODE_USE_SYSTEM_P4EST=OFF -DT8CODE_BUILD_PEDANTIC=ON -DT8CODE_BUILD_WALL=ON -DT8CODE_BUILD_WERROR=ON -DT8CODE_ENABLE_MPI=$MPI -DCMAKE_BUILD_TYPE=$BUILD_TYPE ${additional_config_options}"
```

.github/workflows/build_cmake_tarball.yml
Davknapp: No, because these would be additional options. I want to change the options used when the Debug mode is used, which are written in CMAKE_C/CXX_FLAGS_DEBUG. Additional config options would just append to that.

.github/workflows/build_cmake_tarball.yml
Davknapp: Ahhh, now I understand what you mean. Yes, that would be an option, but currently we want that in our CI we always run with -O1 in Debug mode. So it is not an option, it is a requirement.

.github/workflows/build_cmake_tarball.yml
```diff
@@ -25,20 +25,11 @@ name: CMake Tarball
...
-    branches:
-      - main
-      - develop
-      - CI-*tarball* # for testing this script, all feature branches with "tarball" in their name
-  pull_request:
-    branches:
-      - main
-      - develop
-  workflow_dispatch: 
+  workflow_call:
```
sandro-elsweijer, line 32: Can you add an input for the less test option?

sandro-elsweijer (approved):

PR #1339, opened 2025-01-15
Fix dquad/pquad confusion bug

holke:

The default quad scheme uses the `t8_pquad_t` data type (internally mapped to `p4est_quadrant_t`).
One header defined the data type `t8_dquad_t` ("d" instead of "p") completely differently and used it in the function `t8_dquad_compute_reference_coords` where it was wrongly casted into a `t8_pquad_t`.

This is a serious data conversion bug that may have caused issue in the past.

What we did to resolve it:

- Remove `t8_dquad_t` since it was not used elsewhere
- Use `t8_pquad_t` instead
- Rename the files `src/t8_schemes/t8_default/t8_default_quad/t8_dquad_bits.cxx` to `src/t8_schemes/t8_default/t8_default_quad/t8_default_quad_bits.cxx` to avoid name confusion in the future.

#### General

#### Tests

#### Github action

#### Scripts and Wiki

#### License

#### Tag Label

- [x] The author added the patch/minor/major label in accordance to semantic versioning.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -147,7 +147,7 @@ target_sources( T8 PRIVATE
     t8_schemes/t8_default/t8_default_pyramid/t8_dpyramid_bits.c
     t8_schemes/t8_default/t8_default_pyramid/t8_dpyramid_connectivity.c 
     t8_schemes/t8_default/t8_default_quad/t8_default_quad.cxx
-    t8_schemes/t8_default/t8_default_quad/t8_dquad_bits.c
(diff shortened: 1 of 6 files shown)


lukasdreyer:
> This is a serious data conversion bug that may have caused issue in the past.

I agree that this is a useful and necessary PR, but it only caused confusion for the developers and not for the compiler, as the `dquad_t` was only used to pass pointers, it was not dereferenced.

But I already had that confusion before, so I am glad about the cleanup PR.

PR #1414, opened 2025-02-18
Fix various warnings

dutkalex:
This PR fixes many compiler warnings (at least those from clang) which appear when turning `-Wall -Wextra -Wpedantic` on. Indeed, when trying to increase the warnings level in downstream projects beyond just `-Wall`, it becomes really difficult to navigate the compiler output due to the numerous warnings inherited from t8code's sources. I only fixed those which were obviously false positives, and for which a trivial fix was available. In other words, the handful that remain might be worth looking into because I could not convince myself that they were not bugs in disguise.

#### General

- [x] The code follows the [t8code coding guidelines](https://github.com/DLR-AMR/t8code/wiki/Coding-Guideline)

#### Tests

#### Github action

#### Scripts and Wiki

#### License

#### Tag Label


src/t8_schemes/t8_default/t8_default_hex/t8_default_hex.hxx
```diff
@@ -34,7 +34,7 @@
 #include <t8_schemes/t8_default/t8_default_common/t8_default_common.hxx>
 
 /* Forward declaration of the scheme so we can use it as an argument in the eclass schemes function. */
-class t8_scheme;
+struct t8_scheme;
```
lukasdreyer, line 37: I am not sure what implications these `class` to `struct` changes have.

src/t8_schemes/t8_scheme.h
```diff
@@ -35,7 +35,7 @@
  *  Opaque pointer for C interface.
  *  Detailed documentation at \ref t8_scheme.
  */
-typedef struct t8_scheme t8_scheme_c;
+typedef class t8_scheme t8_scheme_c;
```
lukasdreyer: Also not sure about the other direction here

src/t8_schemes/t8_default/t8_default_hex/t8_default_hex.hxx
dutkalex: @lukasdreyer the warning raised was related to the mismatch between the forward declaration and the implementation. I assumed this was ok and I was a little bit surprised by the warning being raised. From what I could find, this is ok with gcc and clang, but not with msvc (see https://clang.llvm.org/docs/DiagnosticsReference.html#wmismatched-tags for details). Since the fix was trivial I went for it.

src/t8_schemes/t8_scheme.h
dutkalex: This was a mistake because this is a C header and the keyword class can't be used here. It has been reverted if I'm not mistaken.

src/t8_schemes/t8_default/t8_default_hex/t8_default_hex.hxx
lukasdreyer: So you first tried to change the forward declaration in the c file, but this wasn't allowed, so you changed the class to a struct and all related forward declarations?

I think changing the `class` to a `struct` here is a big enough change, that we should discuss it in a separate PR/issue. I want to have input of other developers here as well.

I am happy with the rest of the changes in this PR. Could you revert the class/struct changes so we can discuss them separately?

dutkalex: > So you first tried to change the forward declaration in the c file, but this wasn't allowed, so you changed the class to a struct and all related forward declarations?

Yes exactly. 

> I think changing the class to a struct here is a big enough change, that we should discuss it in a separate PR/issue. I want to have input of other developers here as well.

If I'm not mistaken, there was no member or method in these classes relying on the private default visibility, and the change therefore did not weaken encapsulation. However, I strongly believe is the "we write code for humans to read first" motto and I completely get that such a change has implications beyond just what the compiler has to say about it, and that you want to discuss it with the team. Feel free to tag me if I can help in some way or another regarding this matter.

> I am happy with the rest of the changes in this PR. Could you revert the class/struct changes so we can discuss them separately?

Yes this should be easy to revert.

lukasdreyer (approved):

Issue #1487, opened 2025-03-20
Enable Ghost value output for VTK


holke:
# Feature request

Currently if VTK output is called with the "write_ghosts" flag, then still "0" is written for any user data.
Instead, if "write_ghosts" is set and user data is provided, then the values at the ghost element indices should be written our correctly.

**Is your feature request related to a problem? Please describe.**
It is not properly documented that we do not write the ghost data which leads to misunderstanding of the function and wrong assumptions on the actual data present.

**Describe the solution or feature you'd like**
If "write_ghosts" is set in `t8_forest_write_vtk_ext` then any provided data must contain entries for ghost values and these should be written into the vtu file.

**Describe alternatives you've considered**
Alternatively at leas a "nan" or "inf" value should be written instead of "0" in order to not confuse it with real data.

**Estimated priority**
Which of these is most applicable (remove the others):

"Priority: low" Should be solved eventually

**Additional context**
This came up during a debugging session with an applications and the users did not know that we do not write ghost values and were confused about 0 being written. They thought their code was not setting the ghost values correctly.

holke:
#1486 is an ASCII only untested variant of this that solves part of the problem.

PR #1488, opened 2025-03-20
Feature: Add a scalar and vector data field to VTK write test

holke:

The first step towards #1487.
The vtk_writer example did not write any user data.
Extended the example with 2 data fields, one scalar, one vector and dummy data.
The data fields do already contain values for the ghost elements.

#### General
#### Tests

#### Scripts and Wiki
#### Tag Label
#### License

### DIFF
--- a/test/t8_IO/t8_gtest_vtk_writer.cxx
+++ b/test/t8_IO/t8_gtest_vtk_writer.cxx
@@ -113,6 +113,24 @@ use_c_interface<t8_cmesh_t> (const t8_cmesh_t grid, const char *fileprefix, cons
 #endif
 }
 
+void
+vtk_writer_test_fill_data (const t8_locidx_t cells_to_write_count, std::vector<double> &scalar_data,
+                           std::vector<double> &vector_data)
+{
+  // Fill scalar data vector with entries 0,1/10,...,(N-1)/10
+  //    vector[n] = (n/10.)
+  // Fill vector data vector with entries (0, 0, 42), (0.1,-0.1,42), ...
+  //    vector[n] = (n/10.,-n/10., 42)
+  for (t8_locidx_t icell = 0; icell < cells_to_write_count; ++icell) {
+    const double scalar_value = icell / 10.;
+    const double vector_values[3] = { scalar_value, -scalar_value, 42. };
+    scalar_data.push_back (icell);
+    vector_data.push_back (vector_values[0]);
+    vector_data.push_back (vector_values[1]);
+    vector_data.push_back (vector_values[2]);
+  }
+}
+
 /**
  * Templated class to test the vtk writer for forests and cmeshes. 
  * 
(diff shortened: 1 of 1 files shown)


holke:
Note: even though writing of ghosts was set true (and not changed by this commit), the ghost elements do not seem to be written, which is odd.

holke:
> Note: even though writing of ghosts was set true (and not changed by this commit), the ghost elements do not seem to be written, which is odd.

Could be resolved, the forest in the example was build without a ghost layer.

holke:
The tests fail in VTK API mode due to a ghost tree being requested by grid_element_shape and the function expecting only local trees.
@Davknapp is working on a solution.

test/t8_IO/t8_gtest_vtk_writer.cxx
```diff
@@ -114,6 +116,24 @@ use_c_interface<t8_cmesh_t> (const t8_cmesh_t grid, const char *fileprefix, [[ma
 #endif
 }
 
+void
+vtk_writer_test_fill_data (const t8_locidx_t cells_to_write_count, std::vector<double> &scalar_data,
+                           std::vector<double> &vector_data)
+{
```
Davknapp, line 122:
```suggestion
{
  scalar_data.resize (cells_to_write_count);
  vector_data.resize (3 * cells_to_write_count);
```

I think it is better to use a single allocation instead of `cells_to_write_count` many

test/t8_IO/t8_gtest_vtk_writer.cxx
```diff
@@ -114,6 +116,24 @@ use_c_interface<t8_cmesh_t> (const t8_cmesh_t grid, const char *fileprefix, [[ma
...
+  // Fill scalar data vector with entries 0,1/10,...,(N-1)/10
+  //    vector[n] = (n/10.)
+  // Fill vector data vector with entries (0, 0, 42), (0.1,-0.1,42), ...
+  //    vector[n] = (n/10.,-n/10., 42)
+  for (t8_locidx_t icell = 0; icell < cells_to_write_count; ++icell) {
```
Davknapp, line 127:
```suggestion
 std::iota(scalar_data.begin(), scalar_data.end(), 0);
  std::transform(scalar_data.begin(), scalar_data.end(), scalar_data.begin(), [](double val) { return val / 10.; });

  std::generate(vector_data.begin(), vector_data.end(), [n = 0]() mutable {
    double scalar_value = (n / 10.);
    double vector_values[3] = { scalar_value, -scalar_value, 42. };
    return vector_values[n++ % 3];
  });
```

we could replace the loop by stl commands then. It has the drawback of looping three times over the arrray, though, so I would be okay with just removing the allocation from the loop too.

Davknapp (changes requested):

test/t8_IO/t8_gtest_vtk_writer.cxx
holke: I do not think it makes a real difference due to the intelligent handling of memory of std::vector (i.e. not allocation one item at a time).

test/t8_IO/t8_gtest_vtk_writer.cxx
holke: Yes good point. I used the generator now.

holke:
This currently fails due ghost trees not being considered in 
```c++
grid_element_shape<t8_forest_t> (const t8_forest_t grid, const t8_locidx_t itree, const t8_element_t *element)
{
  const t8_eclass_t eclass = t8_forest_get_eclass (grid, itree);
```
Fix is easy, extend the function to accept ghost tree ids and call "get_ghost_tree_class" if so.
Must still decide whether "ist_ghost" should be added as parameter or will be computed by the function.

src/t8_vtk/t8_vtk_write_ASCII.cxx
```diff
@@ -327,25 +327,40 @@ t8_forest_vtk_cells_elementid_kernel (t8_forest_t forest, [[maybe_unused]] const
   return 1;
 }
 
+// TODO: Move to forest_general.h
+static t8_locidx_t
```
Davknapp, line 331: Please document the parameters of the function.

holke: Done

Davknapp (dismissed):

holke:
@Davknapp 
The tests pass now. I did not add any code, just merged with main.

Can be merged immediately imho.

PR #1516, opened 2025-03-27
Bugfix: CMake Install paths

### LINKED ISSUE #1628: Bugfix: CMake Install paths

Davknapp:
Closes #1628

Instead of hard-coding paths use GNUInstallDirs to automatically set installation dirs for bin, lib etc..

We hard coded some installation paths, while p4est and sc use GNUInstallDirs to automatically set these paths. This PR unifies the behaviour

#### General
#### Tests

#### Scripts and Wiki
#### License


CMakeLists.txt
```diff
@@ -127,8 +128,11 @@ set( BUILD_SHARED_LIBS ON CACHE BOOL "Build libsc as a shared library" )
 set(CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT FALSE)
 
 # Rpath options necessary for shared library install to work correctly in user projects.
-set(CMAKE_INSTALL_NAME_DIR ${CMAKE_INSTALL_PREFIX}/lib)
-set(CMAKE_INSTALL_RPATH ${CMAKE_INSTALL_PREFIX}/lib)
+file(RELATIVE_PATH relativeRpath
+    ${CMAKE_CURRENT_BINARY_DIR}/${CMAKE_INSTALL_BINDIR}
+    ${CMAKE_CURRENT_BINARY_DIR}/${CMAKE_INSTALL_LIBDIR}
+)
```
spenke91, line 134: Your solution may be completely right, but better safe than sorry:
According to CMake documentation, the `RELATIVE_PATH` argument of `file` works like this:

image

It seems to odd to me that we use two directories here instead; and wouldn't this mean `relativePath` would contain the relative path from the `bin` to the `lib`folder, so basically `.../lib`? 🤔 

Or is this supposed to be some correction of the current state only?

example/CMakeLists.txt
```diff
@@ -36,7 +36,7 @@ function( add_t8_example )
         set_target_properties( ${ADD_T8_EXAMPLE_NAME} PROPERTIES EXPORT_COMPILE_COMMANDS ON )
     endif( T8CODE_EXPORT_COMPILE_COMMANDS )
 
-    install( TARGETS ${ADD_T8_EXAMPLE_NAME} DESTINATION bin )
+    install( TARGETS ${ADD_T8_EXAMPLE_NAME} DESTINATION ${CMAKE_INSTALL_BINDIR} )
```
spenke91, line 39: Should this be `install( TARGETS ${ADD_T8_EXAMPLE_NAME} RUNTIME )` , like in the tutorials below?

spenke91 (commented):
I tested it on the Teamsserver and it worked fine - however, it also did so before, so maybe that's not a great test. As you can see, I do have some question about the relative path 🤔

spenke91:
Also, the PR description says it unifies the installation dirs with p4est and sc, but p4est does not seem to use `GNUInstallDirs` - do you think we might run into problems with that again at some point?

CMakeLists.txt
Davknapp: yes, basically it is ../lib, but it does not require any explicit implementation of the folder, hopefully preventing future errors.

Davknapp: One could use cmake_path instead, but this would require an upgrade to CMake 3.20. 
According to the documentation of Cmake the current implementation should work too:
"This command is dedicated to file and path manipulation requiring access to the filesystem."

Davknapp: Ah nevermind, it also works with 3.16

example/CMakeLists.txt
Davknapp: yes

Davknapp:
> Also, the PR description says it unifies the installation dirs with p4est and sc, but p4est does not seem to use `GNUInstallDirs` - do you think we might run into problems with that again at some point?

I don't know, but maybe.

spenke91 (approved):
Thank you!

PR #1468, opened 2025-03-11
Improvement: [1/2] Rename macros T8_WITH_ to T8_ENABLE_

lenaploetzke:
**_Describe your changes here:_**
Rename macros with T8_WITH_... to T8_ENABLE_... for consistency with the related options (T8CODE_ENABLE...) and other macros. 
This way, incorrect usage of #if and #ifdef is also checked in check_macros.sh. 
Related to #1370.

**_All these boxes must be checked by the AUTHOR before requesting review:_**
- [x] The PR is *small enough* to be reviewed easily. If not, consider splitting up the changes in multiple PRs.
- [x] The title starts with one of the following prefixes: `Documentation:`, `Bugfix:`, `Feature:`, `Improvement:` or `Other:`.
- [x] If the PR is related to an issue, make sure to link it.
- [x] The author made sure that, as a reviewer, he/she would check all boxes below.

**_All these boxes must be checked by the REVIEWERS before merging the pull request:_**

As a reviewer please read through all the code lines and make sure that the code is *fully understood, bug free, well-documented and well-structured*.
#### General
- [ ] The reviewer executed the new code features at least once and checked the results manually.
- [ ] The code follows the [t8code coding guidelines](https://github.com/DLR-AMR/t8code/wiki/Coding-Guideline).
- [ ] New source/header files are properly added to the CMake files.
- [ ] The code is well documented. In particular, all function declarations, structs/classes and their members have a proper doxygen documentation.
- [ ] All new algorithms and data structures are sufficiently optimal in terms of memory and runtime (If this should be merged, but there is still potential for optimization, create a new issue).
#### Tests
- [ ] The code is covered in an existing or new test case using Google Test.

If the Pull request introduces code that is not covered by the github action (for example coupling with a new library):
  - [ ] Should this use case be added to the github action?
  - [ ] If not, does the specific use case compile and all tests pass (check manually).
#### Scripts and Wiki
- [ ] If a new directory with source files is added, it must be covered by the `script/find_all_source_files.scp` to check the indentation of these files.
- [ ] If this PR introduces a new feature, it must be covered in an example or tutorial and a Wiki article.
#### Tag Label
- [ ] The author added the patch/minor/major label in accordance to semantic versioning.
#### License
- [ ] The author added a BSD statement to `doc/` (or already has one).

PR #1522, opened 2025-03-28
T8_WITH_ macros that will throw an error when used

holke:

Since #1468 all our macros have been renamed from T8_WITH* to T8_ENABLE*.
We have an internal check_macros script to find all wrongly used T8_WITH in our code base.
However, user code may not be aware of the changes and still use code like
```C++
#if T8_WITH_MPI
// something
#endif
```
This code will still compile after the update and if users do not read and implement the update news, they will run into behaviour that is
a) undesired and
b) very hard to track down

I propose a solution that overwrites the old T8_WITH_ macros so that using them with "#if"  will throw a compile time error.
Thus, users will be immediately warned and informed to use T8_ENABLE instead.

I put this up for discussion.

The following points are still open:
- Using the macros with #ifdef still does not produce an error. #ifdef just checks whether or not the macro is defined and does not care about the actual definition. Would there even be a solution to throw an error of #ifdef?
- clang-format throws the error "Configuration file(s) do(es) not support Objective-C: /localdata1/holk_jo/coding/source/t8code/1370-warn_users_if_WITH_macro_is_used/.clang-format" 
 Even when using "// clang-format off" comments. Hence, we must somehow make the automatic indent script ignore this file.
- I tried several approaches, including the #error directive, but all of them did not work. So i had to use a slightly hacky solution.

#### General
#### Tests

#### Scripts and Wiki
#### License

### DIFF
--- a/.typos.toml
+++ b/.typos.toml
@@ -3,4 +3,4 @@ eles = "eles"
 packageid = "packageid"
 
 [files]
-extend-exclude = ["scripts/t8indent.sh", "thirdparty/", "t8code_logo.png", "cmake/FindOpenCASCADE.cmake"]
+extend-exclude = ["scripts/t8indent.sh", "thirdparty/", "t8code_logo.png", "cmake/FindOpenCASCADE.cmake", "src/t8_with_macro_error.h"]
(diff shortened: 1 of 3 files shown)


Davknapp:
Please fix the conflicts.

holke:
All merge conflicts resolved - giving this back into review. I expect the tests to pass now. If not, i shall return.

PR #1524, opened 2025-03-28
Bugfix: remove constexpr qualifier

dutkalex:
This PR removes a `constexpr` qualifier which causes build breaks with clang 17. 
There are two issues here:
- The primary function template is first declared with `constexpr` qualification, but then specializations are later implemented without
- The specialization for quad elements uses C-style casts which are not allowed in `constexpr` code

#### General
#### Tests

#### Scripts and Wiki
#### License


dutkalex:
It seems like GCC is actually quite bad at catching this kind of problems (`constexpr` and `virtual`/`override` inconsistencies in particular), as I have experienced this firsthand multiple times lately (and heard others have too). Clang tends to be stricter with respect to language rules generally speaking, and maybe adding a Clang build into the CI should be considered.

ole-alb:
Hey @dutkalex we recently introduced a merge queue in this repo. With that you don't have to merge the current main into your branch, to be able to merge. This comes with the one condition that there are no merge conflicts.

src/t8_schemes/t8_standalone/t8_standalone_implementation.hxx
```diff
@@ -1011,7 +1011,7 @@ struct t8_standalone_scheme
    *                        defined in relation to the smaller face.
    * \note \a elem1 and \a elem2 may point to the same element.
    */
-  static constexpr void
+  static void
```
ole-alb, line 1014:
```suggestion
  static inline void
```

ole-alb (commented):
In general a very good catch, thank you!
I think we could also inline the function here.

dutkalex:
> I think we could also inline the function here.

Looks good to me. Looks good to my local clang17 too :)

ole-alb (approved):

PR #1591, opened 2025-04-15
Feature: vector split

### LINKED ISSUE #1592: Feature: vector split

Davknapp:
Closes #1592

Reimplements sc_array_split for vectors. 

#### General
#### Tests

#### Scripts and Wiki
#### License


spenke91 (approved):
Works like a charm, thanks :-) 

I only added a few (optional) inline comments as suggestions

src/CMakeLists.txt
```diff
@@ -229,6 +229,8 @@ install( FILES
     TYPE INCLUDE
 )
 
+install ( DIRECTORY t8_vector_helper DESTINATION PRIVATE_HEADER FILES_MATCHING PATTERN "*.hxx" )
```
spenke91, line 232: Thanks for adjusting this. I just wanna bring the follwoing to your attention, but feel free to proceed anyway if you feel like it is not an issue:

I think so far we set headers private in t8code by excluding them from the header files, see for example below, lines `234ff`, in which all header files in the directory `t8_cmesh` are included (public) except those explicitly stated not to (private). 

Admittedly, I am not entirely sure which option is the recommended version. 
For compiling, we typically don't have to add headers because they will be included by the source files anyway. In this case, however, I do see the advantage of the current version that `t8_vector_helper.hxx` will be included and compiled even if we compile t8code without the gtests...

spenke91 (approved):
Thanks for the update 👍

src/CMakeLists.txt
Davknapp: I think private headers should be moved to, especially those headers providing the implementation of the function. 
In this example the function is templated and code is generated by the compiler. If we do not provide this header in the installed files we might run into the problem that template arguments provided by applications that are passed down to this function cause an error, because the header can not be found and thus the code can not be generated. Therefore we should do it this way in my opinion.

spenke91: Thanks for the explanation, sounds reasonable

PR #1651, opened 2025-05-15
Improvement - Performance of the current VTK writer in t8code

Flatken:
The current version of the t8_vtk_writer uses the native classes for VTK cell types to add cells. Instantiating these classes do a lot of memory allocations not required here. We can directly use vtkIdType arrays for storing the point ids for a cell and just insert these to the vtkCellArray. This avoids allocation overheads using the native types. Additionally, we 
can already allocate memory for the points, cells, and fields earlier and avoid dynamic resizing of these arrays.

Additionally, when inserting points to the point array, duplications can be avoided by using the vtkMergePoints data structure.

Performance measurements:

| Elements (Hexa)  | Points     | Without Opt. | With Alloc  | With vtkIdType Arrays  | Speedup | 
|------------------|------------|--------------|-------------| -----------------------|---------| 
| 1.519.449        | 12.155.592 | 17646.3ms    | 14352.3ms   | 1045.2ms               | 16.8x   |  

Another feature is to use the vtkMergePoints data structure in order to avoid writing out duplicated points for the shared vertices of elements. 

| Elements (Hexa)  | Points Before | With vtkMergePoints | Runtime | 
|------------------|---------------|---------------------| ------- | 
| 1.519.449        | 12.155.592    | 2.423.685           | 3742.4ms|

#### General
#### Tests

#### Scripts and Wiki
#### License


src/t8_vtk/t8_vtk_writer.hxx
```diff
@@ -256,28 +243,28 @@ class vtk_writer {
     if (curved_flag == 0) {
       switch (element_shape) {
       case T8_ECLASS_VERTEX:
-        pvtkCell = vtkSmartPointer<vtkVertex>::New ();
```
Davknapp, line 259: If i see it correctly, we could get rid of the complete switch-Block here and replace it with a single
`vecCellIds = new vtkIdType[num_node]`

src/t8_vtk/t8_vtk_writer.hxx
```diff
@@ -430,6 +423,16 @@ class vtk_writer {
     int *cellTypes = T8_ALLOC (int, num_cells);
     T8_ASSERT (cellTypes != NULL);
 
+    /* Allocate VTK Memory for the arrays */
+    int iMaxCellSize = 20;
```
Davknapp, line 427:
```suggestion
    const int iMaxCellSize = 20;
```
where does the 20 come from?

Davknapp (changes requested):
Thank you for your contributions and this massive speed-up!

src/t8_vtk/t8_vtk_writer.hxx
Flatken: Right

src/t8_vtk/t8_vtk_writer.hxx
Flatken: I just did it. Do you have a correct number when thinking about non linear cell types? 
But it should not be an issue. Its just big enough to avoid dynamic reallocation.

Davknapp:
We need to compute the local bounds of a cmesh to pass the bounds to vtkMergePoints via something like:
```
    points->InitPointInsertion (points_store, unstructuredGrid->GetBounds ());
```

Davknapp:
> We need to compute the local bounds of a cmesh to pass the bounds to vtkMergePoints via something like:
> 
> ```
>     points->InitPointInsertion (points_store, unstructuredGrid->GetBounds ());
> ```

@Flatken you can try to merge with #1659 to use a first version of this computation. This is not reviewed yet and might change, though. The interface will probably stay the same.

src/t8_vtk/t8_vtk_writer.hxx
```diff
@@ -230,105 +216,57 @@ class vtk_writer {
...
     for (int ivertex = 0; ivertex < num_node; ivertex++, (*point_id)++) {
       const size_t offset_3d = 3 * ivertex;
+
       /* Insert the point in the points array. */
-      points->InsertNextPoint (coordinates[offset_3d], coordinates[offset_3d + 1], coordinates[offset_3d + 2]);
-      pvtkCell->GetPointIds ()->SetId (ivertex, *point_id);
+      double vtkCoords[3] = { coordinates[offset_3d], coordinates[offset_3d + 1], coordinates[offset_3d + 2] };
+      if (mergePoints) {
+        points->InsertUniquePoint (vtkCoords, ptId);
+      }
```
Flatken, line 258: I added the possibility to allow duplicated point insertion. The default behaviour is to merge duplicated points

src/CMakeLists.txt
```diff
@@ -271,7 +272,7 @@ install( DIRECTORY t8_forest TYPE INCLUDE FILES_MATCHING PATTERN "*.h"
   PATTERN "t8_forest/t8_forest_balance.h" EXCLUDE
   PATTERN "t8_forest/t8_forest_ghost.h "  EXCLUDE
   PATTERN "t8_forest/t8_forest_private.h" EXCLUDE
-  PATTERN "t8_forest/t8_forest_types.h"   EXCLUDE
+  #PATTERN "t8_forest/t8_forest_types.h"   EXCLUDE
```
Davknapp:
```suggestion
  PATTERN "t8_forest/t8_forest_types.h"   EXCLUDE
```

PR #1656, opened 2025-05-16
Improvement: Better handling of hashes and errors in t8_geometry

holke:

I stumbled on this while testing something else and noticed that if a tree has no geometry, but we query for it (for example when trying to print VTK), we run into a segfault.
While working on this i noticed that the handling of hash values of geometries should be improved.

- Added new header that defines hash handling functions.
- New data type for hashes (a typedef of *size_t*)
- Definition of empty hash, that is a hash for the string "" which we do not allow as name of a geometry and use as hash for a non-existing geometry
- proper error handling when no geometry is registered but asked for. In that case the functions return *nullptr*
- Added a test with a cmesh with no geometry to check that *nullptr* is returned.

#### General
#### Tests

#### Scripts and Wiki
#### License

### DIFF
--- a/src/t8_cmesh/t8_cmesh_geometry.cxx
+++ b/src/t8_cmesh/t8_cmesh_geometry.cxx
@@ -47,9 +47,9 @@ t8_cmesh_set_tree_geometry (t8_cmesh_t cmesh, const t8_gloidx_t gtreeid, const t
 {
   T8_ASSERT (t8_cmesh_is_initialized (cmesh));
   /* Add the hash of the geometry as an attribute to the tree. */
-  size_t hash = geom->t8_geom_get_hash ();
-  t8_cmesh_set_attribute (cmesh, gtreeid, t8_get_package_id (), T8_CMESH_GEOMETRY_ATTRIBUTE_KEY, &hash, sizeof (size_t),
-                          0);
+  t8_geometry_hash_t hash = geom->t8_geom_get_hash ();
+  t8_cmesh_set_attribute (cmesh, gtreeid, t8_get_package_id (), T8_CMESH_GEOMETRY_ATTRIBUTE_KEY, &hash,
+                          sizeof (t8_geometry_hash_t), 0);
 }
 
 const t8_geometry_c *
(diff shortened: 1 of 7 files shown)


src/t8_geometry/t8_geometry_hash.hxx
```diff
@@ -0,0 +1,79 @@
...
+  t8_geometry_hash_t hash (std::hash<std::string> {}(name));
+  return hash;
+}
+
+/**
+ * Query wheyher a given hash value corresponds to en empty string and hence
+ * a non-existing geometry.
+ * 
+ * \param [in] hash A hash value of a geometry.
+ * \return true If \a hash corresponds to e non-existing geometry,i.e. if \a hash == \a t8_geometry_empty_hash.
```
Davknapp, line 68:
```suggestion
 * \return true If \a hash corresponds to a non-existing geometry,i.e. if \a hash == \a t8_geometry_empty_hash.
```

src/t8_cmesh/t8_cmesh_geometry.cxx
```diff
@@ -47,9 +47,9 @@ t8_cmesh_set_tree_geometry (t8_cmesh_t cmesh, const t8_gloidx_t gtreeid, const t
 {
   T8_ASSERT (t8_cmesh_is_initialized (cmesh));
   /* Add the hash of the geometry as an attribute to the tree. */
-  size_t hash = geom->t8_geom_get_hash ();
-  t8_cmesh_set_attribute (cmesh, gtreeid, t8_get_package_id (), T8_CMESH_GEOMETRY_ATTRIBUTE_KEY, &hash, sizeof (size_t),
-                          0);
+  t8_geometry_hash_t hash = geom->t8_geom_get_hash ();
```
sandro-elsweijer, line 50: We decided to not name our types using `_t` since it was not really clear if it is a typedef or type and it is also reserved for POSIX sytsems

sandro-elsweijer (changes requested):

src/t8_cmesh/t8_cmesh_geometry.cxx
holke: Thanks for catching this.

I added this section to our dev guidelines: https://github.com/DLR-AMR/t8code/wiki/Coding-Guideline#data-types-and-classes

sandro-elsweijer (approved):

PR #1705, opened 2025-06-04
Feature: enable logs redirection

dutkalex:
This PR adds the possibility to register an external log function which replaces the default logging backend. The purpose of this feature is to enable downstream projects to incorporate t8code's logs into their own logging system as they see fit.

The only API change is the addition of the `t8_set_external_log_fcn` which activates the logs redirection. It can be called at any time (even before `t8_init`), multiple times at different points during the program execution if necessary, and calling it with `nullptr` as an argument restores the default behavior. This is a non-breaking change, since the behavior is unchanged if the function in never called.

#### General
#### Tests

#### Scripts and Wiki
#### License

### DIFF
--- a/.github/workflows/build_cmake_tarball.yml
+++ b/.github/workflows/build_cmake_tarball.yml
@@ -34,8 +34,9 @@ on:
     inputs:
       TEST_LEVEL:
         required: true
-        type: number
-        description: 'Test level used for configuring (0 for full tests, 1 for less thorough tests, 2 for minimal tests)'
+        type: string
+        description: 'Test level used for configuring (T8_TEST_LEVEL_FULL, T8_TEST_LEVEL_MEDIUM, or T8_TEST_LEVEL_BASIC)'
+        default: 'T8_TEST_LEVEL_FULL'
 
 jobs:
   build:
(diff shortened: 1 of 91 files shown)


src/t8.c
```diff
@@ -34,7 +35,20 @@ t8_get_package_id (void)
 void
 t8_logv (int category, int priority, const char *fmt, va_list ap)
 {
-  sc_logv ("unknown", -1, t8_package_id, category, priority, fmt, ap);
+  char buffer[BUFSIZ];
+
+#ifdef SC_ENABLE_PTHREAD
+  sc_package_lock (t8_package_id);
+#endif
```
sandro-elsweijer: Why do we need to lock this?
Furthermore, the `sc_package_lock` function does not need the compiler macros.

src/t8.c
```diff
@@ -34,7 +35,20 @@ t8_get_package_id (void)
...
+  vsnprintf (buffer, BUFSIZ, fmt, ap);
+#ifdef SC_ENABLE_PTHREAD
+  sc_package_unlock (t8_package_id);
+#endif
```
sandro-elsweijer: Same here

src/t8.h
```diff
@@ -263,6 +263,13 @@ t8_errorf (const char *fmt, ...)
 #endif
   ;
 
+/**
+ * Set a custom logging function to be used by t8code.
```
sandro-elsweijer, line 267:
```suggestion
 * Set a custom logging function to be used by t8code.
 * When setting a custom logging function, the t8code internal logging function will be ignored.
```

sandro-elsweijer (changes requested):
Thank you for your proposal! I only have minor comments.

sandro-elsweijer:
@dutkalex Do not worry about the CI, our Fortran interface test seems to be buggy and fails sometimes. I just restarted it and now everything seems fine :)

PR #1758, opened 2025-07-08
Improvement: Implement missing functions in vertex connectivity

### LINKED ISSUE #1759: Improvement: Implement missing functions in vertex connectivity

sandro-elsweijer:
Closes #1759

There were some declared, but not implemented functions in the vertex connectivity. Namely `treevertex_to_vertex`, `tree_to_vertices` and `vertex_to_trees`. They were not hard to implement, but I got sidetracked and had to improve other things as well.
- Instead of returning raw pointers to the global vertex arrays, the functions now return c++ views. This also makes it easier for the user, because they do not need to query how long this array is.
- Furthermore, the state system (committed, initialized and so on) was not type safe and contained magic numbers, because the states were returned as ints and the user could not query which numeral was which state. I still think this system has to be reworked, but this was all I could do in this time.
- The connectivity now saves a pointer to the associated cmesh during construction, so that it does not have to be passed at the function calls.
- The documentation was did not fit everywhere.

#### General
#### Tests

#### Scripts and Wiki
#### License

### DIFF
--- a/src/t8_cmesh/t8_cmesh_vertex_connectivity/t8_cmesh_vertex_connectivity.hxx
+++ b/src/t8_cmesh/t8_cmesh_vertex_connectivity/t8_cmesh_vertex_connectivity.hxx
@@ -34,6 +34,9 @@
 #include <t8_cmesh/t8_cmesh_vertex_connectivity/t8_cmesh_vertex_conn_vertex_to_tree.hxx>
 #include <t8_cmesh/t8_cmesh_vertex_connectivity/t8_cmesh_vertex_conn_tree_to_vertex.hxx>
 
+/**
+ * A class to hold the vertex connectivity of a cmesh.
+ */
 struct t8_cmesh_vertex_connectivity
 {
  public:
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -254,6 +254,8 @@ install( FILES
     t8_types/t8_vec.h
     t8_version.h
     t8_vtk.h
+    t8_cmesh/t8_cmesh_vertex_connectivity/t8_cmesh_vertex_connectivity.hxx
+    t8_cmesh/t8_cmesh_vertex_connectivity/t8_cmesh_vertex_connectivity.h
     TYPE INCLUDE
 )
 
--- a/src/t8.c
+++ b/src/t8.c
@@ -24,6 +24,7 @@
 #include <t8_version.h>
 
 static int t8_package_id = -1;
+static void (*external_log_fcn) (int category, int priority, const char *msg) = NULL;
 
 int
 t8_get_package_id (void)
@@ -34,7 +35,17 @@ t8_get_package_id (void)
 void
 t8_logv (int category, int priority, const char *fmt, va_list ap)
 {
-  sc_logv ("unknown", -1, t8_package_id, category, priority, fmt, ap);
+  char buffer[BUFSIZ];
+
+  sc_package_lock (t8_package_id);
+  vsnprintf (buffer, BUFSIZ, fmt, ap);
+  sc_package_unlock (t8_package_id);
+
+  if (external_log_fcn) {
+    external_log_fcn (category, priority, buffer);
+    return;
+  }
+  sc_log ("unknown", -1, t8_package_id, category, priority, buffer);
 }
 
 void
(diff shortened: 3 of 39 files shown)


src/t8_cmesh/t8_cmesh_vertex_connectivity/t8_cmesh_vertex_connectivity.hxx
```diff
@@ -126,28 +115,50 @@ struct t8_cmesh_vertex_connectivity
...
+      current_state = state::VTT_AND_TTV_VALID;
+    }
+    else if (vertex_to_tree.is_committed ())
+      current_state = state::VERTEX_TO_TREE_VALID;
+    else if (tree_to_vertex.get_state () == t8_cmesh_vertex_conn_tree_to_vertex::state::FILLED)
+      current_state = state::TREE_TO_VERTEX_VALID;
+    else
+      current_state = state::INITIALIZED;
+    /* Return state */
+    return current_state;
```
holke, line 130: This bit confuses me and at least requires thourough documentation.

- Please explain, why would you need to change the state here?
- The "get_state" function should be const and not change the state itself. If it changes the state, the name is misleading, in particular since for the other classes "get_state" is const.
- If there are cases where an update of the state is necessary, then i think it is better to have an "update_state" function that is called in the appropriate cases.

sandro-elsweijer: I see your problems with this and I will update this.
The problem was, that only the INITIALIZED and the VTT_AND_TTV_VALID state were ever used. The other two were implemented, but never updated or used. 
I tried to use at least the TREE_TO_VERTEX_VALID state, but that meant that the states would be kept up to date at all times. I could either update them in each function call or do a more efficient lazy evaluation.

I now removed the VERTEX_TO_TREE_VALID state, since we never use it and it also never happens. If we need it in the future, we can introduce it again without affecting user code.
I also added more assertions taking the state into account.

src/t8_cmesh/t8_cmesh_vertex_connectivity/t8_cmesh_vertex_connectivity.cxx
```diff
@@ -76,18 +56,21 @@ t8_cmesh_get_num_local_vertices (const t8_cmesh_t cmesh)
 }
 
 const t8_gloidx_t *
-t8_cmesh_get_global_vertices_of_tree (const t8_cmesh_t cmesh, const t8_locidx_t local_tree, const int num_vertices)
+t8_cmesh_get_global_vertices_of_tree (const t8_cmesh_t cmesh, const t8_locidx_t local_tree, int *num_vertices)
```
holke, line 59: `num_vertices` should match the data type of std::span.size() which is `size_t`.
This holds for the other functions as well.

holke (commented):
Looks good and i can approve after the following discussion:

i was at first sceptic about storing a cmesh pointer in the connectivity.
Because i though "what if we copy the cmesh" then two cmeshes use the same connectivity but the connectivity only has a  single cmesh pointer and we would need to do an expensive copy of the connectivity.

However, and please correct me if i am wrong here:
The cmesh pointer is only stored for the "vertex_connectivity" meta class and not for the subclasses.
This meta-class basically has no expensive data/information itself.
So what we would do in the case of copying a cmesh is to copy the meta-class, but let them point to the same "sub-classes" with data in them...
do i see that correctly?

Sidenote: Currently, t8_cmesh_copy does not take the vertex connectivity into account and kind of "forgets" it.
Thats an issue for another time...

holke:
Regarding my comment, see als #1799

src/t8_cmesh/t8_cmesh_vertex_connectivity/t8_cmesh_vertex_connectivity.cxx
sandro-elsweijer: done

sandro-elsweijer: This gets really nasty with type conversions because the global vertices themselves are `t8_gloidx_t`. We should bast change the global vertices to an unsigned datatype which makes more sense. Then we also do not need to cast everything back and forth. I will open an issue for this

PR #1768, opened 2025-07-17
Bugfix: local search

hannesbrandt:
This PR addresses some issues that I encountered while using the local search.
- In `check_queries` the query callback is called inside a `std::ranges::views::filter` used later to define the array `new_active_queries` based on the return value of the callback. On leaf level `new_active_queries` is not computed and due to lazy evaluation, the query callback is not executed on leaf level either. Furthermore, the assignment
`new_active_queries.assign (positive_queries.begin (), positive_queries.end ());`
leads to the same callback being evaluated twice for some indices. This PR replaces the evaluation by a copy_if, which seems to work fine, evaluating every callback exactly once.
- In `update_queries` and `check_queries` the active_queries array is swapped with the new_active_queries array. Since update queries is called after every recursion, this leads to the recursion being entered with the actual new_active_queries array and the old active_queries array in alternating order for the children. This PR replaces the swap with an assignment and moves update_queries before the first recursion.
- This PR proposes a `init_queries` function, that can be called once before every tree recursion to reset the queries to their original state.

#### General
#### Tests

#### Scripts and Wiki
#### License

### DIFF
--- a/src/t8_forest/t8_forest_search/t8_forest_search.hxx
+++ b/src/t8_forest/t8_forest_search/t8_forest_search.hxx
@@ -425,14 +425,11 @@ class t8_search_with_queries: public t8_search<Udata> {
   {
     T8_ASSERT (new_active_queries.empty ());
     if (!this->active_queries.empty ()) {
-      auto positive_queries = this->active_queries | std::ranges::views::filter ([&] (size_t &query_index) {
-                                return this->queries_callback (this->forest, ltreeid, element, is_leaf, leaf_elements,
-                                                               tree_leaf_index, queries[query_index], this->user_data);
-                              });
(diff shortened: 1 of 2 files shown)


src/t8_forest/t8_forest_search/t8_forest_search.hxx
```diff
@@ -308,6 +308,15 @@ class t8_search_base {
   virtual void
   update_queries (std::vector<size_t> &old_query_indices)
     = 0;
+
+  /**
+   * @brief Function gives the user the opportunity to set the queries to the initial
```
Davknapp:
```suggestion
   * Function gives the user the opportunity to set the queries to the initial
```

Issue #1771, opened 2025-07-18
Revise how cmeshes are generated


sandro-elsweijer:
# Feature request

Currently a cmesh is mostly generated from msh or vtk files or by our inbuilt cmesh generation functions. The workflow looks like (not exact, but shows the process):
```cpp
t8_cmesh_t cmesh = t8_cmesh_from_msh_file(filepath, do_partition, main_rank, communicator, dimension, curved_mesh);
//or
t8_cmesh_t cmesh = t8_vtk_reader_cmesh(filepath, do_partition, main_rank, communicator, vtk_file_type);
//or
t8_cmesh_t cmesh = t8_cmesh_new_hypercube (eclass, communicator, bcast, partition, periodic);
```
After that, the user has a committed cmesh. My problem with this approach is, that all options have to be passed to the cmesh generator. Currently, this is manageable, but there will be more options in the near future like if the negative volume check should be performed during commit, if the vertex connectivity should be built and if the cmesh should support mesh deformation. This will clutter our argument lists and we would need to keep them updated:
```cpp
t8_cmesh_t cmesh = t8_cmesh_from_msh_file(filepath, do_partition, main_rank, communicator, dimension, curved_mesh, do_negative_volume_check, create_vertex_connectivity, enable_mesh_deformation, ...[possible_future_options]...);
//or
t8_cmesh_t cmesh = t8_vtk_reader_cmesh(filepath, do_partition, main_rank, communicator, vtk_file_type, do_negative_volume_check, create_vertex_connectivity, enable_mesh_deformation, ...[possible_future_options]...);
//or
t8_cmesh_t cmesh = t8_cmesh_new_hypercube (eclass, communicator, bcast, partition, periodic, create_vertex_connectivity, ...[possible_future_options]...);
```

Instead, I would propose to set the options for the cmesh before handing it to a generator:
```cpp
t8_cmesh_t cmesh;
t8_cmesh_init(&cmesh, comm); // mandatory options during init
t8_cmesh_enable_partition (cmesh, main_rank); // enable partitioning
t8_cmesh_enable_vertex_conn (cmesh); // enable vertex conn
t8_cmesh_enable_cad_linkage (cmesh, cad_filepath, enable_curved_mesh, enable_mesh_deformation); // enable cad features
t8_cmesh_disable_negative_volume_check (cmesh); // disable negative volume check during commit

t8_cmesh_from_msh_file(cmesh, filepath, dim);
//or
[...]

Issue #865, opened 2023-11-28
Connectivity of msh file reader


sandro-elsweijer:
It seems like the face connectivity of the msh reader seems to be broken. @jfussbro experienced this with a tet mesh in parallel (in serial it worked) and now I experience the same in serial with a hybrid 2D mesh. For some reason the trees get glued together randomly.

holke:
Can you specify what the exact behavious is that you are getting and that you expected?

Could you provide a MWE?

lukasdreyer:
Big MWE:
./t8_advection -u7 -l1 -r2 -f ./channel -d3 -C0.5 -b0.5 -X-0.6 -Y-0.1 -Z0 -R0.4
![artifacts](https://github.com/DLR-AMR/t8code/assets/58432628/ae9bc49d-c0af-4613-959a-c1464fdf4387)
[channel.zip](https://github.com/user-attachments/files/15897930/channel.zip)

PR #1786, opened 2025-07-21
Improvement: Remove vertex renumbering and make negative volume check optional

### LINKED ISSUE #1241: Vertex renumbering for negative volumes not fully taken into account
# Bug report

**Describe the bug**
Vertex renumbering due to negative volumes is not taken into account for face-neighbor computation.
Only the tree_vertices array is switched, not the `node_indices` which is passed `vertex_indices` to `t8_cmesh_msh_file_find_neighbors` and used to compute face indices of neighboring elements for `t8_cmesh_set_join`.

Additionally, only switching vertices 0 and 4 for pyramids does not seem correct.
 
Also see:

#1220 and #865 

**Estimated priority**
Which of these is most applicable (remove the others):

"Priority: high"  Should be solved as soon as possible

sandro-elsweijer:
We decided to not mess with user data and shuffle vertices in meshes around. Furthermore, the negative volume check for trees should now be optional for experienced users.
The second negative volume check (yes, for some reason we had two) was completely removed, since it makes assumptions about the mesh geometry and only allows 2d meshes in specific orientations.

closes #1241
related to #1771
maybe this also fixes the problem mentioned in #865

#### General
#### Tests

#### Scripts and Wiki
#### License


src/t8_cmesh/t8_cmesh.cxx
```diff
@@ -128,9 +128,17 @@ t8_cmesh_is_committed (const t8_cmesh_t cmesh)
...
+{
+#if T8_ENABLE_DEBUG
+  cmesh->negative_volume_check = 0;
+#endif
+}
+
 #if T8_ENABLE_DEBUG
 int
-t8_cmesh_validate_geometry (const t8_cmesh_t cmesh)
+t8_cmesh_validate_geometry (const t8_cmesh_t cmesh, const int check_for_negative_volume)
```
Davknapp, line 141: Can this be a bool?

test/t8_cmesh/t8_gtest_cmesh_tree_vertices_negative_volume.cxx
Davknapp, line 1: The negative volume check is still available, right? So we should test its functionality, too.

src/t8_cmesh/t8_cmesh.cxx
sandro-elsweijer: No since it is made public in a C header file :(
t8_cmesh.h

test/t8_cmesh/t8_gtest_cmesh_tree_vertices_negative_volume.cxx
sandro-elsweijer: Sort of.
We actually had two different negative volume check:
One function, which calculated the volume of each cell according to the vertex position and regardless of the trees geometry (which only works for linear geometries) and which also checked the orientation of 2D cells so that they face in the same direction. But since not every 2D mesh is a 2D plane, it crashed also on valid meshes as soon as a cell was in the yz or xz plane. I therefore removed it.
The other check loops over every tree and asks the geometry if the volume is negative. This is currently only implemented for the linear and cartesian geometry, but at least its better than the first one.
No, there is not test for this and since I only deleted things and did not implement anything new, I will open an issue for this :)

sandro-elsweijer:
Converted to draft because I need some general changes in our cmake build system

sandro-elsweijer:
Okay its allright

Davknapp (approved):

Davknapp:
Please resolve the merge conflict, then we are ready to merge into main!

Davknapp (approved):

PR #1884, opened 2025-09-30
Feature: add t8_cmesh_get_mpicomm

dutkalex:
This PR adds the possibility to query a cmesh for its associated MPI communicator.

#### General
#### Tests

#### Scripts and Wiki
#### License


src/t8_cmesh.h
```diff
@@ -480,6 +480,13 @@ t8_cmesh_load_and_distribute (const char *fileprefix, int num_files, sc_MPI_Comm
 int
 t8_cmesh_comm_is_valid (t8_cmesh_t cmesh, sc_MPI_Comm comm);
```
sandro-elsweijer, line 479: Can you please also adapt the documentation?

src/t8_cmesh/t8_cmesh.cxx
```diff
@@ -1245,7 +1242,7 @@ t8_cmesh_reset (t8_cmesh_t *pcmesh)
      * This is useful for debugging. */
     if (t8_cmesh_is_committed (cmesh)) {
       comm = t8_shmem_array_get_comm (cmesh->tree_offsets);
-      T8_ASSERT (t8_cmesh_comm_is_valid (cmesh, comm));
+      T8_ASSERT (t8_cmesh_get_mpicomm (cmesh) == comm);
```
sandro-elsweijer, line 1245: I would still use `t8_cmesh_comm_is_valid ` here because of the straighforward naming
```suggestion
      T8_ASSERT (t8_cmesh_comm_is_valid (cmesh, comm);
```

src/t8_cmesh/t8_cmesh.cxx
```diff
@@ -871,7 +868,7 @@ t8_cmesh_bcast (const t8_cmesh_t cmesh_in, const int root, sc_MPI_Comm comm)
   SC_CHECK_MPI (mpiret);
   if (!meta_info.pre_commit) {
     T8_ASSERT (t8_cmesh_is_committed (cmesh_out));
-    T8_ASSERT (t8_cmesh_comm_is_valid (cmesh_out, comm));
+    T8_ASSERT (t8_cmesh_get_mpicomm (cmesh_out) == comm);
```
sandro-elsweijer, line 871: I would still use `t8_cmesh_comm_is_valid ` here, because of the naming
```suggestion
    T8_ASSERT (t8_cmesh_comm_is_valid (cmesh, comm);
```

src/t8_cmesh/t8_cmesh_commit.cxx
```diff
@@ -234,7 +234,7 @@ t8_cmesh_commit_partitioned_new (t8_cmesh_t cmesh, sc_MPI_Comm comm)
   sc_flops_snap (&fi, &snapshot);
 #endif
 
-  T8_ASSERT (t8_cmesh_comm_is_valid (cmesh, comm));
+  T8_ASSERT (t8_cmesh_get_mpicomm (cmesh) == comm);
```
sandro-elsweijer, line 237: I would still use `t8_cmesh_comm_is_valid ` here because of the straighforward naming
```suggestion
  T8_ASSERT (t8_cmesh_comm_is_valid (cmesh, comm);
```

src/t8_cmesh/t8_cmesh_copy.c
```diff
@@ -54,7 +54,7 @@ t8_cmesh_copy (t8_cmesh_t cmesh, const t8_cmesh_t cmesh_from, sc_MPI_Comm comm)
   cmesh->num_trees = cmesh_from->num_trees;
   cmesh->set_partition = cmesh_from->set_partition;
   cmesh->set_partition_level = cmesh_from->set_partition_level;
-  T8_ASSERT (t8_cmesh_comm_is_valid (cmesh, comm));
+  T8_ASSERT (t8_cmesh_get_mpicomm (cmesh) == comm);
```
sandro-elsweijer, line 57: I would still use `t8_cmesh_comm_is_valid ` here because of the straighforward naming
```suggestion
  T8_ASSERT (t8_cmesh_comm_is_valid (cmesh, comm);
```

src/t8_cmesh/t8_cmesh_partition.cxx
```diff
@@ -131,7 +131,7 @@ t8_cmesh_gather_trees_per_eclass (const t8_cmesh_t cmesh, sc_MPI_Comm comm)
   t8_gloidx_t temp_trees_per_eclass[T8_ECLASS_COUNT];
   int ieclass;
 
-  T8_ASSERT (t8_cmesh_comm_is_valid (cmesh, comm));
+  T8_ASSERT (t8_cmesh_get_mpicomm (cmesh) == comm);
```
sandro-elsweijer, line 134: I would still use `t8_cmesh_comm_is_valid ` here because of the straighforward naming
```suggestion
  T8_ASSERT (t8_cmesh_comm_is_valid (cmesh, comm);
```

src/t8_cmesh/t8_cmesh_partition.cxx
```diff
@@ -191,7 +191,7 @@ t8_cmesh_gather_treecount_ext (const t8_cmesh_t cmesh, sc_MPI_Comm comm, const i
   if (check_commit) {
     T8_ASSERT (t8_cmesh_is_committed (cmesh));
   }
-  T8_ASSERT (t8_cmesh_comm_is_valid (cmesh, comm));
+  T8_ASSERT (t8_cmesh_get_mpicomm (cmesh) == comm);
```
sandro-elsweijer, line 194: I would still use `t8_cmesh_comm_is_valid ` here because of the straighforward naming
```suggestion
  T8_ASSERT (t8_cmesh_comm_is_valid (cmesh, comm);
```

src/t8_cmesh/t8_cmesh_partition.cxx
```diff
@@ -983,7 +983,7 @@ t8_cmesh_partition_sendloop (t8_cmesh_t cmesh, t8_cmesh_t cmesh_from, int *num_r
   int8_t *ghost_flag_send;  /* For each local tree and ghost set to 1 if it is in send_as_ghost */
   const t8_gloidx_t *offset_from, *offset_to;
 
-  T8_ASSERT (t8_cmesh_comm_is_valid (cmesh, comm));
+  T8_ASSERT (t8_cmesh_get_mpicomm (cmesh) == comm);
```
sandro-elsweijer, line 986: I would still use `t8_cmesh_comm_is_valid ` here because of the straighforward naming
```suggestion
  T8_ASSERT (t8_cmesh_comm_is_valid (cmesh, comm);
```

src/t8_cmesh/t8_cmesh_partition.cxx
```diff
@@ -1793,7 +1793,7 @@ t8_cmesh_offset_percent (const t8_cmesh_t cmesh, sc_MPI_Comm comm, const int per
 #endif
 
   T8_ASSERT (t8_cmesh_is_committed (cmesh));
-  T8_ASSERT (t8_cmesh_comm_is_valid (cmesh, comm));
+  T8_ASSERT (t8_cmesh_get_mpicomm (cmesh) == comm);
```
sandro-elsweijer, line 1796: I would still use `t8_cmesh_comm_is_valid ` here because of the straighforward naming
```suggestion
  T8_ASSERT (t8_cmesh_comm_is_valid (cmesh, comm);
```

src/t8_forest/t8_forest.cxx
```diff
@@ -2756,7 +2756,7 @@ t8_forest_set_cmesh (t8_forest_t forest, t8_cmesh_t cmesh, sc_MPI_Comm comm)
     t8_cmesh_unref (&forest->cmesh);
   }
   if (cmesh != NULL) {
-    T8_ASSERT (t8_cmesh_comm_is_valid (cmesh, comm));
+    T8_ASSERT (t8_cmesh_get_mpicomm (cmesh) == comm);
```
sandro-elsweijer, line 2759: I would still use `t8_cmesh_comm_is_valid ` here because of the straighforward naming
```suggestion
    T8_ASSERT (t8_cmesh_comm_is_valid (cmesh, comm);
```

sandro-elsweijer (changes requested):
Thank you for the implementation! Only minor comments

src/t8_cmesh.h
dutkalex: Done

src/t8_cmesh/t8_cmesh.cxx
dutkalex: The change was proposed because it felt clearer that way, since the term "valid" is kind of vague in this context. I would personally argue for endorsing the `==` spelling as the canonical way of checking this and deprecating the `is_valid` spelling, but if you guys feel otherwise I'm not against reverting this either

sandro-elsweijer: I just had another look concerning the `is_valid` function and there is an easily addressable problem:
The cmesh now saves the comm along with the mpirank and mpisize. Then afterwards, `is_valid` only checks if the already saved comm and the provided comm are the same. The mpisize and rank are not checked. This can lead to possible errors in other routines which create a new cmesh. For example in [`t8_cmesh_copy ()`](https://github.com/DLR-AMR/t8code/blob/50b167edaacd197020f110cb63009534fd3f792d/src/t8_cmesh/t8_cmesh_copy.c#L36).
This function takes an empty cmesh, a committed cmesh and a communicator. The new cmesh copies everything of the committed cmesh, including the mpisize and rank. Then the provided comm is checked with the new cmesh. In your case this would lead to a segfault, because the copy routine does not fill the comm of the new cmesh. But even if it filled the new one with the provided comm, the `is_valid` function would return true, even though it was never checked if the comm has the same mpisize and rank as the already saved rank and size. So `is_valid` is still necessary and should not only check if the comms are the same, but also if the comm is in alignment with the stored mpisize and rank.

sandro-elsweijer (commented):
We should check also the other cmesh creation routines like copy and bcast

src/t8_cmesh/t8_cmesh.cxx
dutkalex: I agree that there is indeed room for inconsistency here, which could lead to subtle bugs, so this can't make it into production as-is.
A possible solution is effectively to harden the checks to have this bug class be catched by the debug mode. Although probably effective for debugging purposes, this approach does not prevent the bugs from appearing in the first place, thus implying a (hopefully minor) maintenance burden in the long run.
Another solution could be prevent the inconsistency from appearing by only storing the communicator which already holds all the relevant information. The rank and size could then be queried when necessary using `MPI_Comm_rank` / `MPI_Comm_size`. This solution requires a little bit more effort right now, but is arguably superior because it would completely eliminate this class of bugs.

dutkalex: @sandro-elsweijer While reading the forest implementation, it seems like this `mpicomm`+`mpirank`+`mpisize` pattern could lead to the same kind of issues on the forest side. Maybe this should be addressed in a separate PR then, and your solution is the best strategy for this PR. What do you think?

dutkalex:
> We should check also the other cmesh creation routines like copy and bcast

I am not familiar with these functions. What is precisely the expected behavior for each of them?

PR #1889, opened 2025-10-09
Feature: extend the partition algorithm to handle weights

dutkalex:
This PR makes it possible to define element weights for the partitioning algorithm. This is achieved via a callback pointer. In the null case, the old unweighted algorithm is used. ~This introduces one breaking change in the public API, in the `t8_forest_set_partition` function signature. All the client code in the repo has been updated, but this will require an update of external user code: `t8_forest_set_partition(forest, set_from, set_for_coarsening)` => `t8_forest_set_partition(forest, set_from, set_for_coarsening, nullptr)`~

*Design tradeoffs:*
- ~The new algorithm replaces the old one, even if no weight function is provided. Keeping the old algorithm around could possibly result in slightly better performance since it requires less work, but this would come at the cost of increased maintenance and the need for duplicated testing to cover both code paths. Given that the partition algorithm is not a performance bottleneck in practice, the single code path solution was preferred.~ The final design includes the unweighted case as a fast "early exit" case.
- In the general case, a single pass algorithm is not possible because the total weight of the forest wan't be known _a priori_ and must be computed (first traversal of the forest) to then be able to define the new partition boundaries (second traversal). This means that each element weight is computed twice with the current implementation. If the weight function is expensive to evaluate, this could cause performance problems. There are basically two ways to work around this: the results could either be cached during the first evaluation (this means that we need to allocate a temporary array), or the weights could alternatively be computed and stored in an array ahead-of-time by the user, before calling the partition function. In the second case the user then simply provides a weight function that looks up the precomputed weights. This second option was preferred because it gives the user greater flexibility, and because it offers optimal performance in the case where the weight function is cheap to evaluate which is probably the default case (_"Don't pay for what you don't use"_ principle).

#### General
#### Tests

#### Scripts and Wiki
#### License

### DIFF
--- a/.github/workflows/test_valgrind.yml
+++ b/.github/workflows/test_valgrind.yml
@@ -21,22 +21,40 @@ name: valgrind check
 #  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
 
 on:
+  # Allows you to run this workflow manually.
+  workflow_dispatch:
+    inputs:
+      MAKEFLAGS:
+        type: string
+        description: 'Make flags to use for compilation (like -j4)'
+        default: '-j4'
+      MPI:
+        type: string
+        description: 'Use MPI for compilation (ON/OFF)'
+        default: 'ON'
+      BUILD_TYPE:
+        type: string
+        description: 'Build type (Release/Debug)'
+        default: 'Debug'
+      TEST_LEVEL:
+        type: string
+        description: 'Test level used for configuring (T8_TEST_LEVEL_FULL, T8_TEST_LEVEL_MEDIUM, or T8_TEST_LEVEL_BASIC)'
+        default: 'T8_TEST_LEVEL_BASIC'
   workflow_call:
     inputs:
       MAKEFLAGS:
-        required: true
         type: string
         description: 'Make flags to use for compilation (like -j4)'
+        default: '-j4'
       MPI:
-        required: true
         type: string
         description: 'Use MPI for compilation (ON/OFF)'
+        default: 'ON'
       BUILD_TYPE:
-        required: true
         type: string
         description: 'Build type (Release/Debug)'
+        default: 'Debug'
       TEST_LEVEL:
-        required: true
         type: string
         description: 'Test level used for configuring (T8_TEST_LEVEL_FULL, T8_TEST_LEVEL_MEDIUM, or T8_TEST_LEVEL_BASIC)'
         default: 'T8_TEST_LEVEL_FULL'
@@ -101,9 +119,7 @@ jobs:
         path: build/CMakeFiles/CMakeOutput.log
     - name: ninja
       run: cd build && ninja $MAKEFLAGS
-    - name: ninja install
-      run: cd build && ninja install $MAKEFLAGS
     # Execute script that runs a Valgrind check for all test binaries.
-    # We use 3 parallel processes to run the binaries in parallel with MPI.
+    # We use 4 parallel processes to run the binaries in parallel with MPI.
     - name: Valgrind check
-      run: cd scripts && bash ./check_all_test_binaries_valgrind.sh --ntasks=3
+      run: cd scripts && bash ./check_all_test_binaries_valgrind.sh --ntasks=4
--- a/.github/workflows/testsuite.yml
+++ b/.github/workflows/testsuite.yml
@@ -194,9 +194,6 @@ jobs:
     needs: preparation
     uses: ./.github/workflows/test_valgrind.yml
     with:
-      MAKEFLAGS: -j4
-      MPI: ON
-      BUILD_TYPE: Debug
       TEST_LEVEL: T8_TEST_LEVEL_BASIC # Do Valgrind check for test level T8_TEST_LEVEL_BASIC for performance reasons.
 
   # Generate code coverage and deploy to Codecov.
--- a/mesh_handle/README.md
+++ b/mesh_handle/README.md
@@ -6,6 +6,7 @@ Some application codes are designed for unstructured or uniform meshes and canno
 
 If you want to use the handle, note that is has its own library. Turn the option `T8CODE_BUILD_MESH_HANDLE` to `ON` and link against the target `T8_MESH_HANDLE` in addition to the usual t8code target please.
 
-The [mesh.hxx](mesh.hxx) defines the mesh class of the handle.
-The [element.hxx](element.hxx) defines the elements used in the mesh class.
+The folder's most important files are: 
+The [mesh.hxx](mesh.hxx) defines the mesh class of the handle. This is the central file of the mesh handle. 
+The [element.hxx](element.hxx) defines the elements (mesh or ghost elements) of the mesh handle.
 The [competences.hxx](competences.hxx) defines additional competences/functionality of an element to access additional data.
--- a/mesh_handle/competences.hxx
+++ b/mesh_handle/competences.hxx
@@ -26,7 +26,8 @@ along with t8code; if not, write to the Free Software Foundation, Inc.,
  * is called are provided.
  *
  * All competences have the same inheritance pattern: 
- * We use the CRTP pattern as we may need to access members of the derived class \ref t8_mesh_handle::element. 
+ * We use the CRTP pattern as we may need to access members of the derived classes like 
+ * \ref t8_mesh_handle::element. 
  * The t8_crtp_operator is used for convenience/clear code (avoid to type a static cast explicitly each time 
  * we need functionality of TUnderlying).
  * Especially for the competences to cache functionality, the access of members is not necessary, 
@@ -41,8 +42,6 @@ along with t8code; if not, write to the Free Software Foundation, Inc.,
 #include <t8.h>
 #include <t8_types/t8_operators.hxx>
 #include <t8_types/t8_vec.hxx>
-#include <t8_eclass.h>
-#include <array>
 #include <vector>
 #include <optional>
 
@@ -93,6 +92,50 @@ struct cache_centroid: public t8_crtp_operator<TUnderlying, cache_centroid>
  protected:
   mutable std::optional<t8_3D_point>
     m_centroid; /**< Cache for the coordinates of the centroid. Use optional to allow no value if cache is not filled. */
+};
+
+/**
+ * Competence to cache the neighbors of an element at a specific face at the first function call.
+ * \tparam TUnderlying Use the \ref element with specified competences as template parameter.
+ */
+template <typename TUnderlying>
+struct cache_neighbors: t8_crtp_operator<TUnderlying, cache_neighbors>
+{
+ public:
+  /**
+   * Function that checks if the neighbor cache for a face has been filled.
+   * \param [in] face The face for which the cache should be checked.
+   * \return true if the cache has been filled, false otherwise.
+   */
+  bool
+  neighbor_cache_filled (int face) const
+  {
+    return m_num_neighbors[face].has_value ();
+  }
+
+  /**
+   * Function that checks if the neighbor cache for any face has been filled.
+   * \return true if the cache has been filled, false otherwise.
+   */
+  bool
+  neighbor_cache_filled_any () const
+  {
+    for (int iface = 0; iface < this->underlying ().get_num_faces (); ++iface) {
+      if (neighbor_cache_filled (iface)) {
+        return true;
+      }
+    }
+    return false;
+  }
+
+ protected:
+  mutable std::vector<std::vector<const TUnderlying *>>
+    m_neighbors; /**< Neighboring elements at each face. The length of the vectors is stored in \ref m_num_neighbors. */
+  mutable std::vector<std::optional<int>>
+    m_num_neighbors; /**< Vector with the numbers of neighbor elements at each face. 
+                        num_neighbors is stored to indicate that the cache is filled if a face does not have any neighbor. */
+  mutable std::vector<std::vector<int>>
+    m_dual_faces; /**< Face id's of the neighboring elements' faces for each face. */
 };
 
 }  // namespace t8_mesh_handle
--- a/mesh_handle/element.hxx
+++ b/mesh_handle/element.hxx
@@ -21,7 +21,7 @@ along with t8code; if not, write to the Free Software Foundation, Inc.,
 */
 
 /** \file element.hxx
- * Definition of the elements used in the \ref t8_mesh_handle::mesh class.
+ * Definition of the element class of the \ref t8_mesh_handle::mesh handle (can be ghost or mesh elements).
  */
 
 #ifndef T8_ELEMENT_HXX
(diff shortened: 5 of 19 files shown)


holke:
Thanks again for this addition!

Since 95% of our users do not use weighted partitioning but the classical one and the weighted does add overhead i would argue we should at least skip the weight computation loop in standard mode.
Or keep the standard version as is - but i also see your point of avoiding code duplication which is very important.

dutkalex:
> Since 95% of our users do not use weighted partitioning but the classical one and the weighted does add overhead i would argue we should at least skip the weight computation loop in standard mode.

That's an interesting idea. Maybe we can get the best of both worlds with this kind of "fast-forward" behavior in the simple case. I will look into that.

dutkalex:
I wonder what should be the input of the weight function. Currently it takes a forest, a local tree id and an element id (within that tree). ~I'm not quite convinced this is really usable in practice however, especially if the forest has just been refined or coarsened prior to entering the partition algorithm. Do you have ideas on what would be the right function signature in the general case?~
Since the weight function is always used on a committed forest, then all the non mutating public API functions can be used by this weight function to query for additional information if needed.

dutkalex:
I have refactored to use `t8_shmem_array_allgatherv` in the weighted case as you suggested @holke, but it seems like I'm still missing something to make this work...
https://github.com/dutkalex/t8code/blob/weighted-partition/src/t8_forest/t8_forest_partition.cxx#L519

src/t8_forest/t8_forest_partition.cxx
```diff
@@ -430,49 +430,100 @@ t8_forest_partition_create_tree_offsets (t8_forest_t forest)
...
+  double const forest_weight = [&] () {  // complete sum of the partition weights
+    double total_weight = partition_weight_offset + partition_weight;
+    sc_MPI_Bcast (&total_weight, 1, sc_MPI_DOUBLE, mpisize - 1, comm);
+    return total_weight;
+  }();
+
+  // The [rank_begin, rank_end) slice of the new offsets land in the local partition
+  int const rank_begin = std::ceil (mpisize * partition_weight_offset / forest_weight);
+  int const rank_end = std::ceil (mpisize * (partition_weight_offset + partition_weight) / forest_weight);
+  std::vector<t8_gloidx_t> local_offsets { rank_end - rank_begin, 0 };
```
spenke91: I think this might be the main problem here:
The curly brackets indicate an initializer list, meaning it creates a vector with two entries: [0] `rank_end -  rank_begin`, [1] `0`.
To create a vector of length `rank_end -  rank_begin` filled with zeros, as intended here, we should use parentheses:

```suggestion
  std::vector<t8_gloidx_t> local_offsets ( rank_end - rank_begin, 0 );
```
I did not look into your implementation in detail (yet), but with this fix, all test cases, including `t8_gtest_forest_commit_parallel` pass 👍

dutkalex: Oh you're right! Thanks for spotting that :)
I'll double check that everything works as expected, but if the tests pass I'm quite confident that the feature works.

spenke91:
Thanks a lot for your contribution @dutkalex ! I am out of office this week, but looking forward to reviewing it next week 👍

src/t8_forest/t8_forest_partition.h
```diff
@@ -36,11 +36,14 @@ T8_EXTERN_C_BEGIN ();
...
- * 
+ *
  * \param [in,out]  forest  The forest.
+ * \param [in]      weight_callback A callback function defining element weights for the partitioning
+ * \pre \a weight_callback must be free of side-effects, the behavior is undefined otherwise
+ * \note If \a weight_callback is null, then all the elements are assumed to have the same weight
 */
 void
-t8_forest_partition (t8_forest_t forest);
+t8_forest_partition (t8_forest_t forest, t8_weight_fcn_t *weight_callback);
```
spenke91: Does it make sense to add the `weight_callback` as an extra argument here? To me, it seems like an unnecessary duplication, given that the forest already has the new member `weight_function`. 
My suggestion would be to omit the extra argument here and then use `forest->weight_function` within `t8_forest_partition` rather than the function argument? Otherwise my impression is that we are bound to sooner or later run into hard-to-debug problems of the argument and the member being out of sync.

What do you think?

spenke91: Does it make sense to add the `weight_callback` as an extra argument here? To me, it seems like an unnecessary duplication, given that the forest already has the new member `weight_function`. 
My suggestion would be to omit the extra argument here and then use `forest->weight_function` within `t8_forest_partition` rather than the function argument? Otherwise my impression is that we are bound to sooner or later run into hard-to-debug problems of the argument and the member being out of sync.

What do you think?

```suggestion
t8_forest_partition (t8_forest_t forest);
```

src/t8_forest/t8_forest_partition.cxx
```diff
@@ -430,49 +430,101 @@ t8_forest_partition_create_tree_offsets (t8_forest_t forest)
   }
 }
 
-/* Calculate the new element_offset for forest from
- * the element in forest->set_from assuming a partition without element weights */
+// Compute forest->element_offsets according to the weight function, if provided
 static void
-t8_forest_partition_compute_new_offset (t8_forest_t forest)
+t8_forest_partition_compute_new_offset (t8_forest_t forest, t8_weight_fcn_t *weight_fcn)
```
spenke91: same as in my comment below for `t8_forest_partition`: How about omitting the new argument here and using `forest->weight_function` within this function instead?

src/t8_forest/t8_forest_partition.cxx
```diff
@@ -430,49 +430,101 @@ t8_forest_partition_create_tree_offsets (t8_forest_t forest)
...
-      }
-      element_offsets[forest->mpisize] = forest->global_num_leaf_elements;
+      std::fill_n (element_offsets, mpisize + 1, t8_gloidx_t { 0 });
     }
-    else {
+    t8_shmem_array_end_writing (forest->element_offsets);
+    return;
+  }
+
+  if (not weight_fcn) {
```
spenke91:
```suggestion
  // If no weight function is provided, compute the offsets solely based on the number of elements.
  if (not weight_fcn) {
```

src/t8_forest/t8_forest_general.h
```diff
@@ -322,7 +330,8 @@ t8_forest_get_user_function (const t8_forest_t forest);
  * this setting.
  */
 void
-t8_forest_set_partition (t8_forest_t forest, const t8_forest_t set_from, int set_for_coarsening);
+t8_forest_set_partition (t8_forest_t forest, const t8_forest_t set_from, int set_for_coarsening,
+                         t8_weight_fcn_t *weight_callback);
```
spenke91: After discussing with my colleagues, I suggest not to introduce the remove the weight function as an argument of `t8_forest_set_partition`, but to instead introduce a separate function `t8_forest_set_partition_weights` that assigns weight function to a forest.
This way, we can avoid having to update all examples, tutorials. etc and do not need to deal with nullptr that much.

test/t8_forest/t8_gtest_forest_commit.cxx
```diff
@@ -114,32 +114,29 @@ t8_test_forest_commit_abp (t8_forest_t forest, int maxlevel)
...
   /* balance the forest */
+  t8_forest_t forest_balance;
+  t8_forest_init (&forest_balance);
   t8_forest_set_balance (forest_balance, forest_adapt, 0);
   t8_forest_commit (forest_balance);
 #endif
 
-  /* partition the forest */
+  /* partition the forest with a unit weight function */
+  t8_forest_t forest_partition;
```
spenke91, line 133: I would suggest to revert this test case to its original state, since it is now really about testing the weighted load balancing. Do you happen to already have some other test case you used locally for validation? Otherwise, I can also add a suitable test case in a subsequent PR.

src/t8_forest/t8_forest_partition.cxx
```diff
@@ -430,49 +430,101 @@ t8_forest_partition_create_tree_offsets (t8_forest_t forest)
...
-        element_offsets[i] = 0;
+      for (int i = 0; i < mpisize; ++i) {
+        element_offsets[i] = std::floor (static_cast<double> (global_num_leaf_elements) * i / mpisize);
+      }
+      element_offsets[mpisize] = global_num_leaf_elements;
+    }
+    t8_shmem_array_end_writing (forest->element_offsets);
+    return;
+  }
+
```
spenke91:
```suggestion
// Weighted load balancing:
// ------------------------ 
```

src/t8_forest/t8_forest_partition.cxx
```diff
@@ -430,49 +430,101 @@ t8_forest_partition_create_tree_offsets (t8_forest_t forest)
...
+    double local_sum = 0.;
+    for (t8_locidx_t ltreeid = 0; ltreeid < t8_forest_get_num_local_trees (forest_from); ++ltreeid) {
+      for (t8_locidx_t ielm = 0; ielm < t8_forest_get_tree_num_leaf_elements (forest_from, ltreeid); ++ielm) {
+        local_sum += weight_fcn (forest_from, ltreeid, ielm);
+      }
+    }
+    return local_sum;
+  }();
+
+  double const partition_weight_offset = [&] () {  // partial sum of the partition weights, excluding the local rank
```
spenke91:
```suggestion
  double const partition_weight_offset = [&] () {  // partial sum of the partition weights of all lower-rank processes (excluding the local rank)
```

src/t8_forest/t8_forest_partition.h
dutkalex: I'm actually on the fence regarding this, and it probably shows in the code...

You definitely have a point here, and this problem should be addressed. 

I was actually tempted at some point to simply remove the additional parameter and rely on `forest->weight_function` as you suggest, but from a design perspective this feels just wrong. Conceptually, the weight function is not a property of the forest data structure and should not be stored in the forest. The weight function is a user customization point of the partitioning algorithm. It therefore makes much more sense to have the callback be passed explicitly to the partition algorithm (like you would do when calling `std::transform` or `std::reduce` for exemple), and get rid of `forest->weight_function` instead. However, I could not find a way to make this work because the current API for forest mutation is fundamentally designed around this idea of attaching these customization points to the forest.

This is obviously a problem that goes beyond the scope of this feature and that will have to be addressed at some point. In the meantime, we can either:
- Go with the simple and consistent approach, but this means doubling down on this tech debt.
- Implement a proper separation of concerns in new code to avoid making things worse, at the cost of some inconsistencies across the codebase, and eventually some breaking changes in the APIs.

src/t8_forest/t8_forest_general.h
dutkalex: I 100% hear the backwards compatibility argument, but this would result in a user API that is quite counter-intuitive. Moreover, I would argue that this API will have to be reworked at some point anyway, because introducing a new function for each new customization point is not sustainable long-term

test/t8_forest/t8_gtest_forest_commit.cxx
dutkalex: I agree that this is not the best approach to test this feature, and adding a dedicated test suite for the partitioning algorithm is the right way to do things. However, beyond comparing the result of the unweighted algorithm with the weighted one with a constant weight callback, I don't know how to test this in a robust way. For development purposes, I have used a lot of intrusive whitebox testing manually, but this would not be very helpful in a proper automated test suite...

src/t8_forest/t8_forest_partition.h
dutkalex: For reference, here is a possible solution to make these APIs extensible and future-proof:
```cpp
// public_api.h
struct t8_partition_options {
  t8_weight_fcn_t* weight_fcn;
  int set_for_coarsening;
};
extern const t8_partition_options default_values;

t8forest_t t8_partition(t8forest_t forest_from, t8_partition_options options);

// client_code.c
t8_partition_options partition_options = default_values;
partition_options.weight_fcn = /*...*/ ;
t8forest_t new_forest = t8_partition( old_forest, partition_options );
```

test/t8_forest/t8_gtest_forest_commit.cxx
spenke91: Ok, then I suggest you leave the test case as it currently is in this PR and I will move the weighted partitioning to another test case in a later PR.

src/t8_forest/t8_forest_partition.h
spenke91: Thank you for the detailed explanation. I discussed with @holke about this today. We do see your point about the design flaws of storing information like the weight function as member data of the forest; however, we would still suggest to stick to it in this PR and go for a function `t8_forest_set_partition_weights` for the following reasons:
1. it would be consistent with how this kind of data is handled so far in t8code (whether we like it or not);
2. it would render / keep this PR rather non-intrusive, in particular by not changing the function interfaces;
3. we might change this forest-based design in the future, but we do not have immediate plans to do so (because we also don't have the resources right now). Once we change it, it would be a quite deep change that won't get significantly more extensive by the weighted load balancing.

Would it be okay with you if we go forward with this approach?
Speaking of "we": If you prefer that I can also take over the PR and change it accordingly...

src/t8_forest/t8_forest_general.h
spenke91: As touched upon above, you are right that this strategy does not scale well with an increasing number of options. For this specific feature, however, we still think it would be preferable to go along with it, because it is so simple.

src/t8_forest/t8_forest_partition.h
dutkalex: > Would it be okay with you if we go forward with this approach?

That makes perfect sense, and I understand the consistency and backwards compatibility argument. I just wanted to make sure that this specific design aspect was taken into account before moving forward with this. At the end of the day, you guys have the broader and more insightful vision into this matter, and I trust your jugement.

> Speaking of "we": If you prefer that I can also take over the PR and change it accordingly...

I guess it depends on how fast you'd like this to get merged: I'm a little bit over-subscribed at the moment so it will probably have to wait a little bit on my side. If you want to get this merged now, feel free to take over from here.

test/t8_forest/t8_gtest_forest_commit.cxx
dutkalex: Sharing a (probably half-baked) thought I had this morning about how to test this feature in the general case in a non-intrusive and hopefully robust way:
Given a non-trivial weight function, we could manually compute _a posteriori_ each partition's weight, find the global min and max, and assert that the difference between the two is of the order of magnitude of a single element's weight.

src/t8_forest/t8_forest_partition.h
spenke91: Ok then I try to free some time this week to finalize it 👍

test/t8_forest/t8_gtest_forest_commit.cxx
spenke91: Interesting idea, thanks for sharing! I thought about maybe having some simple weight function that is for example 1 for the first half of the elements and 2 for the other, or that is maybe just equal to the element number. I'm optimistic to find something useful here 👍

dutkalex:
Hi @spenke91!
I found some time to come back to this PR and make the changes you requested (i.e. rollback the breaking API change in favor of an additional function to set the weights callback, and rollback the test hack). I also rebased the PR on the head of main, and fixed the conflicts with the custom offsets feature that recently got merged. IIRC, you implemented this feature, so please let me know if you have questions/comments/complaints on the refactoring.
Btw, feel free to ignore this message. I hope you're having a good time during these winter holidays.
I wish you - and the t8code team - all the best for the new year!

spenke91:
Hi @dutkalex, I just updated the PR. Most of the changes we made were pretty much the same. I hope you don't mind that I do not use the `DeferToScopeExit`: While I like the idea, it felt like an overkill to me and I instead packed the weighted partitioning in an `else` statement. Also, I added the flag `set_weighted_partitioning` which I agree is not required, but in line with how t8code handles other flags and options -- although it is a somewhat ugly solution... Of course I am open to discuss!
The only really new thing I added is the test case.

dutkalex:
> I hope you don't mind that I do not use the `DeferToScopeExit`: While I like the idea, it felt like an overkill to me and I instead packed the weighted partitioning in an else statement

These things are matter of a personal taste: I like to avoid nested if blocks and prefer early returns instead, but your solution is arguably just as readable and maintainable, so LGTM.

>  Also, I added the flag `set_weighted_partitioning` which I agree is not required, but in line with how t8code handles other flags and options -- although it is a somewhat ugly solution...

I would argue that it is redundant and also an opportunity for inconsistency (and therefore bugs and added maintenance work), and that I don't really see the value in adding this. However, at the end of the day this is your call since you guys are the ones which will have to maintain the code. If you really want to go down that road, I would strongly advise to add a check for `nullptr` before calling the weight function.

> The only really new thing I added is the test case.

Perfect 😊

spenke91:
> I would argue that it is redundant and also an opportunity for inconsistency (and therefore bugs and added maintenance work), and that I don't really see the value in adding this. However, at the end of the day this is your call since you guys are the ones which will have to maintain the code. If you really want to go down that road, I would strongly advise to add a check for nullptr before calling the weight function.

Frankly, I agree with you and the check for the nullptr is a good point! I will ask one of the more experienced t8code developers for an oppinion on this -- and then either change it or at least add the nulltptr check.

spenke91:
@dutkalex I gathered some opinions from the other developers and got rid of the extra flag as you suggested 👍

dutkalex:
> @dutkalex I gathered some opinions from the other developers and got rid of the extra flag as you suggested 👍

Ok, perfect then 😊

src/t8_forest/t8_forest_general.h
```diff
@@ -324,6 +331,15 @@ t8_forest_get_user_function (const t8_forest_t forest);
 void
 t8_forest_set_partition (t8_forest_t forest, const t8_forest_t set_from, int set_for_coarsening);
 
+/** Set a user-defined weight function to guide the partitioning.
+ * \param [in, out] forest  The forest.
+ * \param [in]      weight_callback A callback function defining element weights for the partitioning.
+ * \pre \a weight_callback must be free of side-effects, the behavior is undefined otherwise
```
Davknapp: What do you mean by side-effects?

test/t8_forest/t8_gtest_weighted_partitioning.cxx
```diff
@@ -0,0 +1,232 @@
+/*
+  This file is part of t8code.
+  t8code is a C library to manage a collection (a forest) of multiple
+  connected adaptive space-trees of general element classes in parallel.
+
+  Copyright (C) 2025 the developers
```
Davknapp:
```suggestion
  Copyright (C) 2026 the developers
```

test/t8_forest/t8_gtest_weighted_partitioning.cxx
```diff
@@ -0,0 +1,232 @@
...
+{
+
+  // -------------------------------------------
+  // ----- (1.) Create uniform base forest -----
+  // -------------------------------------------
+
+  t8_debugf ("Create uniform base forest.\n");
+
+  // Initial uniform refinement level
+  const int level = 2;
```
Davknapp, line 114: What is the current run-time of the test? Maybe we can allow a higher-level for the production-mode.

spenke91: In the github workflow, it took `3.6s` for Debug and `2.2s` in Release 
You mean setting a higher level depending on the test level?
To be honest, I don't think it adds any value to the test case but simply increases its workfload. 🤔 
We simply partition the forest's leaf elements, so the depth of the tree does not really change anything. In principle, level 1 would already be enough I think, but I set level 2 because it reduces the number of tested forests with very few elements.

src/t8_forest/t8_forest_general.h
spenke91: It should not change the forest for example

dutkalex: Yes, and it should not modify some global state such that calling twice the function on the same element could give different results also.

test/t8_forest/t8_gtest_weighted_partitioning.cxx
spenke91: FYI: Tested locally, going to level 3 roughly doubles the runtime, level 4 brought already an increase of roughly a factor of 14.

PR #1895, opened 2025-10-15
Improvement: Custom competence test for mesh interface

### LINKED ISSUE #1896: Improvement: Custom competence test for mesh interface

lenaploetzke:
Closes #1896

The unstructured mesh interface allows the use of custom competences. A simple test is added for this. 
Moreover there was a problem with the access specifier such that the caches were not actually used (because private members of course are not inherited). I need the public dummy functions because protected members cannot be used for the functionality to check if a cached version exists in the competences. 
Lastly, the test is now faster because the get_centroid function is called less. 

#### General
#### Tests

#### Scripts and Wiki
#### License

### DIFF
--- a/src/t8_unstructured_mesh/t8_unstructured_element.hxx
+++ b/src/t8_unstructured_mesh/t8_unstructured_element.hxx
@@ -71,13 +71,13 @@ class t8_unstructured_mesh_element: public TCompetence<t8_unstructured_mesh_elem
    */
   template <template <typename> class T>
   static constexpr bool
-  cache_vertex_coordinates_defined ()
+  vertex_cache_defined ()
   {
     return requires (T<SelfType>& competence) { competence.vertex_cache_filled (); };
   }
   /* This variable is true if any of the given competences \ref TCompetence implements 
   a function vertex_cache_filled */
-  static constexpr bool vertex_cache_exists = (false || ... || cache_vertex_coordinates_defined<TCompetence> ());
+  static constexpr bool vertex_cache_exists = (false || ... || vertex_cache_defined<TCompetence> ());
 
   /** Helper function to check if class T implements the function centroid_cache_filled.
    * \tparam T The competence to be checked.
@@ -85,13 +85,13 @@ class t8_unstructured_mesh_element: public TCompetence<t8_unstructured_mesh_elem
    */
   template <template <typename> class T>
   static constexpr bool
-  cache_centroid_defined ()
+  centroid_cache_defined ()
   {
     return requires (T<SelfType>& competence) { competence.centroid_cache_filled (); };
   }
   /* This variable is true if any of the given competences \ref TCompetence implements 
   a function centroid_cache_filled. */
-  static constexpr bool centroid_cache_exists = (false || ... || cache_centroid_defined<TCompetence> ());
+  static constexpr bool centroid_cache_exists = (false || ... || centroid_cache_defined<TCompetence> ());
 
  public:
   /**
--- a/test/t8_unstructured_mesh/t8_gtest_custom_competence.cxx
+++ b/test/t8_unstructured_mesh/t8_gtest_custom_competence.cxx
@@ -20,7 +20,10 @@ along with t8code; if not, write to the Free Software Foundation, Inc.,
 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
 */
 
-/** Checks that custom competences can be used in the unstructured mesh interface. */
+/**
+ * \file t8_gtest_custom_competence.cxx
+ * Checks that custom-defined competences can be used as template parameters for the unstructured mesh interface. 
+ */
 
 #include <gtest/gtest.h>
 #include <test/t8_gtest_schemes.hxx>
--- a/test/t8_unstructured_mesh/t8_gtest_unstructured_mesh.cxx
+++ b/test/t8_unstructured_mesh/t8_gtest_unstructured_mesh.cxx
@@ -20,7 +20,10 @@ along with t8code; if not, write to the Free Software Foundation, Inc.,
 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
 */
 
-/** Tests for the unstructured mesh class. */
+/**
+ * \file t8_gtest_unstructured_mesh.cxx
+ * Tests if the unstructured mesh class works as intended for different types of predefined template parameter classes. 
+ */
 
 #include <gtest/gtest.h>
 #include <test/t8_gtest_schemes.hxx>


src/t8_unstructured_mesh/t8_element_competences.hxx
```diff
@@ -53,28 +53,18 @@ along with t8code; if not, write to the Free Software Foundation, Inc.,
 template <typename TUnderlying>
 struct t8_cache_vertex_coordinates: t8_crtp_operator<TUnderlying, t8_cache_vertex_coordinates>
 {
- private:
+ public:
   /**
-   * Returns the previously cached vector with the vertex coordinates of the unstructured mesh element.
-   * \return Constant reference to the cached vector with the vertex coordinates.
+   * Dummy function for the element to check if a cache for the vertex coordinates exists.
+   * \return true as this competence provides a cache for the vertex_coordinates.
```
sandro-elsweijer: Imho this comment is ambiguous. This could either mean that the vertex cache is already filled or it means that this object has the ability to cache vertices.

test/t8_unstructured_mesh/t8_gtest_custom_competence.cxx
```diff
@@ -0,0 +1,101 @@
...
+  const t8_scheme *scheme = t8_scheme_new_default ();
+  t8_forest_t forest = t8_forest_new_uniform (cmesh, scheme, level, 0, sc_MPI_COMM_WORLD);
+  ASSERT_EQ (true, t8_forest_is_committed (forest));
+
+  // Check mesh with custom defined competence.
+  t8_unstructured_mesh<t8_unstructured_mesh_element<dummy_get_level>> unstructured_mesh
+    = t8_unstructured_mesh<t8_unstructured_mesh_element<dummy_get_level>> (forest);
+
+  for (auto it = unstructured_mesh.begin (); it != unstructured_mesh.end (); ++it) {
+    EXPECT_EQ (it->get_level (), it->get_level_dummy ());
```
sandro-elsweijer, line 84: Since get level and get level dummy are basically the same function, I would also check if the output is correct.
```suggestion
    EXPECT_EQ (it->get_level (), it->get_level_dummy ());
    EXPECT_EQ (level, it->get_level_dummy ())
```

test/t8_unstructured_mesh/t8_gtest_custom_competence.cxx
```diff
@@ -0,0 +1,101 @@
...
+  {
+    return 1;
+  }
+};
+
+/** This tests checks that custom defined competences can be used for the unstructured mesh class and that we can use the functionality defined in the competence. Also checks that we can use more than one custom competence and that predefined competences can be additionally used. */
+TEST (t8_gtest_custom_competence, custom_competence)
+{
+  // Define forest to construct unstructured mesh.
+  int level = 3;
```
sandro-elsweijer: I do not think that this test needs to iterate over more than 4000 elements
```suggestion
  const int level = 1;
```

test/t8_unstructured_mesh/t8_gtest_unstructured_mesh.cxx
```diff
@@ -68,15 +68,15 @@ TEST_P (t8_unstructured_mesh_test, test_iterator)
...
+    auto centroid = it->get_centroid ();
+    for (int coord = 0; coord < 3; ++coord) {
+      EXPECT_GE (1, centroid[coord]);
+      EXPECT_LE (0, centroid[coord]);
     }
     // Test dereference operator.
     auto vertex_coordinates = (*it).get_vertex_coordinates ();
     for (int ivertex = 0; ivertex < (int) vertex_coordinates.size (); ++ivertex) {
-      for (int coord = 0; coord < T8_ECLASS_MAX_DIM; ++coord) {
+      for (int coord = 0; coord < 3; ++coord) {
```
sandro-elsweijer: Why these changes?

sandro-elsweijer: By that I mean why do we not check the level anymore and why do we not use `T8_ECLASS_MAX_DIM`?

sandro-elsweijer: Furthermore, I also thin this test does not need to go through all level up to level 3 and through all schemes. You are testing the interface here, not the actual functionality of t8code

lenaploetzke: 1.: I dont use T8_ECLASS_MAX_DIM anymore because the centroid is now returned as `t8_3d_vec` such that we would have a problem if T8_ECLASS_MAX_DIM is changed in the future but in t8_3d_vec, the 3 is hard coded 
2.: Previously, i called get_centroid() 3*2 times, which was not necessary, now the variable centroid is used instead.
3. Yes you are righty, maybe only level 2? Do u think one scheme is maybe also enough?

src/t8_unstructured_mesh/t8_element_competences.hxx
lenaploetzke: You are completely right. Maybe it is a good idea to even rewrite the function to check if the cache has been set, so cache_vertec_coordinates_filled ? Because then we do not have a dummy function but a meaningful function and this function could even be used in the tests (and of course in the element class). What do you think?

test/t8_unstructured_mesh/t8_gtest_unstructured_mesh.cxx
sandro-elsweijer: 1. then why not `for (const auto& coord: centroid)` or `for (auto iter = centroid.begin() ...` ?
3. yes and yes :)

src/t8_unstructured_mesh/t8_element_competences.hxx
sandro-elsweijer: But maybe one also wants to know it the element caches in general. So maybe a function like `cache_exists()` and `cache_filled()`?

lenaploetzke: Hm should i then also provide a cache_exists function that is false if the cache does not exists in the element?

test/t8_unstructured_mesh/t8_gtest_unstructured_mesh.cxx
lenaploetzke: yeah you are righty 
i use the default scheme now but still all eclasses such that it is tested that the interface is fine for all eclasses

src/t8_unstructured_mesh/t8_element_competences.hxx
sandro-elsweijer: I would say it should always exist and just return a bool

src/t8_unstructured_mesh/t8_unstructured_element.hxx
```diff
@@ -65,34 +65,33 @@ class t8_unstructured_mesh_element: public TCompetence<t8_unstructured_mesh_elem
...
   // --- Variables to check which functionality is defined in TCompetence. ---
-  /** Helper function to check if class T implements the function get_vertex_coordinates_cached.
+  /** Helper function to check if class T implements the function vertex_cache_filled.
    * \tparam T The competence to be checked.
    * \return true if T implements the function, false if not.
    */
   template <template <typename> class T>
   static constexpr bool
-  has_get_vertex_coordinates_cached ()
+  cache_vertex_coordinates_defined ()
```
sandro-elsweijer: Can you please use the same order in the function names. So either vertex_cache or cache_vertex, but not both

test/t8_unstructured_mesh/t8_gtest_custom_competence.cxx
```diff
@@ -0,0 +1,104 @@
...
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with t8code; if not, write to the Free Software Foundation, Inc.,
+51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+*/
+
+/** Checks that custom competences can be used in the unstructured mesh interface. */
```
sandro-elsweijer: This is not a proper file documentation

PR #1914, opened 2025-10-22
Feature: Interface as cmake library

### LINKED ISSUE #1916: Feature: Inferface as cmake library

lenaploetzke:
Closes #1916

The unstructured mesh interface is now located in the api/ folder and is its own cmake library. I also tested it with one dummy repo. 

#### General
#### Tests

#### Scripts and Wiki
#### License

### DIFF
--- a/api/t8_unstructured_mesh/CMakeLists.txt
+++ b/api/t8_unstructured_mesh/CMakeLists.txt
@@ -19,7 +19,7 @@
 #  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
 
 # Define own library for mesh interface.
-if ( ${T8CODE_BUILD_AS_SHARED_LIBRARY} )
+if ( T8CODE_BUILD_AS_SHARED_LIBRARY )
     add_library( T8_MESH_INTERFACE SHARED )
     set_target_properties( T8_MESH_INTERFACE PROPERTIES POSITION_INDEPENDENT_CODE ON )
 else()
@@ -30,17 +30,24 @@ add_library( T8CODE::T8_MESH_INTERFACE ALIAS T8_MESH_INTERFACE )
 set_target_properties(T8_MESH_INTERFACE PROPERTIES OUTPUT_NAME t8_mesh_interface)
 
 target_include_directories( T8_MESH_INTERFACE PUBLIC
-  $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/..>
-  $<INSTALL_INTERFACE:include/t8_unstructured_mesh>
+  $<BUILD_INTERFACE:${CMAKE_CURRENT_LIST_DIR}/..>
+  $<INSTALL_INTERFACE:${CMAKE_INSTALL_PREFIX}/include/t8_unstructured_mesh>
 ) 
 
 target_sources(T8_MESH_INTERFACE PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/t8_unstructured_mesh.cxx)
-install( DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR} TYPE INCLUDE FILES_MATCHING PATTERN "*.hxx" )
+install( DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
+         DESTINATION include/
+         FILES_MATCHING PATTERN "*.hxx"
+)
 
 target_link_libraries(T8_MESH_INTERFACE PUBLIC T8)
 
-install( TARGETS T8_MESH_INTERFACE LIBRARY )
-install( TARGETS T8_MESH_INTERFACE EXPORT ${PROJECT_NAME}-targets )
+install( TARGETS T8_MESH_INTERFACE
+         EXPORT ${PROJECT_NAME}-targets
+         ARCHIVE DESTINATION lib
+         LIBRARY DESTINATION lib
+         RUNTIME DESTINATION bin
+)
 
 if( T8CODE_EXPORT_COMPILE_COMMANDS )
     set_target_properties( T8_MESH_INTERFACE PROPERTIES EXPORT_COMPILE_COMMANDS ON )


api/t8_unstructured_mesh/CMakeLists.txt
```diff
@@ -0,0 +1,47 @@
...
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#  GNU General Public License for more details.
+#
+#  You should have received a copy of the GNU General Public License
+#  along with t8code; if not, write to the Free Software Foundation, Inc.,
+#  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+
+# Define own library for mesh interface.
+if ( ${T8CODE_BUILD_AS_SHARED_LIBRARY} )
```
spenke91: I know it is the same in `src/CMakeLists.txt`, but I think it we should omit the dollar and the brackets, like this:
```suggestion
if (T8CODE_BUILD_AS_SHARED_LIBRARY)
```
If I remember correctly, variables in an `if` condition should not use this syntax, because they will be dereferenced before evaluating the conditon, which can lead to quite weird behaviour. See for example https://stackoverflow.com/questions/25809332/when-should-i-wrap-variables-with-in-cmake

api/t8_unstructured_mesh/CMakeLists.txt
```diff
@@ -0,0 +1,47 @@
...
+add_library( T8CODE::T8_MESH_INTERFACE ALIAS T8_MESH_INTERFACE )
+set_target_properties(T8_MESH_INTERFACE PROPERTIES OUTPUT_NAME t8_mesh_interface)
+
+target_include_directories( T8_MESH_INTERFACE PUBLIC
+  $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/..>
+  $<INSTALL_INTERFACE:include/t8_unstructured_mesh>
+) 
+
+target_sources(T8_MESH_INTERFACE PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/t8_unstructured_mesh.cxx)
+install( DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR} TYPE INCLUDE FILES_MATCHING PATTERN "*.hxx" )
```
spenke91:
```suggestion
install(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/
        DESTINATION include/t8_unstructured_mesh
        FILES_MATCHING PATTERN "*.hxx")
```

api/t8_unstructured_mesh/CMakeLists.txt
```diff
@@ -0,0 +1,47 @@
...
+
+target_link_libraries(T8_MESH_INTERFACE PUBLIC T8)
+
+install( TARGETS T8_MESH_INTERFACE LIBRARY )
+install( TARGETS T8_MESH_INTERFACE EXPORT ${PROJECT_NAME}-targets )
+
+if( T8CODE_EXPORT_COMPILE_COMMANDS )
+    set_target_properties( T8_MESH_INTERFACE PROPERTIES EXPORT_COMPILE_COMMANDS ON )
+endif( T8CODE_EXPORT_COMPILE_COMMANDS )
```
spenke91, line 54: I think `CMAKE_EXPORT_COMPILE_COMMANDS ` is a global property?
```suggestion
if(T8CODE_EXPORT_COMPILE_COMMANDS)
    set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
endif(T8CODE_EXPORT_COMPILE_COMMANDS)
```

api/t8_unstructured_mesh/CMakeLists.txt
```diff
@@ -0,0 +1,47 @@
...
+else()
+    add_library( T8_MESH_INTERFACE STATIC )
+endif()
+
+add_library( T8CODE::T8_MESH_INTERFACE ALIAS T8_MESH_INTERFACE )
+set_target_properties(T8_MESH_INTERFACE PROPERTIES OUTPUT_NAME t8_mesh_interface)
+
+target_include_directories( T8_MESH_INTERFACE PUBLIC
+  $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/..>
+  $<INSTALL_INTERFACE:include/t8_unstructured_mesh>
```
spenke91: Don't we have to add the `CMAKE_INSTALL_PREFIX`? I guess something like this:
```suggestion
  $<INSTALL_INTERFACE:${CMAKE_INSTALL_PREFIX}/include/t8_unstructured_mesh>
```

api/t8_unstructured_mesh/CMakeLists.txt
```diff
@@ -0,0 +1,47 @@
...
+    set_target_properties( T8_MESH_INTERFACE PROPERTIES POSITION_INDEPENDENT_CODE ON )
+else()
+    add_library( T8_MESH_INTERFACE STATIC )
+endif()
+
+add_library( T8CODE::T8_MESH_INTERFACE ALIAS T8_MESH_INTERFACE )
+set_target_properties(T8_MESH_INTERFACE PROPERTIES OUTPUT_NAME t8_mesh_interface)
+
+target_include_directories( T8_MESH_INTERFACE PUBLIC
+  $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/..>
```
spenke91: Is there a reason for not using `CMAKE_CURRENT_LIST_DIR` as in the other t8code `CMakeLists.txt` files? so
```suggestion
  $<BUILD_INTERFACE:${CMAKE_CURRENT_LIST_DIR}>
```
?
Here's some info on the difference: https://stackoverflow.com/questions/15662497/difference-between-cmake-current-source-dir-and-cmake-current-list-dir

api/t8_unstructured_mesh/CMakeLists.txt
```diff
@@ -0,0 +1,47 @@
...
+  $<INSTALL_INTERFACE:include/t8_unstructured_mesh>
+) 
+
+target_sources(T8_MESH_INTERFACE PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/t8_unstructured_mesh.cxx)
+install( DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR} TYPE INCLUDE FILES_MATCHING PATTERN "*.hxx" )
+
+target_link_libraries(T8_MESH_INTERFACE PUBLIC T8)
+
+install( TARGETS T8_MESH_INTERFACE LIBRARY )
+install( TARGETS T8_MESH_INTERFACE EXPORT ${PROJECT_NAME}-targets )
```
spenke91: Again, I see it is the same in `src/CMakeLists.txt`, but I think this might be cleaner?
```suggestion
install(TARGETS T8_MESH_INTERFACE
    EXPORT ${PROJECT_NAME}-targets
    ARCHIVE DESTINATION lib
    LIBRARY DESTINATION lib
    RUNTIME DESTINATION bin
)
```
But I may well be wrong. This might be worth a look: https://cmake.org/cmake/help/book/mastering-cmake/chapter/Install.html

api/t8_unstructured_mesh/CMakeLists.txt
lenaploetzke: Very nice comment, i wasn't aware of this! Do you think it is worth changing it also in `src/CMakeLists.txt`?

api/t8_unstructured_mesh/CMakeLists.txt
lenaploetzke: In all other CMakeLists.txt we have similar set_target_properties for the target specified. I don't know if i want to do it differently

api/t8_unstructured_mesh/CMakeLists.txt
lenaploetzke: Thought SOURCE_DIR is the better as this always points to the current folder but after rereading i might be wrong and i corrected it

api/t8_unstructured_mesh/CMakeLists.txt
lenaploetzke: yes youre right,this is much better!

lenaploetzke: Also here: not sure if i should adapt it in the src directory?

api/t8_unstructured_mesh/CMakeLists.txt
spenke91: I guess so 🤔  But I would make it an own issue / PR and not add it here :-)

api/t8_unstructured_mesh/CMakeLists.txt
spenke91: As above, maybe we can collect it in one issue and discuss it in a dev meeting - in case we got something wrong here or there are other opinions on it

api/t8_unstructured_mesh/CMakeLists.txt
spenke91: Ok then let's keep it. Also, I may have been wrong and this one exists too: https://cmake.org/cmake/help/latest/prop_tgt/EXPORT_COMPILE_COMMANDS.html

api/t8_unstructured_mesh/CMakeLists.txt
lenaploetzke: Yes, opened issue #1946

api/t8_unstructured_mesh/CMakeLists.txt
lenaploetzke: See #1947

PR #1944, opened 2025-11-05
Feature: face neighbors and ghosts for handle

### LINKED ISSUE #1945: Feature: face neighbors and ghosts for handle

lenaploetzke:
Closes #1945

The mesh handle is now able to get the face neighbors of a mesh element. For this, ghost functionality is needed. 

#### General
#### Tests

#### Scripts and Wiki
#### License

### DIFF
--- a/test/mesh_handle/t8_gtest_ghost.cxx
+++ b/test/mesh_handle/t8_gtest_ghost.cxx
@@ -51,13 +51,7 @@ class t8_mesh_ghost_test: public testing::TestWithParam<std::tuple<t8_eclass_t, 
     t8_cmesh_t cmesh = t8_cmesh_new_hypercube (eclass, sc_MPI_COMM_WORLD, 0, 1, 0);
     forest = t8_forest_new_uniform (cmesh, scheme, level, 1, sc_MPI_COMM_WORLD);
   }
-  void
-  TearDown () override
-  {
-    if (forest->rc.refcount > 0) {
-      t8_forest_unref (&forest);
-    }
-  }
+
   t8_forest_t forest;
   int level;
 };
--- a/test/mesh_handle/t8_gtest_cache_competence.cxx
+++ b/test/mesh_handle/t8_gtest_cache_competence.cxx
@@ -79,14 +79,6 @@ class t8_gtest_cache_competence: public testing::Test {
     forest = t8_forest_new_uniform (cmesh, scheme, level, 0, sc_MPI_COMM_WORLD);
   }
 
-  void
-  TearDown () override
-  {
-    if (forest->rc.refcount > 0) {
-      t8_forest_unref (&forest);
-    }
-  }
-
   t8_forest_t forest;
   int level;
 };
--- a/test/mesh_handle/t8_gtest_mesh_handle.cxx
+++ b/test/mesh_handle/t8_gtest_mesh_handle.cxx
@@ -48,13 +48,6 @@ class t8_mesh_handle_test: public testing::TestWithParam<std::tuple<t8_eclass_t,
     t8_cmesh_t cmesh = t8_cmesh_new_hypercube (eclass, sc_MPI_COMM_WORLD, 0, 1, 0);
     forest = t8_forest_new_uniform (cmesh, scheme, level, 0, sc_MPI_COMM_WORLD);
   }
-  void
-  TearDown () override
-  {
-    if (forest->rc.refcount > 0) {
-      t8_forest_unref (&forest);
-    }
-  }
 
   t8_forest_t forest;
   int level;


mesh_handle/abstract_element.hxx
```diff
@@ -20,12 +20,12 @@ along with t8code; if not, write to the Free Software Foundation, Inc.,
 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
 */
 
-/** \file element.hxx
- * Definition of the elements used in the \ref t8_mesh_handle::mesh class.
+/** \file abstract_element.hxx
```
sandro-elsweijer: This class is not abstract since it does not have a virtual function. How about element_base?

mesh_handle/element.hxx
```diff
@@ -62,10 +62,25 @@ class mesh;
...
+class abstract_element: public TCompetence<abstract_element<TCompetence...>>... {
+ protected:
+  using SelfType
+    = abstract_element<TCompetence...>;    /**< Type of the current class with all template parameters specified. */
+  using mesh_class = mesh<TCompetence...>; /**< Type of the mesh class the used. */
+  friend mesh_class; /**< Define mesh_class as friend to be able to access e.g. the constructor. */
+
+  /**
+   * Protected constructor for an element of a mesh. 
+   * This constructor of the abstract class should only be called by child classes.
```
sandro-elsweijer:
```suggestion
   * This constructor of the base class should only be called by child classes.
```

mesh_handle/abstract_element.hxx
sandro-elsweijer: Nevermind, it has one virtual function. But I still find base a better description

mesh_handle/element.hxx
```diff
@@ -205,9 +261,75 @@ class element: public TCompetence<element<TCompetence...>>... {
     return coordinates;
   }
 
+  /** Getter for the face neighbors of the mesh element at given face.
```
sandro-elsweijer:
```suggestion
  /** Getter for the face neighbors of the mesh element at a given face.
```

mesh_handle/element.hxx
```diff
@@ -205,9 +261,75 @@ class element: public TCompetence<element<TCompetence...>>... {
...
+   * For ghost elements, the functionality to calculate face neighbors is currently not provided.
```
sandro-elsweijer: This is a bit tricky. In practice, this function can only give ghosts which were communicated. Currently, we only communicate face neighbors as ghosts. Theoretically, we can communicate more elements (like vertex neighbors or second degree neighbors) and then this function should be able to return these as well. This all depends on the behavior of @holke's leaf face neighbor function and the used ghost algorithm.
To be agnostic about this we would need some sort of class, which defines what a neighbor is. This definition should then be used by the ghost and by some sort of `t8_forest_element_get_neighbor`. But this is just a wish for the future.

Second thing is, that imho you do not need to check the user input here. `t8_forest_leaf_face_neighbors` already checks if the element is local. You just need to pass the user input to the function. Otherwise, we have this check at several locations and then we might forget to remove this when this limitation is lifted.

mesh_handle/element.hxx
```diff
@@ -205,9 +261,75 @@ class element: public TCompetence<element<TCompetence...>>... {
...
+   * This function uses the cached version defined in TCompetence if available and calculates if not.
+   * \param [in]  face          The index of the face across which the face neighbors are searched.
+   * \param [out] dual_faces    On output the face id's of the neighboring elements' faces.
```
sandro-elsweijer, line 268: In the future we probably also need some sort of orientation for our DG users. But for now this is allright

mesh_handle/element.hxx
```diff
@@ -205,9 +261,75 @@ class element: public TCompetence<element<TCompetence...>>... {
...
+   * \return Vector of length num_neighbors with pointers to the elements neighboring at the given face.
+   */
+  std::vector<const SelfType*>
+  get_face_neighbors (int face, std::vector<int>* dual_faces = nullptr) const
```
sandro-elsweijer:
```suggestion
  get_face_neighbors (int face, std::optional<std::reference_wrapper<std::vector<int>>> dual_faces= std::nullopt) const
```

mesh_handle/element.hxx
```diff
@@ -205,9 +261,75 @@ class element: public TCompetence<element<TCompetence...>>... {
...
+  {
+    SC_CHECK_ABORT (!m_is_ghost_element, "get_face_neighbors is not implemented for ghost elements.\n");
+    if constexpr (neighbor_cache_exists) {
+      if (this->neighbor_cache_filled (face)) {
+        if (dual_faces) {
+          *dual_faces = this->m_dual_faces[face];
```
sandro-elsweijer:
```suggestion
          dual_faces->get() = this->m_dual_faces[face];
```

mesh_handle/element.hxx
sandro-elsweijer: https://godbolt.org/z/fn4zhq11x

mesh_handle/element.hxx
```diff
@@ -34,25 +34,25 @@ along with t8code; if not, write to the Free Software Foundation, Inc.,
 #include <t8_forest/t8_forest_geometrical.h>
 #include <t8_schemes/t8_scheme.hxx>
 #include <t8_types/t8_vec.hxx>
-#include <array>
 #include <vector>
```
sandro-elsweijer, line 37:
```suggestion
#include <vector>
#include <optional>
```

mesh_handle/element.hxx
```diff
@@ -227,29 +349,55 @@ class element: public TCompetence<element<TCompetence...>>... {
...
+  get_flat_element_id () const
+  {
+    if (m_is_ghost_element) {
+      return m_mesh->get_num_local_elements ()
+             + t8_forest_ghost_get_tree_element_offset (m_mesh->m_forest,
+                                                        m_tree_id - t8_forest_get_num_local_trees (m_mesh->m_forest))
+             + m_element_id;
+    }
+    return t8_forest_get_tree_element_offset (m_mesh->m_forest, m_tree_id) + m_element_id;
+  }
```
sandro-elsweijer: I would implement this a s an optional competence. It is only needed to access the elements data and if you simplify the data access then this is never needed and only confuses the user

mesh_handle/element.hxx
```diff
@@ -227,29 +349,55 @@ class element: public TCompetence<element<TCompetence...>>... {
...
-   * Getter for the leaf element of the mesh element.
+   * Getter for the leaf element of the element.
    * \return The leaf element.
    */
   const t8_element_t*
   get_element () const
   {
-    return t8_forest_get_leaf_element_in_tree (m_mesh->m_forest, m_tree_id, m_element_id);
+    return m_element;
   }
```
sandro-elsweijer, line 403: Same here. Ideally, the user does not need this. Furthermore, it is confusing that an element has a function named `get_element`. If the element was called `element_handle`, then this would make sense, since it is a handle to the underlying element

mesh_handle/mesh.hxx
```diff
@@ -70,6 +69,26 @@ class mesh {
...
 
+  /**
+   * Getter for the number of local ghost elements.
+   * \return Number of local ghost elements in the mesh.
+   */
+  t8_locidx_t
+  get_num_local_ghosts () const
+  {
+    return t8_forest_get_num_ghosts (m_forest);
+  }
```
sandro-elsweijer, line 90: ghosts are by definition remote, so they cannot be local.
```suggestion
  /**
   * Getter for the number of ghost elements.
   * \return Number of ghost elements in the mesh.
   */
  t8_locidx_t
  get_num_ghosts () const
  {
    return t8_forest_get_num_ghosts (m_forest);
  }
```

mesh_handle/mesh.hxx
```diff
@@ -70,6 +69,26 @@ class mesh {
...
+
+  /** 
+   * Getter for the dimension the mesh handle lives in.
```
sandro-elsweijer:
```suggestion
   * Getter for the dimension of the mesh.
```

mesh_handle/mesh.hxx
```diff
@@ -119,6 +147,7 @@ class mesh {
   set_forest (t8_forest_t input_forest)
   {
     m_forest = input_forest;
+    T8_ASSERT (t8_forest_is_committed (m_forest));
```
sandro-elsweijer:
```suggestion
    T8_ASSERT (t8_forest_is_committed (input_forest));
    m_forest = input_forest;
```

test/mesh_handle/t8_gtest_ghost.cxx
```diff
@@ -0,0 +1,229 @@
...
+      EXPECT_LE (0, coordinate);
+    }
+    auto vertex_coordinates = (*it).get_vertex_coordinates ();
+    for (int ivertex = 0; ivertex < (int) vertex_coordinates.size (); ++ivertex) {
+      for (const auto& coordinate : vertex_coordinates[ivertex]) {
+        EXPECT_GE (1, coordinate);
+        EXPECT_LE (0, coordinate);
+      }
+    }
+  }
```
sandro-elsweijer: Istnt this checked in another test? Tests should test one specific functionality.

mesh_handle/element.hxx
lenaploetzke: I will keep it in mind

mesh_handle/element.hxx
lenaploetzke: Hm i also have a use case in the adapt routine. (and as you said in the element data). I could do it manually, but I somehow like the convenience. Also maybe cool for user defined competences

mesh_handle/element.hxx
lenaploetzke: But this function is in the private area so the user does not see it.

mesh_handle/mesh.hxx
lenaploetzke: Youre right haha

mesh_handle/mesh.hxx
lenaploetzke: thanks!

test/mesh_handle/t8_gtest_ghost.cxx
lenaploetzke: The thing above? In other tests we do not have ghosts and my idea was to also ensure that non-ghost elements still work as expected. But you are right maybe this is unnecessary.

mesh_handle/element.hxx
lenaploetzke: t8_forest_leaf_face_neighbors only checks for local elements in debug mode right? Because of the `T8_ASSERT`. But i get your point.  I would really like to have an error message in all build types. Do i miss another check in `t8_forest_leaf_face_neighbors `? But i will delete the comment.

mesh_handle/element.hxx
lenaploetzke: I did not really dislike the nullptr version but maybe this is more explicit. Works out. :)

mesh_handle/element.hxx
```diff
@@ -91,23 +123,26 @@ class element: public TCompetence<element<TCompetence...>>... {
...
-   * \param [in] mesh           Pointer to the mesh the element should belong to.
-   * \param [in] tree_id        The tree id of the element in the forest defining the mesh.
-   * \param [in] element_id     The element id of the element in the forest defining the mesh.
+  /** Helper function to check if class T implements the function neighbor_cache_filled.
+   * \tparam T The competence to be checked.
+   * \return true if T implements the function, false if not.
    */
-  element (mesh<SelfType>* mesh, t8_locidx_t tree_id, t8_locidx_t element_id)
-    : m_mesh (mesh), m_tree_id (tree_id), m_element_id (element_id)
+  template <template <typename> class T>
```
sandro-elsweijer: We require more telling template parameter names than `T`

mesh_handle/element.hxx
```diff
@@ -77,8 +109,8 @@ class element: public TCompetence<element<TCompetence...>>... {
   {
```
sandro-elsweijer, line 109: We require more telling template parameter names than `T`

lenaploetzke: Renamed T to TCompetence and the variadic template parameter from TCompetence to TCompetences, as I found this more explanatory.

PR #2005, opened 2025-11-28
Feature: Add element and user data to mesh handle

### LINKED ISSUE #2006: Feature: Add element and user data to mesh handle

lenaploetzke:
Closes #2006

The mesh handle class is now able to handle element and user data. Template parameters are used to specify the classes (instead of the void* used in the forest for the user data). To specify them as template parameters, i needed to somehow pack the Competences into one template parameter (see parameter_pack.hxx) (because i wanted defaults for the element and user data classes but a parameter pack also always has to be the last template parameter). I think its a feature to have them all in one template parameter for the user and makes it easier to use (although slightly more complicated as a developer). Additionally, i can now provide predefined parameter packs e.g. cache_competences, where all cache competences are included.
The element data is stored in a vector in the mesh (and not in the elements). This is easier for the ghost exchange.
Additionally, the volume of the elements can now be accessed and cached. A tutorial used this as element data which is why i got inspired and also implemented the functionality for the handle. The calculation can be somehow complicated which is why i also provide a cached version.

#### General
#### Tests

#### Scripts and Wiki
#### License

### DIFF
--- a/mesh_handle/element.hxx
+++ b/mesh_handle/element.hxx
@@ -427,8 +427,8 @@ class element: public TCompetences<element<mesh_class, TCompetences...>>... {
   get_element_data () const
   {
     const t8_locidx_t handle_id = get_element_handle_id ();
-    T8_ASSERTF (handle_id < m_mesh->m_element_data.size (), "Element data not set.\n");
-    return m_mesh->m_element_data[get_element_handle_id ()];
+    T8_ASSERTF (static_cast<size_t> (handle_id) < m_mesh->m_element_data.size (), "Element data not set.\n");
+    return m_mesh->m_element_data[handle_id];
   }
 
   /** 
--- a/mesh_handle/mesh.hxx
+++ b/mesh_handle/mesh.hxx
@@ -233,7 +233,7 @@ class mesh {
   void
   set_element_data (std::vector<ElementDataType> element_data)
   {
-    T8_ASSERT (element_data.size () == get_num_local_elements ());
+    T8_ASSERT (element_data.size () == static_cast<size_t> (get_num_local_elements ()));
     m_element_data = std::move (element_data);
     m_element_data.reserve (get_num_local_elements () + get_num_ghosts ());
     m_element_data.resize (get_num_local_elements ());


mesh_handle/element.hxx
```diff
@@ -390,6 +412,34 @@ class element: public TCompetence<element<TCompetence...>>... {
     return m_is_ghost_element;
   }
 
+  // --- Getter and setter for element data. ---
+  /** Getter for the element data.
+   * For ghost elements ensure that \ref mesh::exchange_ghost_data is called on each process first.
+   * Element data for non-ghost elements can be accessed (if set) directly.
+   * \return Element data with data of Type mesh_class::ElementDataType.
+   */
+  template <typename E = typename mesh_class::ElementDataType, typename = std::enable_if_t<!std::is_void<E>::value>>
```
sandro-elsweijer: Please use proper template parameter names

mesh_handle/element.hxx
```diff
@@ -390,6 +412,34 @@ class element: public TCompetence<element<TCompetence...>>... {
...
+  {
+    return m_mesh->m_element_data[get_element_handle_id ()];
+  }
+
+  /** 
+   * Set the element data for the element. 
+   * \note You can only set element data for non-ghost elements.
+   * \param [in] element_data The element data to be set.
+   */
+  template <typename E = typename mesh_class::ElementDataType, typename = std::enable_if_t<!std::is_void<E>::value>>
```
sandro-elsweijer: Parameter name

mesh_handle/element.hxx
```diff
@@ -390,6 +412,34 @@ class element: public TCompetence<element<TCompetence...>>... {
...
+  void
+  set_element_data (E element_data)
+  {
+    SC_CHECK_ABORT (!m_is_ghost_element, "Element data cannot be set for ghost elements.\n");
+    // Resize for the case that no data vector has been set previously.
+    m_mesh->m_element_data.resize (m_mesh->get_num_local_elements () + m_mesh->get_num_ghosts ());
+    m_mesh->m_element_data[get_element_handle_id ()] = element_data;
+  }
```
sandro-elsweijer, line 449: IMHO we could also overload the dereference operator. In my head it makes total sense that an element handle stores data and that you can access its data by dereferencing it.

mesh_handle/element.hxx
```diff
@@ -390,6 +412,34 @@ class element: public TCompetence<element<TCompetence...>>... {
...
+   * For ghost elements ensure that \ref mesh::exchange_ghost_data is called on each process first.
+   * Element data for non-ghost elements can be accessed (if set) directly.
+   * \return Element data with data of Type mesh_class::ElementDataType.
+   */
+  template <typename E = typename mesh_class::ElementDataType, typename = std::enable_if_t<!std::is_void<E>::value>>
+  const E&
+  get_element_data () const
+  {
+    return m_mesh->m_element_data[get_element_handle_id ()];
+  }
```
sandro-elsweijer: Do we need a check if the element has data in the first place?

mesh_handle/element.hxx
sandro-elsweijer: Ah but then we cannot check if the user tries to set data for a ghost element

mesh_handle/element.hxx
```diff
@@ -390,6 +412,34 @@ class element: public TCompetence<element<TCompetence...>>... {
...
+   * \param [in] element_data The element data to be set.
+   */
+  template <typename E = typename mesh_class::ElementDataType, typename = std::enable_if_t<!std::is_void<E>::value>>
+  void
+  set_element_data (E element_data)
+  {
+    SC_CHECK_ABORT (!m_is_ghost_element, "Element data cannot be set for ghost elements.\n");
+    // Resize for the case that no data vector has been set previously.
+    m_mesh->m_element_data.resize (m_mesh->get_num_local_elements () + m_mesh->get_num_ghosts ());
+    m_mesh->m_element_data[get_element_handle_id ()] = element_data;
```
sandro-elsweijer:
```suggestion
    m_mesh->m_element_data[get_element_handle_id ()] = std::move(element_data);
```

sandro-elsweijer:
```suggestion
    m_mesh->m_element_data[get_element_handle_id ()] = std::move(element_data);
```

mesh_handle/mesh.hxx
```diff
@@ -40,21 +42,33 @@ namespace t8_mesh_handle
  * Wrapper for a forest that enables it to be handled as a simple mesh object.
  * \tparam TCompetence The competences you want to add to the default functionality of the mesh.
  *         \see element for more details on the choice of the template parameter.   
+ *         \note Please pack your competences using the \ref competence_pack class.
+ * \tparam TUserData The user data type you want to associate with the mesh. Use void (this is also the default) if you do not want to set user data.
+ * \tparam TElementData The element data type you want to use for each element of the mesh. 
+ *         Use void (this is also the default) if you do not want to set element data.
  */
-template <template <typename> class... TCompetence>
+template <typename TCompetencePack = competence_pack<>, typename TUserData = void, typename TElementData = void>
```
sandro-elsweijer: We save the user data in a void pointer and this often gets exploited in t8code. It is for example used to pass data to the adapt function. And I think it also gets used at other occasions. So the datatype of the user data changes regularly and the template locks this to a specific datatype. This is typesafe, but unfortunately breaks t8code. Maybe we can put the userdata into a `std::any`

sandro-elsweijer: The element data has to fulfill certain properties to be able to be sent via MPI. Most importantly it cannot contain pointers, since MPI would only send the pointers and not the data. Until now, we had to trust the user, but with this approach we can use a concept for that :)

```cpp
template<typename TType>
concept T8MPISafeType =
    std::is_trivially_copyable_v<TType> &&
    std::is_standard_layout_v<TType>;
```

```cpp
template <typename TCompetencePack = competence_pack<>, T8MPISafeType TElementData = void>
```

mesh_handle/mesh.hxx
```diff
@@ -104,18 +118,38 @@ class mesh {
...
    */
   mesh_const_iterator
-  end () const
+  cend () const
   {
     return m_elements.cend ();
   }
 
+  /**
+   * Non-const version of \ref cbegin.
```
sandro-elsweijer:
```suggestion
   * Returns an iterator to the first (local) mesh element.
```

mesh_handle/mesh.hxx
```diff
@@ -104,18 +118,38 @@ class mesh {
...
+   * \return Iterator to the first (local) mesh element.
+   */
+  mesh_iterator
+  begin ()
+  {
+    return m_elements.begin ();
+  }
+
+  /**
+   * Non-const version of \ref end.
```
sandro-elsweijer:
```suggestion
   * Returns an iterator to a mesh element following the last (local) element of the mesh.
```

mesh_handle/mesh.hxx
```diff
@@ -140,15 +185,92 @@ class mesh {
...
   void
   set_forest (t8_forest_t input_forest)
   {
     T8_ASSERT (t8_forest_is_committed (input_forest));
     m_forest = input_forest;
     update_elements ();
+    if constexpr (!std::is_void<TElementData>::value) {
+      t8_global_infof (
+        "The elements of the mesh handle have been updated. Please note that the element data is not interpolated "
+        "automatically. Use the function set_element_data() to provide new adapted element data.\n");
```
sandro-elsweijer: We also use the word interpolation for the interpolation of one elements data to the children. E.G. after adaptation of the mesh. I would therefore rephrase this:
```suggestion
      t8_global_infof (
        "The forest of the mesh handle has been updated. Please note that the element data in the mesh has to be updated accordingly. Use the function set_element_data() to provide element data fitting to the new forest.\n");
```

mesh_handle/mesh.hxx
```diff
@@ -140,15 +185,92 @@ class mesh {
...
 
   /** 
-  * Setter for the forest. 
-  * \param [in] input_forest The forest from which the mesh should be a wrapper. 
-  */
+   * Setter for the forest. 
+   * \param [in] input_forest The forest from which the mesh should be a wrapper. 
+   */
   void
   set_forest (t8_forest_t input_forest)
```
sandro-elsweijer: After thinking longer about this function it appears, that it is not that useful to the user and may introduce possibilities for misuse.
The mesh handle is templated after the element data. But if we put a new forest into the existing mesh handle the data might be in a different datatype.
Furthermore, as you show in your info message, the user can just update the forest without updating the element data. Maybe it would be better for the user to generate a new mesh handle entirely.

It will be beneficial for a workflow where the user commits the forest. Then, the internal forest has to be updated. But this will probably be done by an internal function and not the user. So maybe we can make this function private.

mesh_handle/mesh.hxx
```diff
@@ -140,15 +185,92 @@ class mesh {
...
+        "The elements of the mesh handle have been updated. Please note that the element data is not interpolated "
+        "automatically. Use the function set_element_data() to provide new adapted element data.\n");
+    }
+  }
+
+  /** 
+   * Set the user data of the mesh. This can i.e. be used to pass user defined arguments to the adapt routine.
+   * \param [in] data The user data of class TUserData. Data will never be touched by mesh handling routines.
+   */
+  template <typename U = TUserData, typename = std::enable_if_t<!std::is_void<U>::value>>
```
sandro-elsweijer: Please use proper template parameter names

sandro-elsweijer: Why do you not keep using the name `TUserData`?

mesh_handle/mesh.hxx
```diff
@@ -140,15 +185,92 @@ class mesh {
...
+  void
+  set_user_data (const U& data)
```
sandro-elsweijer: template parameter name

mesh_handle/mesh.hxx
```diff
@@ -140,15 +185,92 @@ class mesh {
...
+  {
+    t8_forest_set_user_data (m_forest, data);
+  }
+
+  /** 
+   * Get the user data of the mesh. 
+   * \return The user data previously set using \ref set_user_data.   
+   */
+  template <typename U = TUserData, typename = std::enable_if_t<!std::is_void<U>::value>>
```
sandro-elsweijer: template parameter name

mesh_handle/mesh.hxx
```diff
@@ -140,15 +185,92 @@ class mesh {
...
+  {
+    return *static_cast<const U*> (t8_forest_get_user_data (m_forest));
+  }
+
+  /** 
+   * Set the element data vector. The vector should have the length of num_local_elements.
+   * \param [in] element_data The element data vector to set with one entry of class TElementData 
+   *            for each local mesh element (excluding ghosts).
+   */
+  template <typename E = TElementData, typename = std::enable_if_t<!std::is_void<E>::value>>
```
sandro-elsweijer: template parameter name

mesh_handle/mesh.hxx
sandro-elsweijer:
```suggestion
template <typename TCompetencePack = competence_pack<>, typename TUserDataType = void, typename TElementDataType = void>
```

While reading the code I often confused the type of the data with the actual data and other things. This would make this clear

mesh_handle/mesh.hxx
```diff
@@ -140,15 +185,92 @@ class mesh {
...
+  void
+  set_element_data (const std::vector<E>& element_data)
+  {
+    T8_ASSERT (element_data.size () == get_num_local_elements ());
+    m_element_data.clear ();
+    m_element_data.resize (get_num_local_elements () + get_num_ghosts ());
+    std::copy (element_data.begin (), element_data.end (), m_element_data.begin ());
+  }
```
sandro-elsweijer, line 240: This is a little bit more efficient if the user does `mesh.set_element_data(std::move(vector))` since it does not copy anything. If the user does `mesh.set_element_data(vector)` there is still only one copy operation happening, just as in your code.

```suggestion
  void
  set_element_data (std::vector<E> element_data)
  {
    T8_ASSERT (element_data.size () == get_num_local_elements ());
    m_element_data = std::move(element_data);
    m_element_data.resize (get_num_local_elements () + get_num_ghosts ());
  }
```

sandro-elsweijer: What happens if the user sets the element data and then tries to access the data of a ghost element? We somehow have to prevent the user from accessing the ghost data if it was not filled. Maybe we need a flag for this or we do not resize here. Then the ghost data function can check if ghost values are present.

mesh_handle/mesh.hxx
```diff
@@ -140,15 +185,92 @@ class mesh {
...
+    sc_array* sc_array_wrapper;
+    T8_ASSERT (m_element_data.size () == get_num_local_elements () + get_num_ghosts ());
+    sc_array_wrapper = sc_array_new_data (m_element_data.data (), sizeof (TElementData),
+                                          get_num_local_elements () + get_num_ghosts ());
+
+    // Data exchange: entries with indices > num_local_elements will get overwritten.
+    t8_forest_ghost_exchange_data (m_forest, sc_array_wrapper);
+
+    sc_array_destroy (sc_array_wrapper);
+    return m_element_data;
```
sandro-elsweijer: Why does this function return any values?

test/mesh_handle/t8_gtest_cache_competence.cxx
```diff
@@ -89,20 +105,45 @@ class t8_gtest_cache_competence: public testing::Test {
...
  * checking that the functionality actually outputs this unrealistic value.
  */
 TEST_F (t8_gtest_cache_competence, cache_vertex_coordinates)
 {
-  using mesh_class = t8_mesh_handle::mesh<cache_vertex_coordinates_overwrite>;
-  using element_class = mesh_class::element_class;
-  mesh_class mesh = mesh_class (forest);
+  using mesh_class = t8_mesh_handle::mesh<t8_mesh_handle::competence_pack<cache_vertex_coordinates_overwrite>>;
+  using element_class = typename mesh_class::element_class;
+  const mesh_class mesh = mesh_class (forest);
```
sandro-elsweijer: Tell me that you learned Java in school without saying that you leaned Java in school.
```suggestion
  const mesh_class mesh (forest);
```

test/mesh_handle/t8_gtest_cache_competence.cxx
```diff
@@ -128,14 +169,13 @@ TEST_F (t8_gtest_cache_competence, cache_vertex_coordinates)
  */
 TEST_F (t8_gtest_cache_competence, cache_centroid)
 {
-  using mesh_class = t8_mesh_handle::mesh<cache_centroid_overwrite>;
+  using mesh_class = t8_mesh_handle::mesh<t8_mesh_handle::competence_pack<cache_centroid_overwrite>>;
   using element_class = mesh_class::element_class;
-  mesh_class mesh = mesh_class (forest);
-  EXPECT_FALSE (element_class::has_vertex_cache ());
+  const mesh_class mesh = mesh_class (forest);
```
sandro-elsweijer:
```suggestion
  const mesh_class mesh (forest);
```

test/mesh_handle/t8_gtest_compare_handle_to_forest.cxx
```diff
@@ -45,12 +46,12 @@ TEST (t8_gtest_compare_handle_to_forest, compare_handle_to_forest)
   const t8_scheme *init_scheme = t8_scheme_new_default ();
   t8_forest_t forest = t8_forest_new_uniform (cmesh, init_scheme, level, 0, sc_MPI_COMM_WORLD);
 
-  t8_mesh_handle::mesh<> mesh = t8_mesh_handle::mesh<> (forest);
+  const t8_mesh_handle::mesh<> mesh = t8_mesh_handle::mesh<> (forest);
```
sandro-elsweijer:
```suggestion
  const t8_mesh_handle::mesh<> mesh (forest);
```

test/mesh_handle/t8_gtest_custom_competence.cxx
```diff
@@ -89,18 +90,20 @@ TEST (t8_gtest_custom_competence, custom_competence)
   t8_forest_t forest = t8_forest_new_uniform (cmesh, scheme, level, 0, sc_MPI_COMM_WORLD);
 
   // Check mesh with custom defined competence.
-  t8_mesh_handle::mesh<dummy_get_level> mesh = t8_mesh_handle::mesh<dummy_get_level> (forest);
+  using mesh_class_custom = t8_mesh_handle::mesh<t8_mesh_handle::competence_pack<dummy_get_level>>;
+  const mesh_class_custom mesh = mesh_class_custom (forest);
```
sandro-elsweijer:
```suggestion
  const mesh_class_custom mesh (forest);
```

test/mesh_handle/t8_gtest_ghost.cxx
```diff
@@ -65,7 +66,7 @@ TEST_P (t8_mesh_ghost_test, check_ghosts)
 {
   t8_forest_ghost_print (forest);
 
-  t8_mesh_handle::mesh<> mesh = t8_mesh_handle::mesh<> (forest);
+  const t8_mesh_handle::mesh<> mesh = t8_mesh_handle::mesh<> (forest);
```
sandro-elsweijer:
```suggestion
  const t8_mesh_handle::mesh<> mesh (forest);
```

test/mesh_handle/t8_gtest_ghost.cxx
```diff
@@ -101,12 +102,12 @@ TEST_P (t8_mesh_ghost_test, compare_neighbors_to_forest)
 {
   ASSERT_TRUE (t8_forest_is_committed (forest));
 
-  t8_mesh_handle::mesh<> mesh = t8_mesh_handle::mesh<> (forest);
+  const t8_mesh_handle::mesh<> mesh = t8_mesh_handle::mesh<> (forest);
```
sandro-elsweijer:
```suggestion
  const t8_mesh_handle::mesh<> mesh (forest);
```

test/mesh_handle/t8_gtest_ghost.cxx
```diff
@@ -178,9 +179,9 @@ struct cache_neighbors_overwrite: public t8_mesh_handle::cache_neighbors<TUnderl
  */
 TEST_P (t8_mesh_ghost_test, cache_neighbors)
 {
-  using mesh_class = t8_mesh_handle::mesh<cache_neighbors_overwrite>;
-  using element_class = mesh_class::element_class;
-  mesh_class mesh = mesh_class (forest);
+  using mesh_class = t8_mesh_handle::mesh<t8_mesh_handle::competence_pack<cache_neighbors_overwrite>>;
+  using element_class = typename mesh_class::element_class;
+  const mesh_class mesh = mesh_class (forest);
```
sandro-elsweijer:
```suggestion
  const mesh_class mesh (forest);
```

test/mesh_handle/t8_gtest_handle_data.cxx
```diff
@@ -0,0 +1,152 @@
...
+TEST (t8_gtest_handle_data, set_and_get_user_data)
+{
+  // Define forest and mesh handle.
+  const int level = 2;
+  t8_cmesh_t cmesh = t8_cmesh_new_hypercube_hybrid (sc_MPI_COMM_WORLD, 0, 0);
+  const t8_scheme *init_scheme = t8_scheme_new_default ();
+  t8_forest_t forest = t8_forest_new_uniform (cmesh, init_scheme, level, 0, sc_MPI_COMM_WORLD);
+
+  using mesh_class = t8_mesh_handle::mesh<t8_mesh_handle::competence_pack<>, dummy_user_data>;
+  mesh_class mesh = mesh_class (forest);
```
sandro-elsweijer:
```suggestion
  mesh_class mesh (forest);
```

test/mesh_handle/t8_gtest_handle_data.cxx
```diff
@@ -0,0 +1,152 @@
...
+TEST (t8_gtest_handle_data, set_and_get_element_data)
+{
+  // Define forest and mesh handle.
+  const int level = 2;
+  t8_cmesh_t cmesh = t8_cmesh_new_hypercube_hybrid (sc_MPI_COMM_WORLD, 0, 0);
+  const t8_scheme *init_scheme = t8_scheme_new_default ();
+  t8_forest_t forest = t8_forest_new_uniform (cmesh, init_scheme, level, 1, sc_MPI_COMM_WORLD);
+
+  using mesh_class = t8_mesh_handle::mesh<t8_mesh_handle::competence_pack<>, void, data_per_element>;
+  mesh_class mesh = mesh_class (forest);
```
sandro-elsweijer:
```suggestion
  mesh_class mesh (forest);
```

test/mesh_handle/t8_gtest_mesh_handle.cxx
```diff
@@ -63,13 +64,13 @@ TEST_P (t8_mesh_handle_test, test_iterator)
 {
   // --- Check default functionality. ---
   using mesh_class = t8_mesh_handle::mesh<>;
-  using element_class = mesh_class::element_class;
-  mesh_class mesh = mesh_class (forest);
+  using element_class = typename mesh_class::element_class;
+  const mesh_class mesh = mesh_class (forest);
```
sandro-elsweijer:
```suggestion
  const mesh_class mesh (forest);
```

test/mesh_handle/t8_gtest_mesh_handle.cxx
```diff
@@ -97,14 +98,15 @@ TEST_P (t8_mesh_handle_test, test_iterator)
 TEST_P (t8_mesh_handle_test, test_competences)
 {
   // --- Version with cached vertex coordinates. ---
-  using mesh_class_vertex = t8_mesh_handle::mesh<t8_mesh_handle::cache_vertex_coordinates>;
-  using element_class_vertex = mesh_class_vertex::element_class;
-  mesh_class_vertex mesh_vertex = mesh_class_vertex (forest);
+  using competence_vertex_coordinates = t8_mesh_handle::competence_pack<t8_mesh_handle::cache_vertex_coordinates>;
+  using mesh_class_vertex = t8_mesh_handle::mesh<competence_vertex_coordinates>;
+  using element_class_vertex = typename mesh_class_vertex::element_class;
+  const mesh_class_vertex mesh_vertex = mesh_class_vertex (forest);
```
sandro-elsweijer:
```suggestion
  const mesh_class_vertex mesh_vertex (forest);
```

test/mesh_handle/t8_gtest_mesh_handle.cxx
```diff
@@ -133,14 +135,15 @@ TEST_P (t8_mesh_handle_test, test_competences)
   }
 
   // --- Version with cached centroid variable. ---
-  using mesh_class_centroid = t8_mesh_handle::mesh<t8_mesh_handle::cache_centroid>;
-  using element_class_centroid = mesh_class_centroid::element_class;
-  mesh_class_centroid mesh_centroid = mesh_class_centroid (forest);
+  using competence_centroid = t8_mesh_handle::competence_pack<t8_mesh_handle::cache_centroid>;
+  using mesh_class_centroid = t8_mesh_handle::mesh<competence_centroid>;
+  using element_class_centroid = typename mesh_class_centroid::element_class;
+  const mesh_class_centroid mesh_centroid = mesh_class_centroid (forest);
```
sandro-elsweijer:
```suggestion
  const mesh_class_centroid mesh_centroid (forest);
```

test/mesh_handle/t8_gtest_mesh_handle.cxx
```diff
@@ -160,17 +163,20 @@ TEST_P (t8_mesh_handle_test, test_competences)
 TEST_P (t8_mesh_handle_test, test_2_competences)
 {
   // --- Use competences to cache level and centroid. ---
-  using mesh_class = t8_mesh_handle::mesh<t8_mesh_handle::cache_vertex_coordinates, t8_mesh_handle::cache_centroid>;
-  using element_class = mesh_class::element_class;
-  mesh_class mesh = mesh_class (forest);
+  using mesh_class = t8_mesh_handle::mesh<t8_mesh_handle::cache_competences>;
+  using element_class = typename mesh_class::element_class;
+  const mesh_class mesh = mesh_class (forest);
```
sandro-elsweijer:
```suggestion
  const mesh_class mesh (forest);
```

mesh_handle/mesh.hxx
lenaploetzke: Oh yes, i also only thought of the interpolation after adaptation but you are right!

mesh_handle/mesh.hxx
lenaploetzke: I need `template <typename U = TUserData, typename = std::enable_if_t<!std::is_void<U>::value>>` instead of `template <typename = std::enable_if_t<!std::is_void<U>::value>>`. One alternative is `template <typename U = TUserData> ...   requires (!std::is_void_v<U>)` but we really need the alias because we work with TUserData& which throws an error in the void case. Maybe i should just work with * instead of &? This does not lead to errors bc void* is ok. I think that is the best option.

mesh_handle/mesh.hxx
lenaploetzke: If i rewrite set_element_data as you suggested, i again need the template alias because (void ...) is not a valid function argument. With ->requires, the function signature is evaluated even if requires evaluates to false.

mesh_handle/mesh.hxx
lenaploetzke: Ok i implemented it with the * but i dont like the solution as it gets really complicated and i like that the user cannot get or set nullptr. I will just stick to the template alias but of course find a better name

mesh_handle/mesh.hxx
lenaploetzke: Because i thought it is very nice if we have exchange and get in one function (because if you exchanged data you often want to access them directly or? And i thought if you do not want the data you can just ignore the return without any drawbacks

mesh_handle/mesh.hxx
lenaploetzke: As i decided against * and for &, this option is very cool!

mesh_handle/mesh.hxx
lenaploetzke: The primary use of this is the adapt () routine which is currently not implemented inside of the mesh but gets a mesh that should be adapted (and updates the forest). Maybe i should even move the adapt into the mesh?? Not sure

mesh_handle/mesh.hxx
lenaploetzke: This is very cool!

mesh_handle/element.hxx
lenaploetzke: Maybe an via an optional? Hm, alsp not a good idea bc of the MPI send?

mesh_handle/mesh.hxx
lenaploetzke: Is it really the case that we have to switcxh the user data type?

PR #2010, opened 2025-12-01
Feature: Automatically create VTK header file containing T8_ENABLE_VTK and version macros.

### LINKED ISSUE #2008: Feature: VTK lib header file with linkage macros.
# Feature request

When a library links against t8code, they cannot access T8_ENABLE_VTK.
We should make these available in t8code header.
This is possible using cmake `configure_file`.

External libraries using t8code cannot see if t8code is linked against another library (for example VTK) and with which version of that library.

Create a header file with

```C++
T8_ENABLE_VTK // 0 or 1 depending on whether t8code is linked against LIBNAME or not.
T8_VTK_MAJOR_VERSION // Defined only if T8_ENABLE_VTK  is true. The major version number of VTK that t8code links again.
T8_VTK_MINOR_VERSION // Defined only if T8_ENABLE_VTK  is true. The minor version number of VTK that t8code links again.  
```

N/A

**Estimated priority**

"Priority: medium"  Should be solved within half a year 

**Additional context**

holke:

Closes #2008 

External libraries were not able to see if t8code is linked against VTK and if so against which version of it.

CMake now builds a header file containing the macros:

```C++
/** Macro that is 1 if we link against VTK on 0 if we don't. */
#define T8_ENABLE_VTK 1
/** VTK major version number - Only defined if T8_ENABLE_VTK is 1 */
#define T8_VTK_MAJOR_VERSION 9
/** VTK minor version number - Only defined if T8_ENABLE_VTK is 1 */
#define T8_VTK_MINOR_VERSION 1
```

This header file is also installed and available to thirdparty libs.

I also modified the vtk linkage test to incorporate these macros.

#### General
#### Tests

#### Scripts and Wiki
#### License

### DIFF
--- a/src/t8_vtk/t8_vtk_linkage.hxx.in
+++ b/src/t8_vtk/t8_vtk_linkage.hxx.in
@@ -31,9 +31,9 @@
 
 /** Macro that is 1 if we link against VTK on 0 if we don't. */
 #cmakedefine01 T8_ENABLE_VTK
-/** VTK major version number */
+/** VTK major version number - Only defined if T8_ENABLE_VTK is 1 */
 #cmakedefine T8_VTK_MAJOR_VERSION @VTK_MAJOR_VERSION@
-/** VTK minor version number */
+/** VTK minor version number - Only defined if T8_ENABLE_VTK is 1 */
 #cmakedefine T8_VTK_MINOR_VERSION @VTK_MINOR_VERSION@
 
 #endif // T8CODE_VTK_LINKAGE_HXX


Davknapp:
Have you tried to export the targets using 
```
install(EXPORT targets 
FILE
....
DESTINATION
)
```
https://cmake.org/cmake/help/latest/command/install.html#export
?

I think this is the way we should actually go here (if it works).

holke:
> Have you tried to export the targets using
> 
> ```
> install(EXPORT targets 
> FILE
> ....
> DESTINATION
> )
> ```
> 
> https://cmake.org/cmake/help/latest/command/install.html#export ?
> 
> I think this is the way we should actually go here (if it works).

As i understand EXPORT is for build files from the CMaker syste and this would create a .cmake file " By default the generated file will be called <export-name>.cmake"?

We want the generated header to be copied over to the projects include instead.

Or do i miss something?

Davknapp:
You can find an alternative approach in #2012

PR #2098, opened 2026-01-15
Improvement: use struct instead of class everywhere

### LINKED ISSUE #2077: Improvement: Rename all instances of class to struct
Currently, we sometimes use `struct` and sometimes `class` in t8code, mostly because we have some C parts in the code. While the two are (almost) identical in C++, only `struct` exists in plain C. To ensure both compatibility and consistency, we decided in a dev meeeting to only use `struct` in the future and change all existing code.
This will also solve the warnings reported in #2068.

holke:

Closes #2077

Changes "class" to "struct" throughout the code base (not in mesh handle).
We decided to do this since some of our C code requires structs and classes and structs got mixed up and we encountered some unnecessary compiler errors.
To have a unique solution we switch to struct everywhere.

Please double check that i added the "private/protected/public" modifiers if they were necessary.

I decided against changing the wording of comments. I.e. a comment saying "This class does ..." still reads class, since i feel it is still valid to talk about a struct as a "class" in a comment.

#### General
#### Tests

#### Scripts and Wiki
#### License

### DIFF
--- a/mesh_handle/element.hxx
+++ b/mesh_handle/element.hxx
@@ -21,11 +21,10 @@ along with t8code; if not, write to the Free Software Foundation, Inc.,
 */
 
 /** \file element.hxx
- * Definition of the element class of the \ref t8_mesh_handle::mesh handle (can be ghost or mesh elements).
+ * Definition of an element of the \ref t8_mesh_handle::mesh handle (can be ghost or mesh elements).
  */
 
-#ifndef T8_ELEMENT_HXX
-#define T8_ELEMENT_HXX
+#pragma once
 
 #include <t8.h>
 #include <t8_element.h>
@@ -39,9 +38,9 @@ along with t8code; if not, write to the Free Software Foundation, Inc.,
 
 namespace t8_mesh_handle
 {
-/** Forward declaration of the \ref mesh class of the handle.
+/** Forward declaration of the \ref mesh the elements will belong to.
  */
-template <template <typename> class... TCompetences>
+template <template <typename> class... TCompetence>
 struct mesh;
 
 /** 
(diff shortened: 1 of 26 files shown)


mesh_handle/element.hxx
```diff
@@ -63,7 +63,8 @@ class mesh;
  * \tparam TCompetences The competences you want to add to the default functionality of the element.
  */
 template <template <typename> class... TCompetences>
-class element: public TCompetences<element<TCompetences...>>... {
+struct element: public TCompetences<element<TCompetences...>>...
+{
```
spenke91, line 66: Is the new line here (and in many other files) intentional? Personally, I happen to be a big fan of having the opening bracket in a new line, but I think that's not how we do it in t8code, is it?

mesh_handle/mesh.hxx
```diff
@@ -42,7 +42,8 @@ namespace t8_mesh_handle
  *         \see element for more details on the choice of the template parameter.   
  */
 template <template <typename> class... TCompetences>
-class mesh {
+struct mesh
+{
```
spenke91, line 45: See above -- but don't worry, I won't repeat it for the occurrences to follow ;-)

holke:
Hi @spenke91 

the new line was added by our indent script.
At first i did not add it and then the script complained that the code is not indented. So i ran the script and it did this change.
So it seems to be the t8code way. Unfortunately clang-format seems to indent class and struct differently here.

Yes, an addition to NEWS would be good and i shall write one!

spenke91:
> the new line was added by our indent script. At first i did not add it and then the script complained that the code is not indented. So i ran the script and it did this change. So it seems to be the t8code way. Unfortunately clang-format seems to indent class and struct differently here.

Ah, I see, thanks! Then let's keep it👍 
(If I were to decide, every developer world-wide would do this for every opening bracket anyway 😆)

holke:
Added NEWS.md and added a section in the developers guidelines: https://github.com/DLR-AMR/t8code/wiki/Coding-Guideline#struct-instead-of-class

holke:
Would have also loved to add a check for classes into our CI/indent script.
But i cannot see how to implement this properly.

We could grep for a regex like "class NAME", but that would hit a lot of false positives inside of comments...

So let us merge it as it is and should classes keep showing up in the future, we need to be inventive to prevent it ;)

spenke91:
> Would have also loved to add a check for classes into our CI/indent script. But i cannot see how to implement this properly.
> 
> We could grep for a regex like "class NAME", but that would hit a lot of false positives inside of comments...
> 

Yes, that would be ideal -- I also briefly looked into it, but using something like clang-tidy for that would be an overkill I'm afraid 🤔

PR #1736, opened 2025-07-04
Feature: extend leaf face neighbor to non-balanced forests and ghosts [N/N]

holke:
Closes #1737 and closes #1299

Will also close #328 

**Disclaimer**

~~This PR is not 2k lines as it shows currently.
It is based on many changes that are parts of different PRs but until these are not merged, their lines will show up in the count here.
I listed all the PRs below and also marked the sections that do not need review in the "Files Changed" tab.~~

**_Describe your changes here:_**

~~This has gotten quite large and will be split up and take some time.
I am putting this up here now to start the review early.~~

~~I would like to have a review on this because the algorithm is now working.
The review can focus mostly on the leaf_face_neighbor functionality.
It is still in draft status, because~~

~~a) there is still low-level functionality missing to make this work on all element shapes
b) i will split this up into multiple PRs eventually
c) not all tests pass, due to missing low-level functionality~~

Here is an overview over the exsiting PRs that cover changes from this one.
So a review of this PR can exclude all the listed once below:

- [x] Merge #1301 first (t8_element_array_find, t8_forest_element_is_leaf_or_ghost, t8_forest_tree_is_ghost, t8_forest_element_is_ghost)
- [x] Merge #1714 first (element_is_ancestor and ancestor search, adapt callbacks in test environment)
- [x] Merge #1806 first (t8_cmesh_get_tree_face_neighbor_eclass )
- [x] Merge #1952 first
- [x] Merge #1954 first (Bugfix in VTK empty partitions)
- [x] Merge #1955 first (eclass invalid string)
- [x] Merge #1958 first (element_face_get_ancestor_face)
- [x] Merge #1968 first (const params for face iterator)

**Description**

This PR extends the leaf face neighbor computation in two ways simultaneously:
a) extend it to ghosts, thus we can compute neighbors of ghosts and neighbors that are ghosts
b) extend it to non-balanced forests. So far a 2:1 condition was required, this is no longer the case.

The main contributions are
- extending many "low-level" element and element array handling functions to ghosts (Could be own PR)
- extending the face iteration function to ghosts 
- t8_element_array_find function to search for an element in an array (Should be its own PR)
- t8_forest_leaf_face_neighbors_ext extension to non-balances and ghosts
- t8_forest_same_level_leaf_face_neighbor_index  to compute the data indices of neighbors if only single neighbor (TODO: this was
implemented while the lfn function did not work for adaptive forests, need to check whether to extend this function to all forests now) (Should be its own PR)
[...]

Issue #1299, opened 2024-11-13
Extend leaf_face_neighbor function to ghost elements


holke:
# Feature request

**Is your feature request related to a problem? Please describe.**
For some applications it is required to compute the face neighbors of ghost elements (if they are local element or ghosts).
Currently our lfn function does not support ghost elements as input.

**Describe the solution or feature you'd like**
Extend the lfn function to accept ghost elements as input parameters.
Return only those neighbors that are local elements or ghosts.

**Describe alternatives you've considered**
n/a

**Estimated priority**

"Priority: high"  Should be solved as soon as possible 

I need this currently.

**Additional context**
I started working on this. Mostly created the issue to inform other about the existing plans.

PR #2214, opened 2026-03-06
Improvement: avoid segfault when passing a nullptr to t8_forest_element_face_neighbor

dutkalex:
The last argument of `t8_forest_element_face_neighbor` is an output parameter which gets filled with the dual face index. In some situations, this additional output is not desired and one would like to pass `nullptr` (as is often done in other t8code APIs) to signal this. Currently, this results in a hard crash. This PR enables this use case.

#### General
#### Tests

#### Scripts and Wiki
#### License


holke:
Thank you for this addition!
This is currently on hold since it conflicts with our large changes in #1299. But i will see how we integrate both of these.

src/t8_forest/t8_forest.cxx
```diff
@@ -1462,6 +1462,11 @@ t8_gloidx_t
 t8_forest_element_face_neighbor (t8_forest_t forest, t8_locidx_t ltreeid, const t8_element_t *elem, t8_element_t *neigh,
                                  const t8_eclass_t neigh_eclass, int face, int *neigh_face)
 {
+  int dummy = 0;
```
benegee, line 1463: I was first uncertain about this dummy. After looking through the code it is clear the variable is required locally. It is still an output, and if a user does not actually need it, it seems ok to just discard it.

src/t8_forest/t8_forest_general.h
```diff
@@ -893,7 +893,7 @@ t8_forest_element_neighbor_eclass (const t8_forest_t forest, const t8_locidx_t l
  *                          the data could be modified arbitrarily.
  * \param [in] neigh_eclass The eclass of \a neigh.
  * \param [in] face         The number of the face along which the neighbor should be constructed.
- * \param [out] neigh_face  The number of the face viewed from perspective of \a neigh.
+ * \param [out] neigh_face  The number of the face viewed from perspective of \a neigh. Can be nullptr.
```
benegee:
```suggestion
 * \param [out] neigh_face  The number of the face viewed from perspective of \a neigh.
 *                          Can be nullptr, in which case the output is discarded.
```

src/t8_forest/t8_forest.cxx
benegee: Checked this. I think the TODO can be removed.

benegee (changes requested):
Thanks for your suggestion!

With #1736 merged, this is still possible from my point of view.

src/t8_forest/t8_forest.cxx
dutkalex: This is the simplest and least intrusive change to allow for `nullptr` to be passed to this function. Conceptually, it boils down to moving inside of the function the call site boilerplate that callers are doing right now if they don't care about `neigh_face`. Reading your comment it is not 100% clear to me if you agree with this implementation or if you want me to change something however.

src/t8_forest/t8_forest.cxx
dutkalex: I don't quite understand what this TODO comment is referring to. Is it related to the change implemented in this PR?

src/t8_forest/t8_forest.cxx
benegee: Sorry for the confusion!
I agree!
But thanks for the explanation and pointing at other location in the code where this paradigm is applied!

src/t8_forest/t8_forest.cxx
benegee: No, I just noted the TODO in passing, checked it and decided it could be removed. However, I didn't want to commit to your PR. If you are fine with it, you could remove the TODO.

dutkalex: Done

benegee (approved):

Issue #2215, opened 2026-03-06
Add a helper function to compute neighbor element subfaces


dutkalex:
# Feature request
Implement a helper function to complement the information provided by `t8_forest_leaf_face_neighbors_ext` in the case of coarser neighbors.

**Is your feature request related to a problem? Please describe.**
For high-order (DG-like) solvers, it is common to have to build a detailed face connectivity data structure, most of which can be filled by calling `t8_forest_leaf_face_neighbors_ext` on each of the partition-local leaves. However, in the particular case where the neighbor across a face is coarser, this function does not tell which subface of the neighbor is touching the considered element.

**Describe the solution or feature you'd like**
Adding a helper function that can be called after `t8_forest_leaf_face_neighbors_ext` to compute the neighbor subface id would be an efficient and backwards-compatible way to solve the problem.

**Describe alternatives you've considered**
As discussed with @holke, a possible alternative (that only works for local neighbors for now but will shortly become a general solution when #1736 gets merged) is to call a second time `t8_forest_leaf_face_neighbors_ext` on the neighbor. However, this implies a lot of wasted work, since only a small fraction of the output would be useful (the major part is redundant).

**Estimated priority**
Which of these is most applicable (remove the others):

"Priority: high"  Should be solved as soon as possible 

"Priority: medium"  Should be solved within half a year 

"Priority: low" Should be solved eventually

**Additional context**
This is somewhat related to #2190

spenke91:
Closed by #2216

PR #2216, opened 2026-03-06
Feature: add neighbor subface helper

dutkalex:
Implements #2215 

#### General
#### Tests

#### Scripts and Wiki
#### License


src/t8_forest/t8_forest.cxx
```diff
@@ -1885,6 +1885,38 @@ t8_forest_leaf_face_neighbors (t8_forest_t forest, t8_locidx_t ltreeid, const t8
...
 }
 
+int
+t8_forest_leaf_neighbor_subface (t8_forest_t forest, t8_locidx_t ltreeid, const t8_element_t *leaf, int face,
+                                 t8_eclass_t neighbor_tree_class, const t8_element_t *neighbor_leaf, int neighbor_face)
+{
+  t8_scheme const *scheme = t8_forest_get_scheme (forest);
+
+  t8_element_t *target = nullptr;
+  scheme->element_new (neighbor_tree_class, 1, &target);
```
spenke91:
```suggestion
  // Preparation.
  t8_scheme const *scheme = t8_forest_get_scheme (forest);
  t8_element_t *target = nullptr;
  scheme->element_new (neighbor_tree_class, 1, &target);
```

src/t8_forest/t8_forest_general.h
```diff
@@ -645,6 +645,21 @@ t8_forest_leaf_face_neighbors_ext (t8_forest_t forest, t8_locidx_t ltreeid, cons
...
+ * \param [in] forest       The forest. Must be committed and balanced.
+ * \param [in] ltreeid      A local tree id.
+ * \param [in] leaf         A leaf in \a ltreeid.
+ * \param [in] face         The face index of \a leaf to consider.
+ * \param [in] neighbor_tree_eclass The eclass of the neighbor element.
+ * \param [in] neighbor_leaf The leaf of \a forest on the other side of the face of index \a face of element \a leaf. Must be one level coarser than \a leaf.
+ * \param [in] neighbor_face The face index of \a neighbor_leaf (i.e. the dual face of \a face).
+ * \returns The index of the subface of \a neighbor_face which corresponds to \a face.
+ * \note This function is designed to be called after \ref t8_forest_leaf_face_neighbors_ext to complement its output.
+ */
```
spenke91, line 693:
```suggestion
 * \note Since this function requires the \a neighbor_leaf to be one level coarser than \a leaf, this function is primarily intended for balanced forests.
 */
```

src/t8_forest/t8_forest.cxx
```diff
@@ -1885,6 +1885,38 @@ t8_forest_leaf_face_neighbors (t8_forest_t forest, t8_locidx_t ltreeid, const t8
...
+
+  int dummy;  // Can't pass a nullptr to t8_forest_element_face_neighbor below (see #2214)
+  t8_forest_element_face_neighbor (forest, ltreeid, leaf, target, neighbor_tree_class, face, &dummy);
```
spenke91: Maybe we can find a more expressive name than `target`? e.g.
```suggestion
  // Determine the (virtual) face neighbor
  // Note: For this function to work properly, this face neighbor has to be a direct child of neighbor_leaf
  t8_element_t *virtual_face_neighbor= nullptr;
  scheme->element_new (neighbor_tree_class, 1, &virtual_face_neighbor);
  int dummy;  // Can't pass a nullptr to t8_forest_element_face_neighbor below (see #2214)
  t8_forest_element_face_neighbor (forest, ltreeid, leaf, virtual_face_neighbor, neighbor_tree_class, face, &dummy);
```
Or maybe we can even find something better? `sub_neighbor`?

src/t8_forest/t8_forest.cxx
```diff
@@ -1885,6 +1885,38 @@ t8_forest_leaf_face_neighbors (t8_forest_t forest, t8_locidx_t ltreeid, const t8
...
+
+  int const num_children = scheme->element_get_num_face_children (neighbor_tree_class, neighbor_leaf, neighbor_face);
+
+  std::array<t8_element_t *, 4> children;  // assumes a 2:1 balanced forest
+  scheme->element_new (neighbor_tree_class, 4, children.begin ());
+
+  scheme->element_get_children_at_face (neighbor_tree_class, neighbor_leaf, neighbor_face, children.begin (),
+                                        num_children, nullptr);
+
```
spenke91, line 2094:
```suggestion
Iterate over the neighbor's children and compare with virtual face neighbor to find the sub-face ID.
```

src/t8_forest/t8_forest.cxx
```diff
@@ -1885,6 +1885,38 @@ t8_forest_leaf_face_neighbors (t8_forest_t forest, t8_locidx_t ltreeid, const t8
...
+  int result = -1;
+  for (int i_child = 0; i_child < num_children; ++i_child) {
+    if (scheme->element_compare (neighbor_tree_class, target, children[i_child]) == 0) {
+      result = i_child;
+    }
+  }
+
```
spenke91, line 2100:
```suggestion

// Free memory and return sub-face ID.
```

src/t8_forest/t8_forest.cxx
```diff
@@ -1885,6 +1885,38 @@ t8_forest_leaf_face_neighbors (t8_forest_t forest, t8_locidx_t ltreeid, const t8
...
+{
+  t8_scheme const *scheme = t8_forest_get_scheme (forest);
+
+  t8_element_t *target = nullptr;
+  scheme->element_new (neighbor_tree_class, 1, &target);
+
+  int dummy;  // Can't pass a nullptr to t8_forest_element_face_neighbor below (see #2214)
+  t8_forest_element_face_neighbor (forest, ltreeid, leaf, target, neighbor_tree_class, face, &dummy);
+
+  int const num_children = scheme->element_get_num_face_children (neighbor_tree_class, neighbor_leaf, neighbor_face);
```
spenke91, line 2087:
```suggestion
  // Get the number of the neighbor's children touching the neighbor_face or, equivalenlty, the face's number of children.
  int const num_neighbor_face_children = scheme->element_get_num_face_children (neighbor_tree_class, neighbor_leaf, neighbor_face);
```

src/t8_forest/t8_forest.cxx
```diff
@@ -1885,6 +1885,38 @@ t8_forest_leaf_face_neighbors (t8_forest_t forest, t8_locidx_t ltreeid, const t8
...
+
+  std::array<t8_element_t *, 4> children;  // assumes a 2:1 balanced forest
+  scheme->element_new (neighbor_tree_class, 4, children.begin ());
+
+  scheme->element_get_children_at_face (neighbor_tree_class, neighbor_leaf, neighbor_face, children.begin (),
+                                        num_children, nullptr);
```
spenke91: It may well be that I missed something here, but shouldn't we use `num_face_children` rather than 4? 
```suggestion
  // Create an array containing the neighbor's children at neighbor_face.
  std::array<t8_element_t *, num_neighbor_face_children> neighbor_children_at_face;
  scheme->element_new (neighbor_tree_class, num_neighbor_face_children, neighbor_children_at_face.begin ());
  scheme->element_get_children_at_face (neighbor_tree_class, neighbor_leaf, neighbor_face, neighbor_children_at_face.begin (), num_neighbor_face_children, nullptr);
```

src/t8_forest/t8_forest.cxx
```diff
@@ -1885,6 +1885,38 @@ t8_forest_leaf_face_neighbors (t8_forest_t forest, t8_locidx_t ltreeid, const t8
...
+
+  int result = -1;
+  for (int i_child = 0; i_child < num_children; ++i_child) {
+    if (scheme->element_compare (neighbor_tree_class, target, children[i_child]) == 0) {
```
spenke91:
```suggestion
  for (int i_child = 0; i_child < num_neighbor_face_children; ++i_child) {
    if (scheme->element_compare (neighbor_tree_class, virtual_face_neighbor, neighbor_children_at_face[i_child]) == 0) {
```

src/t8_forest/t8_forest.cxx
spenke91, line 2100: Should we maybe assert that results was set here?
```suggestion
  // Make sure the face was found.
  T8_ASSERT(results != -1)
```

spenke91 (changes requested):
Thank you very much @dutkalex for another great contribution! :-) I am sorry it took us so long to reply...
I mostly have suggestions regarding variable names and code comments (which, I know, not everybody loves as much as I do 🤓 ). Aside from that, I only have a question regarding the hard-coded `4`.

src/t8_forest/t8_forest.cxx
dutkalex: _"There are only two hard things in computer science: cache invalidation and naming things."_ 😉

I see your point, and I guess since I could not come up with a better name at the time and went for the rather undescriptive but concise option. I feel like `virtual_face_neighbor` is more descriptive than `sub_neighbor`, but it looses the "it's what we're looking for" idea. It seems hard to find a (reasonably sized) name which encodes unambiguously all this information. Maybe this means a small comment is necessary to clarify exactly what this is... What do you think?

src/t8_forest/t8_forest.cxx
dutkalex: I can see why you'd want the variable name to be a little bit more explicit, but isn't the comment just paraphrasing what `element_get_num_face_children`'s documentation already says?
I tend to think that comments should be used to explain why something is done a certain way, but not what the code does (that should be clear by reading the code, and if you think it's not the case maybe I should change the code then).

src/t8_forest/t8_forest.cxx
dutkalex: There are two reasons for why we can't use `num_face_children`/`num_neighbor_face_children` here:
1. because we need an integral constant expression here (i.e. a `constexpr int`-like thing)
2. because what we need here is the upper-bound of the number of face neighbors (this happens to be 4 for a 2:1 balanced interface) so that this piece of code works for any element type

Maybe there is a properly named global constant which could be used instead of this magic number?

src/t8_forest/t8_forest.cxx
dutkalex: I guess this is meant to be a comment. However, after re-reading with fresh eyes this piece of code, I think this is just a `std::find_if` and that refactoring the code to use this is the proper way to make the intent explicit here

src/t8_forest/t8_forest.cxx
dutkalex: I think the "free memory" part is quite self evident here, but the "return sub-face ID" part is indeed not clear with the current state of the code. However maybe this just means that `result` is a bad name ? 😅

dutkalex: I'll add this manually to correct the typo

dutkalex:
> Thank you very much @dutkalex for another great contribution! :-) I am sorry it took us so long to reply...

No worries @spenke91. Thank you for taking the time, I know you guys have a lot on your plate.

> I mostly have suggestions regarding variable names and code comments (which, I know, not everybody loves as much as I do 🤓 ). Aside from that, I only have a question regarding the hard-coded `4`.

I get it, I often say that getting the compiler to understand the code is the easy part and that the real challenge it to write code that humans can comprehend 😉. I have either committed your suggestions directly, or answered your comments/questions to get your opinion. Let me know what you think and I'll revise the code accordingly.

src/t8_forest/t8_forest.cxx
dutkalex: I'll make the adequate changes here when the previous comments are settled.

src/t8_forest/t8_forest.cxx
spenke91: Love the quote!😄 

I think a comment is a good idea anyway (as I said, I will basically never say "no" to code comments 🙈 )
We all do not love long variable names, but what's the harm of using maybe `target_virtual_neighbor` or even `target_virtual_face_neighbor` -- after all, this isn't some old Fortran code with 72 characters per line 😉 
With a comment, however, I also don't mind sticking to `target` if that's what you prefer.

src/t8_forest/t8_forest.cxx
spenke91: Ah, of course this should be a comment not C++ code 😬

spenke91: Good idea, `std::find_if` would make it nicer and clearer 👍

src/t8_forest/t8_forest.cxx
spenke91: > I tend to think that comments should be used to explain why something is done a certain way, but not what the code does (that should be clear by reading the code...)

I guess I have a fundamentally different philosophy here, but that is definitely not a problem at all! :-) Because it is in fact just my personal preference and not part of any t8code coding guidelines. So if you feel like some of the comments I suggested are too much, don't worry and feel free not to add them 👍 

In case you are interested, my philosphy is more that reading the doxygen header of a function along with the inline code comments should already give an idea about what the code does -- and how it does it, meaning the structure of the algorithm. For example, the line code comments I suggested alone would read

```
  // Determine the (virtual) face neighbor
  // Note: For this function to work properly, this face neighbor has to be a direct child of neighbor_leaf

  // Get number of neighbor's children

  // Allocate memory

  // Iterate over the neighbor's subfaces and compare with target to find the sub-face ID.

  // Free memory and return sub-face ID.

```
which, I'd argue, already gives you a rough but useful outline of the function without reading one single line of code.
The motivation behind this is that I am convinced even a proficient software developer can still read English faster than C++ 😉  

(And as for the added work of writing the comments, I am certain the time is neglible compared to the time saved for future readers -- after all, while writing the code we know what we're doing anyway!)

So regarding your quote here

> I get it, I often say that getting the compiler to understand the code is the easy part and that the real challenge it to write code that humans can comprehend 😉

this would be my answer, that is, paraphrasing the code in a language made for humans to understand, not compilers 😉 
But as I said, I know not everybody would agree, which is fine 👍

src/t8_forest/t8_forest.cxx
spenke91: Maybe `target_face_id` or something like that?

src/t8_forest/t8_forest.cxx
spenke91: Whoops, the integral constant expression point is in fact quite obvious here, sorry for missing that.

How about defining a new constant to `t8_eclass.h`
```
/** The maximal number of children a face may have.*/
#define T8_ECLASS_MAX_FACE_CHILDREN 4
```
or
```
/** The maximal number of children a 2D element may have.*/
#define T8_ECLASS_MAX_CHILDREN_2D 4
```
and using it here? Then it should also be clear where the number 4 comes from.

spenke91 (commented):
Thanks a lot for your comments and replies!

src/t8_forest/t8_forest.cxx
dutkalex: I like `target_virtual_face_neighbor`. Let's go with that :)

src/t8_forest/t8_forest.cxx
dutkalex: I see your point. I tend to be somewhat cautious with comments as I feel they are often used as excuses for writing bad code, and they tend to age poorly and become obsolete. In my experience, there is nothing more frustrating than cryptic code except for cryptic code with outdated misleading comments lying around 😅
However, at the end of the day this is your call and I'm ok adding the comment if you feel this is better for the code

src/t8_forest/t8_forest.cxx
dutkalex: No need to apologize, we've all been there 😉
I like your suggestion. I'll make the change

src/t8_forest/t8_forest.cxx
spenke91: >  there is nothing more frustrating than cryptic code except for cryptic code with outdated misleading comments lying around

I see :D 
No, let's keep it as you suggested here, that is totally fine and the comments are not necessary here -- I just happened to be in the mood for discussing some general thoughts on this ;-)

spenke91 (dismissed):
Thanks a lot @dutkalex for this contribution! 👍

src/t8_forest/t8_forest_general.h
```diff
@@ -675,6 +675,26 @@ t8_locidx_t
 t8_forest_same_level_leaf_face_neighbor_index (const t8_forest_t forest, const t8_locidx_t element_index,
                                                const int face_index, const t8_gloidx_t global_treeid, int *dual_face);
 
+/** Compute the subface index for a coarser neighbor
+ * \param [in] forest       The forest. Must be committed.
+ * \param [in] ltreeid      A local tree id.
+ * \param [in] leaf         A leaf in \a ltreeid.
+ * \param [in] face         The face index of \a leaf to consider.
+ * \param [in] neighbor_tree_eclass The eclass of the neighbor element.
```
spenke91: Just a little typo... this should fix the doxygen error:

```suggestion
 * \param [in] neighbor_tree_class The eclass of the neighbor element.
```

spenke91 (commented):
The doxygen workflow is failing due to a minor typo...

spenke91 (dismissed):

spenke91:
@dutkalex The failing doxygen workflow has nothing to do with your PR specifically. We have a rather unstable setup there which I will fix in the next days (#2331). Until then, I tried restarting the workflows.

spenke91 (approved):

PR #2256, opened 2026-04-01
Feature: OpenFOAM reader [1/n] - First structure

sandro-elsweijer:
Closes #2257

Implements a first structure of the OpenFOAM reader.
The first draft of the OpenFOAM reader struct reads the files inside an OpenFOAM case and fills internal data structures.
The conversion of this data into a cmesh follows in the next PRs. Otherwise this PR would be too big to review.
The reader is not inside cmesh/IO/OpenFOAM, since the reader will not only provide a cmesh, but also a forest with data. So there is no clear distinction between the reader cmesh and forest and the OpenFOAM functionality is its own category.

This PR also implements an OpenFOAM example and it has two purposes:
1. The OpenFOAM reader is a header file and this example makes cmake compile the header.
2. During development, the example can be used to test and debug the code.

There is no test yet, since the OpenFOAM reader is not finished yet. I plan to implement two tests:
1. A test where a very simple mesh is read and manually checked.
2. A component test where we use an OpenFOAM writer to write a mesh and afterwards we read it using the OpenFOAM reader. The output and input mesh should be equal.
Both tests can only be written, when the reader is finished.

#### General
#### Tests

#### Scripts and Wiki
#### License


src/t8_openfoam/t8_openfoam_reader.hxx
```diff
@@ -0,0 +1,490 @@
...
+ * Reads an OpenFOAM case into t8code data structures.
+ */
+struct t8_openfoam_reader
+{
+ public:
+  /** Abbreviation for std::filesystem::path */
+  using t8_path = std::filesystem::path;
+
+  /**
+   * Reader for OpenFOAM cases.
```
spenke91:
```suggestion
   * Constructor of OpenFOAM reader.
```

src/t8_openfoam/t8_openfoam_reader.hxx
```diff
@@ -0,0 +1,490 @@
...
+    const t8_path case_neighbor_dir = m_case_dir / "constant/polyMesh/neighbour";
+    const t8_path case_owner_dir = m_case_dir / "constant/polyMesh/owner";
+    const t8_path case_points_dir = m_case_dir / "constant/polyMesh/points";
+
+    /* The file reading needs to happen in this order, since these functions depend on each other. */
+    bool error = 0;
+    error = !read_points (case_points_dir);
+    error &= !read_faces (case_faces_dir);
+    error &= !read_owner (case_owner_dir);
+    error &= !read_neighbor (case_neighbor_dir);
```
spenke91: As far as I know, `&=` is for bitwise AND, not logical AND. While it does in fact work for bools, I think we should reseve it for bitwise operations for code readability. I thus suggest:
```suggestion
    error = !read_points (case_points_dir);
    error = error && !read_faces (case_faces_dir);
    error = error && !read_owner (case_owner_dir);
    error = error && !read_neighbor (case_neighbor_dir);
```
or with `and` instead of `&&`, which, personally, I like even more ;-)

src/t8_openfoam/t8_openfoam_reader.hxx
```diff
@@ -0,0 +1,490 @@
+/*
```
spenke91, line 1: What is your motivation behind having the full implementation in the header file instead of a source file? None of the standard cases (template, inline, ...) seems to apply, or did I miss something? 🤔

src/t8_openfoam/t8_openfoam_reader.hxx
```diff
@@ -0,0 +1,490 @@
...
+    //  (0 0 0)
+    //  (0.5 0 0)
+    //  (1 0 0)
+    //  (0 0.5 0)
+    //  [ ... ]
+
+    std::string line;
+    bool format_checked = false;
+
+    /* Loop through file till end of file or thill we break manually at the right position. */
```
spenke91:
```suggestion
    /* Loop through file till end of file or till we break manually at the right position. */
```

src/t8_openfoam/t8_openfoam_reader.hxx
```diff
@@ -0,0 +1,490 @@
...
+        }
+      }
+
+      /* We do not need to check for numeric values when the format string did not appear yet. */
+      if (format_checked) {
+        /* Skip to first numeric value and then revert to start of numeric value */
+        size_t numeric_value;
+        if (line_stream >> numeric_value) {
+          /* Numeric value found, move stream to start of line again */
+          input_stream.seekg (-static_cast<std::streamoff> (line.size ()) - 1, std::ios_base::cur);
```
spenke91: Why do we have to move input_stream back to the start of the line? Isn't it still at the start of the line anyway? Because the only command in the `while` loop modifying it is in the condition, i.e., `std::getline (input_stream, line))`. Or did I miss something?

src/t8_openfoam/t8_openfoam_reader.hxx
```diff
@@ -0,0 +1,490 @@
...
+
+    /* Get list opening delimiter */
+    char opening_delimiter = '\0';
+    input_stream >> opening_delimiter;
+
+    values.clear ();
+    values.reserve (list_size);
+
+    /* Expanded or compact list: <num_values>(val_0 ... val_n) with or without line break */
+    if (opening_delimiter == '(') {
```
spenke91: How about a switch statement with the default case throwing an error à la "Unrecognized encoding of label list"?

src/t8_openfoam/t8_openfoam_reader.hxx
```diff
@@ -0,0 +1,490 @@
...
+        m_points.clear ();
+        return false;
+      }
+      m_points.emplace_back (point);
+    }
+    return true;
+  }
+
+  /**
+   * Reads in an OpenFOAM faces file stores the faces.
```
spenke91:
```suggestion
   * Reads in an OpenFOAM faces file and stores the faces in the member variable \a m_face_points.
```

src/t8_openfoam/t8_openfoam_reader.hxx
```diff
@@ -0,0 +1,490 @@
...
+      /* Assign uniform values to list. */
+      values.assign (list_size, uniform_value);
+      return true;
+    }
+
+    return false;
+  }
+
+  /**
+   * Reads in an OpenFOAM points file and stores the points.
```
spenke91:
```suggestion
   * Reads in an OpenFOAM points file and stores the points in the member variable \a m_points.
```

src/t8_openfoam/t8_openfoam_reader.hxx
```diff
@@ -0,0 +1,490 @@
...
+  /**
+   * Reads in an OpenFOAM owner file, assigns the faces to their cells and
+   * also fills one half of the neighborhoods.
+   * \param [in] owner_dir  The path to the owner file.
+   * \return                True on success.
+   */
+  bool
+  read_owner (const t8_path& owner_dir)
+  {
+    T8_ASSERT (!m_face_points.empty ());
```
spenke91: Could maybe even be more than a T8_ASSERT and checked in Release mode too? But I don't mind either way

src/t8_openfoam/t8_openfoam_reader.hxx
```diff
@@ -0,0 +1,490 @@
...
+    const t8_path case_points_dir = m_case_dir / "constant/polyMesh/points";
+
+    /* The file reading needs to happen in this order, since these functions depend on each other. */
+    bool error = 0;
+    error = !read_points (case_points_dir);
+    error &= !read_faces (case_faces_dir);
+    error &= !read_owner (case_owner_dir);
+    error &= !read_neighbor (case_neighbor_dir);
+
+    if (error) {
```
spenke91: Aren't these files rather than directories? How about renaming them accordingly? :-)
```suggestion
    /* Build paths to the standard OpenFOAM mesh files. */
    const t8_path case_faces_file = m_case_dir / "constant/polyMesh/faces";
    const t8_path case_boundary_file = m_case_dir / "constant/polyMesh/boundary";
    const t8_path case_neighbor_file = m_case_dir / "constant/polyMesh/neighbour";
    const t8_path case_owner_file = m_case_dir / "constant/polyMesh/owner";
    const t8_path case_points_file = m_case_dir / "constant/polyMesh/points";

    /* The file reading needs to happen in this order, since these functions depend on each other. */
    bool error = 0;
    error = !read_points (case_points_file);
    error &= !read_faces (case_faces_file);
    error &= !read_owner (case_owner_file);
    error &= !read_neighbor (case_neighbor_file);

    if (error) {
```
If you feel like `file` sounds too much like it's just the name without the path, maybe `case_faces_path` etc.?

src/t8_openfoam/t8_openfoam_reader.hxx
```diff
@@ -0,0 +1,490 @@
...
+      values.assign (list_size, uniform_value);
+      return true;
+    }
+
+    return false;
+  }
+
+  /**
+   * Reads in an OpenFOAM points file and stores the points.
+   * \param [in] points_dir The path to the points file.
```
spenke91: See comment above: I suggest to name it `points_file`.

src/t8_openfoam/t8_openfoam_reader.hxx
```diff
@@ -0,0 +1,490 @@
...
+        return false;
+      }
+      m_points.emplace_back (point);
+    }
+    return true;
+  }
+
+  /**
+   * Reads in an OpenFOAM faces file stores the faces.
+   * \param [in] faces_dir  The path to the faces file.
```
spenke91: See comment above: I suggest to name it `faces_file`

src/t8_openfoam/t8_openfoam_reader.hxx
```diff
@@ -0,0 +1,490 @@
...
+      }
+      m_face_points.emplace_back (std::move (points));
+    }
+    return true;
+  }
+
+  /**
+   * Reads in an OpenFOAM owner file, assigns the faces to their cells and
+   * also fills one half of the neighborhoods.
+   * \param [in] owner_dir  The path to the owner file.
```
spenke91: See comment above: I suggest to name it `owner_file`.

src/t8_openfoam/t8_openfoam_reader.hxx
```diff
@@ -0,0 +1,490 @@
...
+    /** List of read values from owners file. Position is equal to face id and value
+     * is the owner of the face. face_id -> cell_id */
+    std::vector<size_t> owner_list;
+    if (!read_openfoam_label_list (file, owner_list)) {
+      t8_errorf ("ERROR: Unable to parse file: %s\n", owner_dir.c_str ());
+      return false;
+    }
+
+    /* The number of cells in the mesh is the highest owner_id + 1 (since it starts at 0). */
+    const size_t num_cells = *std::max_element (owner_list.begin (), owner_list.end ()) + 1;
```
spenke91: I guess this part is not expected to be very performance-critical. Otherwise it might make sense to keep track of the maximum already in the for loop above filling the vector. This way we would not have to iterate through the whole vector again. But it would just be a memory access / caching thing, not so much a complexity issue, so I guess it is not important at all :D

src/t8_openfoam/t8_openfoam_reader.hxx
```diff
@@ -0,0 +1,490 @@
...
+      const size_t cell_id = owner_list[face_id];
+      /* Add face to owner cell. Normal points outwards (1). */
+      m_cell_faces[cell_id].emplace_back (std::make_pair (face_id, 1));
+    }
+
+    return true;
+  }
+
+  /**
+   * Reads in an OpenFOAM owner file, assigns the faces to their cells and
```
spenke91:
```suggestion
   * Reads in an OpenFOAM neighbor file, assigns the faces to their cells and
```
Right?

src/t8_openfoam/t8_openfoam_reader.hxx
```diff
@@ -0,0 +1,490 @@
...
+   * also fills the other half of the neighborhoods.
+   * \param [in] neighbor_dir  The path to the points file.
```
spenke91: As above, maybe `neighbor_file`?

src/t8_openfoam/t8_openfoam_reader.hxx
```diff
@@ -0,0 +1,490 @@
...
+
+  /**
+   * Build the cmesh from the raw mesh data.
+   * \return  True on success
+   */
+  bool
+  build_cmesh ()
+  {
+    t8_global_errorf ("ERROR: Not implemented yet. \n");
+    return 0;
```
spenke91:
```suggestion
    return false;
```

example/IO/cmesh/OpenFOAM/t8_read_openfoam_case.cxx
```diff
@@ -0,0 +1,120 @@
+/*
```
spenke91, line 1: Could you add an example case file? I think it would really improve this example if the user could test it for an example file easily without having to find one him-/herself 😬

spenke91 (changes requested):
Thanks a lot for this feature @sandro-elsweijer ! I really like the code structure and style 👍 I do have some remarks, though, most of them being minor issues...

I went through the code, but did not run the example myself yet, since I did not have an OpenFOAM case readily available and would argue that we should add an example case anyway 😇

src/t8_openfoam/t8_openfoam_reader.hxx
sandro-elsweijer: done

src/t8_openfoam/t8_openfoam_reader.hxx
sandro-elsweijer: done

src/t8_openfoam/t8_openfoam_reader.hxx
sandro-elsweijer: done

src/t8_openfoam/t8_openfoam_reader.hxx
sandro-elsweijer: done

src/t8_openfoam/t8_openfoam_reader.hxx
sandro-elsweijer: done

src/t8_openfoam/t8_openfoam_reader.hxx
sandro-elsweijer: done

src/t8_openfoam/t8_openfoam_reader.hxx
sandro-elsweijer: done

src/t8_openfoam/t8_openfoam_reader.hxx
sandro-elsweijer: Moved it into a source file

src/t8_openfoam/t8_openfoam_reader.hxx
sandro-elsweijer: because we already called `std::getline()` and the next call would move return the next line. But in the functions we need exactly this line and there is no way to call `std::getline()` again on the same line

src/t8_openfoam/t8_openfoam_reader.hxx
sandro-elsweijer: done

src/t8_openfoam/t8_openfoam_reader.hxx
sandro-elsweijer: this shouldnt happen anyway. This is just a check for the internal logic and since this is not open to the user they cannot mess this up. And if something fails the test (which i am about to write) should fail in the ci in debug

src/t8_openfoam/t8_openfoam_reader.hxx
sandro-elsweijer: yeah but then i would have to add a feature for this one use case in the function. so i chose not to do it since it is not really critical

example/IO/cmesh/OpenFOAM/t8_read_openfoam_case.cxx
sandro-elsweijer: done in t8data

src/t8_openfoam/t8_openfoam_reader.hxx
spenke91: Thanks 👍

src/t8_openfoam/t8_openfoam_reader.hxx
spenke91: Fine with me 👍

src/t8_openfoam/t8_openfoam_reader.hxx
spenke91: Thanks👍

example/IO/cmesh/OpenFOAM/t8_read_openfoam_case.cxx
spenke91: Great, thanks 👍

src/t8_openfoam/t8_openfoam_reader.hxx
spenke91: Perfect, thank you! 😇

src/t8_openfoam/t8_openfoam_reader.hxx
spenke91: Ah, so I did indeed miss something 😆 Thanks for the clarification!
I still feel like this is not a very aesthetic solution, but alternatives like additionally passing the current line as string to the subsequent functions don't seem any better, so let's keep it as it is 👍

spenke91 (commented):
Thanks for the update! The code changes all look good to me.
I will test the reader later today with the example case -- and then most probably give you a green light 🟢

spenke91 (dismissed):

spenke91 (approved):

PR #2276, opened 2026-04-16
Improvement: add test for `t8_forest_element_neighbor_eclass`

benegee:

I am trying to resolve #2240.

My plan was:
- pick a hybrid cmesh
- build a uniform forest
- iterate once over all trees in the forest
- for each tree record its eclass
- in the actual test, run over all trees, elements, faces
- get the neighboring tree in the forest
- get the eclass of this tree
- and compare to the output of `t8_forest_element_neighbor_eclass`

I had to adjust this a bit and now record the and compare to eclasses of all cmesh trees. (I was not able to move form cmesh trees to forest trees because some of them were not forest local).

#### General
#### Tests

#### Scripts and Wiki
#### License


benegee:
Not a brilliant test because the test code is actually quite close to the code being tested. Anyways, this is a start.

I think it might be useful to have a partitioned cmesh, but I was not able to produce one so far.

test/t8_forest/t8_gtest_element_neighbor_eclass.cxx
```diff
@@ -0,0 +1,128 @@
+/*
+  This file is part of t8code.
+  t8code is a C library to manage a collection (a forest) of multiple
+  connected adaptive space-trees of general element classes in parallel.
+
+  Copyright (C) 2015 the developers
```
Davknapp:
```suggestion
  Copyright (C) 2026 the developers
```

test/t8_forest/t8_gtest_element_neighbor_eclass.cxx
```diff
@@ -0,0 +1,128 @@
...
+          else {
+            EXPECT_EQ (neigh_class, tree_eclass_ref[cmesh_dual_itree]);
+          }
+        }
+      }
+    }
+  }
+}
+
+INSTANTIATE_TEST_SUITE_P (t8_gtest_element_neighbor_eclass, element_neighbor_eclass, AllSchemeCollections);
```
Davknapp: we should parametrize it over some levels too, not only level 1. 
Can we also start testing on level 0 and use a mesh with multiple trees?.

test/t8_forest/t8_gtest_element_neighbor_eclass.cxx
```diff
@@ -0,0 +1,128 @@
...
+          const int tree_face = scheme->element_get_tree_face (tree_eclass, element, iface);
+          t8_locidx_t const cmesh_dual_itree
+            = t8_cmesh_get_face_neighbor (cmesh, ltreeid_in_cmesh, tree_face, nullptr, nullptr);
+
+          if (cmesh_dual_itree < 0) {
+            // No neighbor
+            EXPECT_EQ (neigh_class, T8_ECLASS_INVALID);
+          }
+          else {
+            EXPECT_EQ (neigh_class, tree_eclass_ref[cmesh_dual_itree]);
```
Davknapp: I am not too happy with the tree_eclass_ref-vector here. I am not sure if this is the way neighbor-information should be handled, as it causes additional memory in the size of of O(trees). I understand that is is possibly cheaper (in runtime) to use this approach, which is why I think it is ok to use it for testing. 
however we should know that our tests can be used as examples to users, therefore we should at least add a comment that a decision about the tradeoff between memory/runtime has been made here.

Davknapp (changes requested):

test/t8_forest/t8_gtest_element_neighbor_eclass.cxx
benegee: Done! Now running over levels 0 to 4 (because I saw it somewhere else)

benegee: I picked `t8_cmesh_new_full_hybrid` because I wanted to cover as many element types as possible.
I thought I would already get multiple trees then, but I do not really know.

test/t8_forest/t8_gtest_element_neighbor_eclass.cxx
benegee: Agreed! This is not something anyone should consider a how to. If you want to query eclasses of your neighbors you should use `t8_forest_element_neighbor_eclass`.

In fact, I wanted to write a consistency check, which somehow fiddles out the same information as provided by `t8_forest_element_neighbor_eclass`, but in a different way. I failed in doing so. If you replace line 120 by a call to `t8_cmesh_get_tree_class`, you get essentially what it currently done inside `t8_forest_element_neighbor_eclass`.

test/t8_forest/t8_gtest_element_neighbor_eclass.cxx
Davknapp: Yes, my bad, that covers a couple of trees.

test/t8_forest/t8_gtest_element_neighbor_eclass.cxx
benegee: That being said, I now removed the ref arrays, as there were not useful.

In doing so I realized I had actually not considered ghost trees, i.e. I iterated over all local and ghost trees, but always used `t8_cmesh_get_tree_class`, never `t8_cmesh_get_ghost_class`. I now do it hopefully correctly 
https://github.com/DLR-AMR/t8code/blob/36f516f2be93f18b951753e5e00053ef418d1dd6/test/t8_forest/t8_gtest_element_neighbor_eclass.cxx#L110-L111
but it makes me wonder why it had worked before.

test/t8_forest/t8_gtest_element_neighbor_eclass.cxx
```diff
@@ -0,0 +1,121 @@
...
+        // Inside the current tree
+        if (!scheme->element_is_root_boundary (tree_eclass, element, iface)) {
+          // We expect to get the same eclass
+          EXPECT_EQ (neigh_class, tree_eclass);
+        }
+        // Crossing a tree boundary
+        else {
+          // Get neighbor tree id
+          const int tree_face = scheme->element_get_tree_face (tree_eclass, element, iface);
+          t8_locidx_t const neighbor_id
```
Davknapp:
```suggestion
        const t8_locidx_t neighbor_id
```
we should be consistent about the placement of const. Or is there a reason for your approach?

Davknapp (changes requested):

test/t8_forest/t8_gtest_element_neighbor_eclass.cxx
Davknapp: I don't know why it had worked before.

test/t8_forest/t8_gtest_element_neighbor_eclass.cxx
benegee: Indeed thanks!

benegee:
Thanks for looking into this @Davknapp!
I now do not store `cmesh` and `scheme` in the testing struct anymore (which would have required some refs).

Remaining question would be how the cmesh can initially be partitioned.

PR #2280, opened 2026-04-17
Fix PFC on root level elements

### LINKED ISSUE #2281: Fix PFC on root level elements

Niklas997:
Closes #2281

When partition for coarsening is applied and the border element is a root level element, a PFC-Message is constructed and the parent element therein is the actual pointer to the root element in the mesh. Upon destruction of the PFC-Message, the parent element (in this case  the actual root element in the mesh) is destroyed. In all other cases (i.e. the border element is not a root element) additional storage for the parent is allocated and it is filled with the corresponding information and this newly constructed element will be deallocated within the destructor. Currently, this leads to a crash in case the partition boundary falls on a root element. In order to account for this, there needs to be an additional element-allocation for the parent in the root level case as well which is then filled by an element_copy of the actual root element, such that it can be safely dealloacted upon destruction of the mesasge.

In order to re-create the issue, one can just construct a uniform forest with refinement level 0 from a cmesh and call partiton for caoarsening on it.

#### General
#### Tests

#### Scripts and Wiki
#### License


src/t8_forest/t8_forest_pfc_message.hxx
```diff
@@ -214,7 +214,11 @@ struct t8_forest_pfc_message
 
     // If we are already the root element, we cannot be part of a split family, so we send any(the root) element and no num_siblings.
     if (scheme->element_get_level (eclass, element_closest_to_receiver) == 0) {
-      parent = element_closest_to_receiver;
+      // Allocate memory for the parent element.
+      t8_element_new (scheme, eclass, 1, &parent);
```
spenke91, line 218: Would you mind moving this part before the if statement (and deleting line 226) to avoid code duplication?

spenke91 (changes requested):
Thanks a lot for catching this! :-) I just have one minor suggestion, hope you don't mind...

src/t8_forest/t8_forest_pfc_message.hxx
Niklas997: Yes, we can do this. Do you think we need to add an additional test or include this as a test case in the PFC test? Since this root level case is not yet captured in the code coverage, the codecov/patch test in this PR fails or is the quick fix enough for now?

spenke91: I'm sorry, but it would be much cleaner to test it. The easiest (though not the most runtime-efficient) option might be to add the `level` as a test parameter in `test/t8_forest/t8_gtest_partition_for_coarsening.cxx` , e.g., like this:

```
INSTANTIATE_TEST_SUITE_P (t8_gtest_partition_for_coarsening, t8_test_partition_for_coarsening_test,
                          testing::Combine (AllSchemeCollections, AllCmeshsParam, testing::Values(0,2)));
```
And then of course adjust the `std::tuple` and read the level from `std::get<2> (GetParam ())` instead of hard-coding it to 2...

This way, we would test all cmeshes with level 0 and level 2 (because as you said, running with level 0 causes exatcly this issue). If this gets too expensive, we might also change it later e.g. to using level 1 instead of two for TEST_LEVEL_BASIC. But the important thing for me right now would be to cover this issue you found 👍 

Would you mind to / know how to adjust the test case? 🤔

Niklas997: Okay, thank you for the hints to do so. Yes, I will try to adapt the test and add it to this PR.

spenke91: Thank you! :-)

Niklas997: Sorry for the delay! I hope I have accounted for the requested changes :)

spenke91 (approved):
Thanks!

PR #1722, opened 2025-06-27
Improvement: Implement FetchContent

ole-alb:
Closes #1723

**_Describe your changes here:_**

This PR changes the way how we integrate our submodules. We switch from git submodule to cmake's FetchContent. With this sc and p4est are automatically downloaded when cmake is run the first time. 

UPDATE: I also changed that creation of p4est. The release installation of p4est did not link against the release version of sc. I changed it, such that the installation of p4est in release mode links against sc in release mode. 

ToDo:
- [ ] Update installation guide 

**_All these boxes must be checked by the AUTHOR before requesting review:_**
- [x] The PR is *small enough* to be reviewed easily. If not, consider splitting up the changes in multiple PRs.
- [x] The title starts with one of the following prefixes: `Documentation:`, `Bugfix:`, `Feature:`, `Improvement:` or `Other:`.
- [x] If the PR is related to an issue, make sure to link it.
- [x] The author made sure that, as a reviewer, he/she would check all boxes below.

**_All these boxes must be checked by the REVIEWERS before merging the pull request:_**

As a reviewer please read through all the code lines and make sure that the code is *fully understood, bug free, well-documented and well-structured*.
#### General
- [ ] The reviewer executed the new code features at least once and checked the results manually.
- [ ] The code follows the [t8code coding guidelines](https://github.com/DLR-AMR/t8code/wiki/Coding-Guideline).
- [ ] New source/header files are properly added to the CMake files.
- [ ] The code is well documented. In particular, all function declarations, structs/classes and their members have a proper doxygen documentation.
- [ ] All new algorithms and data structures are sufficiently optimal in terms of memory and runtime (If this should be merged, but there is still potential for optimization, create a new issue).
#### Tests
- [ ] The code is covered in an existing or new test case using Google Test.
- [ ] The code coverage of the project (reported in the CI) should not decrease. If coverage is decreased, make sure that this is reasonable and acceptable.
- [ ] Valgrind doesn't find any bugs in the new code. [This script](https://github.com/DLR-AMR/t8code/blob/main/scripts/check_valgrind.sh) can be used to check for errors; see also this [wiki article](https://github.com/DLR-AMR/t8code/wiki/Debugging-with-valgrind).

If the Pull request introduces code that is not covered by the github action (for example coupling with a new library):
  - [ ] Should this use case be added to the github action?
  - [ ] If not, does the specific use case compile and all tests pass (check manually).
#### Scripts and Wiki
[...]

PR #2289, opened 2026-04-27
Revert "Improvement: Implement FetchContent"

sandro-elsweijer:
Reverts DLR-AMR/t8code#1722 due to multiple errors in the PR like:
- NetCDF was reintroduced
- Cache does not work
- General inconsistencies


spenke91 (approved):
Thanks!

Issue #2361, opened 2026-07-06
post-installation sanity test with pre-installed minimal build scripts


michelemartone:
# Feature idea 

Currently there is no trivial way to check the sanity of an installation of t8code.
Checking that is currently a manual task.
As such, it may cost time to detect: more to a new user, perhaps less to t8code developers.

A post-installation testing procedure, no matter how minimal, could catch some installation problems early.
One may install minimal `Makefile`s, ` CMakeLists.txt` and/or scripts first, and complement that with a thin `make post_install_tests` target invoking that.
These files, pre-installed in the directories where documentation lays, would also help the user to write own `t8code`-based programs -- which helps the adoption of `t8code`.

About alternatives:
Documentation mentioning how to build a test program is surely useful.
But it does not replace a readily usable post-installation procedure provided by the developers, perhaps turning on/off features as to reflect the build-time configuration of the library.

The feature request has little priority.
It clusters thematically to #2225 .

PR #2362, opened 2026-07-06
add post-installation test (answer issue #2361)

### LINKED ISSUE #2363: add post-installation test (answer issue #2361)

michelemartone:
Closes #2363

Address issue #2361 by adding a post-installation test goal (`post_install_tests`).

This works by:
 - generating a top-level CMakeLists.txt file with hardcoded paths reflecting the installation directories and installing it in a directory with documentation
 - installing two simple CMakeLists.txt files meant as t8code use examples; the two are located in respective sub-directories, both containing the same sample t8code program (the first tutorial program actually)
 - ensuring that the top-level CMakeLists.txt file is be usable from anywhere without further configuration, and uses the two CMakeLists.txt files to build two executables and run them (via the `tests` target)
 - having a `post_install_tests` target, to build the above two programs in the temporary `post_install_tests_dir` directory for test purposes
 - assuming the `post_install_tests` target to only work after `install`
 - keeping the test to be basic - the goal is to have detection of the t8code library to work, as well as build and execution of the two test programs

Current shortcomings:
 - test limited to Linux
 - not documented anywhere ( ideal would be in this files' tree, but current t8code practice uses t8code.wiki, which is separate )
 - it may assume that the two MPI-enabled executables can be ran straight-ahead
 - installation happens in a documentation directory, but to be enabled, this feature relies on T8CODE_BUILD_TUTORIALS, not T8CODE_BUILD_DOCUMENTATION

p.s.: Since this is my first contribution to t8code, I'm including my
      authorization to treat this contribution under the "FreeBSD License"
      and I'm signing (at least) this commit with my usual GPG key.

#### General
#### Tests

#### Scripts and Wiki
#### License

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -281,6 +281,17 @@ endif()
 
 if ( T8CODE_BUILD_TUTORIALS )
     add_subdirectory( ${CMAKE_CURRENT_LIST_DIR}/tutorials )
+    if ( CMAKE_SYSTEM_NAME MATCHES "Linux" ) # post-install tests currently limited to our main development OS
+        add_custom_target(post_install_tests # notice that add_test() would not be appropriate for this test
+            COMMAND rm -fR post_install_tests_dir # we perform the test out of this cmake tree
+            COMMAND mkdir post_install_tests_dir
+            COMMAND cd post_install_tests_dir && cmake ${CMAKE_INSTALL_PREFIX}/share/doc/T8CODE/cmake_examples/ && make VERBOSE=1 && make VERBOSE=1 tests
+            COMMAND rm -fR post_install_tests_dir
+            WORKING_DIRECTORY ${PROJECT_BINARY_DIR}
+            VERBATIM
+            COMMENT "Post-install tests' target. It creates, uses, and deletes temporary directory 'post_install_tests_dir'. It assumes running under Linux."
+        )
+    endif()
 endif()
 
 if ( T8CODE_BUILD_EXAMPLES )
--- a/doc/cmake_examples/a/CMakeLists.txt
+++ b/doc/cmake_examples/a/CMakeLists.txt
@@ -1,0 +1,29 @@
+#  This file is part of t8code.
+#  t8code is a C library to manage a collection (a forest) of multiple
+#  connected adaptive space-trees of general element types in parallel.
+#
+#  Copyright (C) 2026 the developers
+#
+#  t8code is free software; you can redistribute it and/or modify
+#  it under the terms of the GNU General Public License as published by
+#  the Free Software Foundation; either version 2 of the License, or
+#  (at your option) any later version.
+#
+#  t8code is distributed in the hope that it will be useful,
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#  GNU General Public License for more details.
+#
+#  You should have received a copy of the GNU General Public License
+#  along with t8code; if not, write to the Free Software Foundation, Inc.,
+#  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+
+# A CMakeLists.txt example file showing a way of detecting an installed instance of T8CODE and using it in program test.cxx.
+# This example relies on what is being declared in files installed by T8CODE.
+
+cmake_minimum_required(VERSION 3.22.0)
+project(t8test DESCRIPTION "t8code test program (requires CMAKE_PREFIX_PATH to point to T8CODE's package files are)" LANGUAGES CXX)
+find_package(T8CODE)
+add_executable(test_a test.cxx)
+target_link_libraries(test_a PRIVATE T8CODE::T8)
+add_custom_target(tests_a COMMAND ./test_a)
--- a/doc/cmake_examples/b/CMakeLists.txt
+++ b/doc/cmake_examples/b/CMakeLists.txt
@@ -1,0 +1,43 @@
+#  This file is part of t8code.
+#  t8code is a C library to manage a collection (a forest) of multiple
+#  connected adaptive space-trees of general element types in parallel.
+#
+#  Copyright (C) 2026 the developers
+#
+#  t8code is free software; you can redistribute it and/or modify
+#  it under the terms of the GNU General Public License as published by
+#  the Free Software Foundation; either version 2 of the License, or
+#  (at your option) any later version.
+#
+#  t8code is distributed in the hope that it will be useful,
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#  GNU General Public License for more details.
+#
+#  You should have received a copy of the GNU General Public License
+#  along with t8code; if not, write to the Free Software Foundation, Inc.,
+#  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+
+# A CMakeLists.txt file showing a way of detecting an installed instance of T8CODE for use in program test.cxx.
+# This example does not rely on what is being declared in files installed by T8CODE.
+
+cmake_minimum_required(VERSION 3.22)
+include(CheckIncludeFileCXX)
+include(CheckLibraryExists)
+project(t8test DESCRIPTION "t8code test program (needs T8TEST_ROOT to contain relevant include and lib directories and CMAKE_PREFIX_PATH to point to T8CODE's package files are.)" LANGUAGES CXX)
+option(T8TEST_ROOT "Custom directory containing t8code's include, lib, ..." "")
+find_library(T8CODE_LIBRARY t8 REQUIRED) # one may have done this in the if conditional construct after check_library_exists
+CHECK_LIBRARY_EXISTS(t8 t8_init "${T8TEST_ROOT}/lib" HAVE_T8) # check_library_exists looks for a library and sets one variable accordingly
+if(NOT HAVE_T8)
+    message(FATAL_ERROR "no t8 library found!")
+else()
+    message("t8 library found in ${T8TEST_ROOT}/lib")
+endif()
+find_path(SC_INCLUDE_DIR sc.h REQUIRED)
+find_path(T8CODE_INCLUDE_DIR t8.h REQUIRED)
+add_executable(test_b test.cxx)
+target_include_directories(test_b PUBLIC "${SC_INCLUDE_DIR}" )
+target_include_directories(test_b PUBLIC "${T8CODE_INCLUDE_DIR}" )
+target_link_directories(test_b PUBLIC "${T8TEST_ROOT}/lib" )
+target_link_libraries(test_b PRIVATE t8 sc)
+add_custom_target(tests_b COMMAND ./test_b)
--- a/tutorials/CMakeLists.txt
+++ b/tutorials/CMakeLists.txt
@@ -46,6 +46,7 @@ function( add_t8_tutorial )
     endif( T8CODE_EXPORT_COMPILE_COMMANDS )
 
     install( TARGETS ${ADD_T8_TUTORIAL_NAME} RUNTIME )
+    install( FILES ${ADD_T8_TUTORIAL_SOURCES} DESTINATION ${CMAKE_INSTALL_DOCDIR}/tutorials/general/ )
 endfunction()
 
 function( add_mesh_handle_tutorial )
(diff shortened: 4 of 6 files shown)


doc/cmake_examples/a/CMakeLists.txt
```diff
@@ -0,0 +1,29 @@
...
+#  along with t8code; if not, write to the Free Software Foundation, Inc.,
+#  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+
+# A CMakeLists.txt example file showing a way of detecting an installed instance of T8CODE and using it in program test.cxx.
+# This example relies on what is being declared in files installed by T8CODE.
+
+cmake_minimum_required(VERSION 3.22.0)
+project(t8test DESCRIPTION "t8code test program (requires CMAKE_PREFIX_PATH to point to T8CODE's package files are)" LANGUAGES CXX)
+find_package(T8CODE)
+add_executable(test_a test.cxx)
```
sandro-elsweijer, line 27: Can you please avoid using single letter file names, variable names, etc.?

doc/cmake_examples/b/CMakeLists.txt
```diff
@@ -0,0 +1,43 @@
...
+find_library(T8CODE_LIBRARY t8 REQUIRED) # one may have done this in the if conditional construct after check_library_exists
+CHECK_LIBRARY_EXISTS(t8 t8_init "${T8TEST_ROOT}/lib" HAVE_T8) # check_library_exists looks for a library and sets one variable accordingly
+if(NOT HAVE_T8)
+    message(FATAL_ERROR "no t8 library found!")
+else()
+    message("t8 library found in ${T8TEST_ROOT}/lib")
+endif()
+find_path(SC_INCLUDE_DIR sc.h REQUIRED)
+find_path(T8CODE_INCLUDE_DIR t8.h REQUIRED)
+add_executable(test_b test.cxx)
```
sandro-elsweijer, line 38: same here

tutorials/CMakeLists.txt
```diff
@@ -46,6 +46,7 @@ function( add_t8_tutorial )
     endif( T8CODE_EXPORT_COMPILE_COMMANDS )
 
     install( TARGETS ${ADD_T8_TUTORIAL_NAME} RUNTIME )
+    install( FILES ${ADD_T8_TUTORIAL_SOURCES} DESTINATION ${CMAKE_INSTALL_DOCDIR}/tutorials/general/ )
```
sandro-elsweijer, line 49: Can you please add a new configuration argument to do this and place this in its own cmake file inside the test folder? We want to keep all directories true to their purpose and the tutorial folder is just for tutorials. Same for the documentation folder.

tutorials/CMakeLists.txt
```diff
@@ -81,6 +82,15 @@ copy_tutorial_file (features/t8_features_curved_meshes_generate_cmesh_quad.geo)
...
 copy_tutorial_file (features/t8_features_curved_meshes_generate_cmesh_tri.geo)
 
+# notice that in the following, we install files from doc/ and tutorials/ together:
+file(GLOB T8CODE_CMAKE_EXAMPLES RELATIVE "${PROJECT_SOURCE_DIR}/doc/cmake_examples/" "${PROJECT_SOURCE_DIR}/doc/cmake_examples/?") # for each single-character directory; notice there's no explicit listing here
+configure_file( ${PROJECT_SOURCE_DIR}/doc/cmake_examples/CMakeLists.txt.in ${CMAKE_BINARY_DIR}/doc/cmake_examples/CMakeLists.txt @ONLY)
+install(FILES ${CMAKE_BINARY_DIR}/doc/cmake_examples/CMakeLists.txt DESTINATION ${CMAKE_INSTALL_DOCDIR}/cmake_examples/ RENAME CMakeLists.txt) # reminder that we may want to name this differently (yet, no necessity for .in)
+foreach(_dir IN LISTS T8CODE_CMAKE_EXAMPLES)
+     install(DIRECTORY ${PROJECT_SOURCE_DIR}/doc/cmake_examples/${_dir} DESTINATION ${CMAKE_INSTALL_DOCDIR}/cmake_examples/) # notice we take directories from doc...
+     install(FILES ${PROJECT_SOURCE_DIR}/tutorials/general/t8_step0_helloworld.cxx DESTINATION ${CMAKE_INSTALL_DOCDIR}/cmake_examples/${_dir}/ RENAME test.cxx) #... and add the same sample tutorial program as a test
+endforeach()
```
sandro-elsweijer, line 92: You can happily re-use the tutorial files as test source code, but the actual testing should stay outside the tutorial folder

CMakeLists.txt
```diff
@@ -281,6 +281,17 @@ endif()
...
+        add_custom_target(post_install_tests # notice that add_test() would not be appropriate for this test
+            COMMAND rm -fR post_install_tests_dir # we perform the test out of this cmake tree
+            COMMAND mkdir post_install_tests_dir
+            COMMAND cd post_install_tests_dir && cmake ${CMAKE_INSTALL_PREFIX}/share/doc/T8CODE/cmake_examples/ && make VERBOSE=1 && make VERBOSE=1 tests
+            COMMAND rm -fR post_install_tests_dir
+            WORKING_DIRECTORY ${PROJECT_BINARY_DIR}
+            VERBATIM
+            COMMENT "Post-install tests' target. It creates, uses, and deletes temporary directory 'post_install_tests_dir'. It assumes running under Linux."
+        )
+    endif()
```
sandro-elsweijer, line 294: I tried to translate this to cmake commands as good as possible. This way cmake peserves the used generator, build type, etc. and it is also OS agnostic.
The `BINARY_DIR` could be extracted from the (`PROJECT_BINARY_DIR`)[PROJECT_BINARY_DIR] cmake variable and passed to the test in the CMakeLists calling this build script.

```suggestion
# Remove any previous build directory.
file(REMOVE_RECURSE "${BINARY_DIR}/post_install_tests")
file(MAKE_DIRECTORY "${BINARY_DIR}")

set(EXAMPLE_DIR
    "${INSTALL_PREFIX}/path_to_dir"
)

# Run configure step
execute_process(
    COMMAND
        ${CMAKE_COMMAND}
        -S "${EXAMPLE_DIR}"
        -B "${BINARY_DIR/post_install_tests}"
    RESULT_VARIABLE result
)

if(NOT result EQUAL 0)
    message(FATAL_ERROR "Failed to configure installed example.")
endif()

# Run  build step
execute_process(
    COMMAND
        ${CMAKE_COMMAND}
        --build "${BINARY_DIR/post_install_tests}"
    RESULT_VARIABLE result
)

if(NOT result EQUAL 0)
    message(FATAL_ERROR "Failed to build installed example.")
endif()

# Run tests
execute_process(
    COMMAND
        ${CMAKE_COMMAND}
        --build "${BINARY_DIR/post_install_tests}"
        --target tests
    RESULT_VARIABLE result
)

if(NOT result EQUAL 0)
    message(FATAL_ERROR "Installed example tests failed.")
endif()

file(REMOVE_RECURSE "${BINARY_DIR/post_install_tests}")
```

sandro-elsweijer (changes requested):
Thanks for the contribution, this is a good and useful addition to our testing pipeline, especially for testing the validity of the installed packages.

For merging, we’ll need it a bit more aligned with our coding guidelines: clearer structure (one responsibility per folder), more descriptive naming instead of single-letter identifiers, and a more CMake-native approach instead of shell/system-specific assumptions. Also, hardcoding should really be avoided unless there’s no reasonable alternative.

With those adjustments, this looks ready to go.

CMakeLists.txt
michelemartone: You removed the `post_install_tests` target along the way. We want to have that.

tutorials/CMakeLists.txt
michelemartone: I don't understand your comment. Testing happens anyway in a `post_install_tests_dir` directory.

tutorials/CMakeLists.txt
michelemartone: Why are you suggesting a new configuration variable? `test/CMakeLists.txt` depends on `T8CODE_BUILD_TESTS`, which is an already existing configuration variable.

doc/cmake_examples/b/CMakeLists.txt
michelemartone: It'd be good to have the directories under `doc/cmake_examples` easily enumerable via globbing from elsewhere -- i.e. with no need to know their identifier in other places. The current scheme was "single letter". I could introduce a prefix (e.g. `example_mini` and `example_midi`); that would be an epsilon more informative (not worth the cognitive effort IMO) but would be introducing redundancy (which I'm striving to reduce). Would you still prefer that way (say `example_mini` and `example_midi`) ?

tutorials/CMakeLists.txt
spenke91: I think he meant that it should not be installed in ${ADD_T8_TUTORIAL_SOURCES}.

doc/cmake_examples/b/CMakeLists.txt
sandro-elsweijer: What about `post_installation_check_1` and `post_installation_check_2`?
There is nothing against using enumerations of any sort, but `test_b` and `test.cxx` are not really descriptive and it could mean anything.

tutorials/CMakeLists.txt
sandro-elsweijer: No, I meant that this is a post installation test and that it does not have anything to do with the tutorials. Yes, it uses a tutorial source file as test subject, but as a user i would not expect a post installation test in the tutorial folder. I would expect this to be in the test folder.

Also, I would not expect this post installation test to be dependent on the `T8CODE_BUILD_TUTORIALS` option since, as i already explained, it is a test.

tutorials/CMakeLists.txt
sandro-elsweijer: I meant, that the main purpose of this whole PR is to introduce a software test. Therefore, everything should be located in the `test` folder. The PR is not meant as documentation and also not as a tutorial. Yes, users could look at this to inspire them, but the main purpose is still that this is a software test.

CMakeLists.txt
sandro-elsweijer: Ah sorry, I added it back

tutorials/CMakeLists.txt
spenke91: I agree wih @sandro-elsweijer here. I think any post-install test should be logically separated from the tutorials. Yes, it currently uses the tutorial as test, but that is just a more or less arbitrary choice we made (for now).

tutorials/CMakeLists.txt
spenke91: Again, I am with @sandro-elsweijer on this one. We should separate the testing from the tutorials in a clear/clean way.

doc/cmake_examples/b/CMakeLists.txt
spenke91: > What about `post_installation_check_1` and `post_installation_check_2`? There is nothing against using enumerations of any sort, but `test_b` and `test.cxx` are not really descriptive and it could mean anything.

I agree and upvote this reply ☝️

CMakeLists.txt
```diff
@@ -281,6 +281,17 @@ endif()
 
 if ( T8CODE_BUILD_TUTORIALS )
     add_subdirectory( ${CMAKE_CURRENT_LIST_DIR}/tutorials )
+    if ( CMAKE_SYSTEM_NAME MATCHES "Linux" ) # post-install tests currently limited to our main development OS
```
spenke91, line 284: I know we briefly talked about it, but I (still?) don't like this mixture of purposes for the tutorials and the post-install check. I think the post-install testing should be completely separated from the tutorials, even if it might "happen to" use the same code snippets.

CMakeLists.txt
spenke91: I am not entirely sure about the CMake snippets @sandro-elsweijer suggested (e.g., do we have to use "COMMAND" for CMake tasks like that?), but I agree that we should aim to handle as much as possible in CMake and not just use CMake as a means of calling shell commands.

CMakeLists.txt
```diff
@@ -281,6 +281,17 @@ endif()
...
+        add_custom_target(post_install_tests # notice that add_test() would not be appropriate for this test
```
spenke91, line 285: similar to above, I think the folder `post_install_tests` should not be inside ´tutorials´

doc/cmake_examples/a/CMakeLists.txt
spenke91: > Can you please avoid using single letter file names, variable names, etc.?

👍

doc/cmake_examples/b/CMakeLists.txt
```diff
@@ -0,0 +1,43 @@
...
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#  GNU General Public License for more details.
+#
+#  You should have received a copy of the GNU General Public License
+#  along with t8code; if not, write to the Free Software Foundation, Inc.,
+#  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+
+# A CMakeLists.txt file showing a way of detecting an installed instance of T8CODE for use in program test.cxx.
+# This example does not rely on what is being declared in files installed by T8CODE.
```
spenke91, line 22: I would maybe mention this test, unlike the other, does not use the CMake magic of `find_package` :-)

spenke91 (changes requested):
Thanks @michelemartone for your contribution! 👍
I am happy to see us getting closer to having post-install tests ;-)
I just have some remarks, which mostly align with those of @sandro-elsweijer

PR #2395, opened 2026-08-13
Feature: Add subelement scheme for quad and triangle hanging node resolution

### LINKED ISSUE #2396: Feature: Add subelement scheme for quad and triangle hanging node resolution

lenaploetzke:
Closes #2396

This PR adds a general subelement scheme with specializations for quad and triangle hanging node resolution. It also provides two examples and one test. 
I implemented only the functions necessary for the visualization. Next steps are all functions necessary for simulations. This means for example neighbors. 

#### General
#### Tests

#### Scripts and Wiki
#### License


src/t8_schemes/t8_scheme.cxx
```diff
@@ -24,10 +24,12 @@
  * Implements functions declared in \ref t8_scheme.h.
  */
 
+#include "t8.h"
```
sandro-elsweijer: the file is not in the same folder
```suggestion
#include <t8.h>
```

test/t8_gtest_adapt_callbacks.cxx
```diff
@@ -56,3 +56,15 @@ t8_test_adapt_first_child (t8_forest_t forest, [[maybe_unused]] t8_forest_t fore
...
+int
+t8_test_adapt_even_global_id ([[maybe_unused]] t8_forest_t forest, t8_forest_t forest_from, t8_locidx_t which_tree,
+                              [[maybe_unused]] t8_eclass_t eclass, t8_locidx_t lelement_id,
+                              [[maybe_unused]] const t8_scheme *scheme, [[maybe_unused]] const int is_family,
+                              [[maybe_unused]] const int num_elements, [[maybe_unused]] t8_element_t *elements[])
+{
+  if ((t8_forest_get_tree_element_offset (forest_from, which_tree) + lelement_id) % 2 == 0) {
+    return 1;
+  }
+  return 0;
```
spenke91, line 69: I think this is actually the local element ID, not the global one. It relies on the tree's member variable
`elements_offset`, which is process-local, see here:

Image

test/t8_gtest_adapt_callbacks.cxx
```diff
@@ -56,3 +56,15 @@ t8_test_adapt_first_child (t8_forest_t forest, [[maybe_unused]] t8_forest_t fore
   }
   return 0;
 }
+
+int
+t8_test_adapt_even_global_id ([[maybe_unused]] t8_forest_t forest, t8_forest_t forest_from, t8_locidx_t which_tree,
```
spenke91, line 61:
```suggestion
t8_test_adapt_even_local_id ([[maybe_unused]] t8_forest_t forest, t8_forest_t forest_from, t8_locidx_t which_tree,
```

test/t8_gtest_adapt_callbacks.hxx
```diff
@@ -53,4 +53,22 @@ t8_test_adapt_first_child (t8_forest_t forest, t8_forest_t forest_from, t8_locid
...
+ * \param [in] which_tree   The local tree containing \a elements.
+ * \param [in] eclass       The eclass of \a which_tree.
+ * \param [in] lelement_id  The local element id in \a forest_from in the tree of the current element.
+ * \param [in] scheme       The scheme of the forest.
+ * \param [in] is_family    If 1, the first \a num_elements entries in \a elements form a family. If 0, they do not.
+ * \param [in] num_elements The number of entries in \a elements that are defined
+ * \param [in] elements     Pointers to a family or, if \a is_family is zero, pointer to one element.
+ */
+int
+t8_test_adapt_even_global_id (t8_forest_t forest, t8_forest_t forest_from, t8_locidx_t which_tree, t8_eclass_t eclass,
```
spenke91, line 70:
```suggestion
t8_test_adapt_even_local_id (t8_forest_t forest, t8_forest_t forest_from, t8_locidx_t which_tree, t8_eclass_t eclass,
```

spenke91: I'd suggest to use the callback `refine_every_nth_element_callback` I recently added 🙂 
(which is adjusted to global elements in #2415 )

src/t8_schemes/t8_subelement/README.md
```diff
@@ -0,0 +1,25 @@
+# t8_subelement
+
+This folder provides **subelement schemes**. Subelements are inserted *after* the standard recursive refinement and enable one additional refinement level that uses a different scheme.
```
spenke91, line 3:
```suggestion
This folder provides **subelement schemes**. Subelements are inserted *after* the standard recursive refinement and enable one additional refinement step that uses a different scheme.
```
I think we should not call it level to avoid confusion :thinking:

src/t8_schemes/t8_subelement/t8_subelement_scheme.hxx
```diff
@@ -0,0 +1,1259 @@
...
+  /** Constructor. */
+  t8_subelement_scheme_common () noexcept
+    : element_size (sizeof (TSubelementType)), scheme_context (sc_mempool_new (element_size)) {};
+
+ protected:
+  size_t element_size;  /**< The size in bytes of an element. */
+  void *scheme_context; /**< Anonymous implementation context. */
+
+ public:
+  // #################################____Constructor & Destructor...____###############################################
```
spenke91, line 64: Is there a specific reason the constructor above is not sorted into this category?

src/t8_schemes/t8_subelement/t8_subelement_scheme.hxx
```diff
@@ -0,0 +1,1259 @@
...
+    return *this;
+  }
+
+  // ################################################____GENERAL INFO____###############################################
+
+  /** Return the size of any element.
+   * \return The size of an element.
+   */
+  static constexpr size_t
+  get_element_size (void) noexcept
```
spenke91, line 123: I think we do not need the `void` argument in the C++ part, do we? If not, it also applies to all `(void)` arguments below, of course

src/t8_schemes/t8_subelement/t8_subelement_scheme.hxx
```diff
@@ -0,0 +1,1259 @@
...
+    * \return             The level of \b elem.
+    */
+  int
+  element_get_level (const t8_element_t *elem) const noexcept
+  {
+    T8_ASSERT (element_is_valid (elem));
+    return derived ().underlying_scheme.element_get_level (element_to_standalone (elem));
+  }
+
+  // ################################################____GENERAL HELPER____#############################################
```
spenke91, line 255:
```suggestion
  // ################################################____GENERAL HELPERS____#############################################
```

src/t8_schemes/t8_subelement/t8_subelement_scheme.hxx
```diff
@@ -0,0 +1,1259 @@
...
+
+  /** Copy all entries of \b source to \b dest. \b dest must be an existing
+   *  element. No memory is allocated by this function.
+   * \param [in] source The element whose entries will be copied to \b dest.
+   * \param [in,out] dest This element's entries will be overwrite with the entries of \b source.
```
spenke91, line 260:
```suggestion
   * \param [in,out] dest This element's entries will be overwritten by the entries of \b source.
```

src/t8_schemes/t8_subelement/t8_subelement_scheme.hxx
```diff
@@ -0,0 +1,1259 @@
...
+   * Non-const version.
+   * \param[in] element Element of class t8_element_t that can be interpreted as subelement. 
+   */
+  static TSubelementType *
+  as_subelement (t8_element_t *element) noexcept
+  {
+    return reinterpret_cast<TSubelementType *> (element);
+  }
+
+  /** Reset the subelement-specific data.
```
spenke91, line 1221: Do I understand correctly that this `reinterpret_cast` is fine because we know all `element` arguments in these calls will actually be pointers to `TSubelementType`, right?

src/t8_schemes/t8_subelement/t8_subelement_scheme.hxx
```diff
@@ -0,0 +1,1259 @@
...
+   * \param [in,out] subelement The element that is filled with the root of the subelement.
+   */
+  static void
+  reset_subelement_values (TSubelementType *subelement) noexcept
```
spenke91, line 1225: Should be call it `unset` or something to point out we are essentially "removing" the subelement information? `reset` might sound too much like we are actually setting some subelement information here. But maybe "unset" is not a good term either :-)

src/t8_schemes/t8_subelement/t8_subelement_scheme.hxx
```diff
@@ -0,0 +1,1259 @@
...
+  */
+  int
+  element_is_equal (const t8_element_t *elem1, const t8_element_t *elem2) const noexcept
+  {
+    T8_ASSERT (element_is_valid (elem1) && element_is_valid (elem2));
+    const auto *el1 = as_subelement (elem1);
+    const auto *el2 = as_subelement (elem2);
+    if (el1->subelement_type != el2->subelement_type) {
+      return 0;
+    }
```
spenke91, line 287: Probably a simple answer, but why don't we compare `subelement_id` here?

src/t8_schemes/t8_subelement/t8_subelement_scheme.hxx
```diff
@@ -0,0 +1,1259 @@
...
+
+  /** This defines how an element is refined in subelements using a specified subelement type. 
+   * \param [in] elem The element to be refined.
+   * \param [in] length   The length of the output array \a c must match the number of subelements.   
+   *                      See \ref element_get_num_children.
+   * \param [in, out] c An array of allocated elements that will be filled with the subelements of \a elem. 
+   * \param [in] type The subelement type to be used for refinement.
+   */
+  void
+  element_get_children (const t8_element_t *elem, const int length, t8_element_t *c[], int type) const noexcept
```
spenke91, line 455:
```suggestion
   * \param [in] sub_elem_type The subelement type to be used for refinement.
   */
  void
  element_get_children (const t8_element_t *elem, const int length, t8_element_t *c[], int sub_elem_type) const noexcept
```

src/t8_schemes/t8_subelement/t8_subelement_scheme.hxx
```diff
@@ -0,0 +1,1259 @@
...
+   * \param [out] desc    The last element in a uniform refinement of \a elem of the given level.
+   * \param [in] level    The level, at which the descendant is computed.
+   */
+  void
+  element_get_last_descendant (const t8_element_t *elem, t8_element_t *desc,
+                               const t8_element_level level) const noexcept
+  {
+    derived ().underlying_scheme.element_get_last_descendant (element_to_standalone (elem),
+                                                              element_to_standalone (desc), level);
+    reset_subelement_values ((TSubelementType *) desc);
```
spenke91, line 602: Why not like this
```suggestion
    reset_subelement_values (as_subelement(desc));
```
?

src/t8_schemes/t8_subelement/t8_subelement_scheme.hxx
```diff
@@ -0,0 +1,1259 @@
...
+   * \param [in] length   The number of elements to be deinitialized.
+   * \param [in,out] elems On input an array of \a length many allocated and initialized elements, on output an array of
+   *                       \a length many allocated, but not initialized elements.
+   * \note Call this function if you called element_init on the element pointers.
+   * \see element_init
+   */
+  static constexpr void
+  element_deinit ([[maybe_unused]] const int length, [[maybe_unused]] t8_element_t *elems) noexcept
+  {
+  }
```
spenke91, line 1004: Missing implementation?

src/t8_schemes/t8_subelement/specializations/t8_scheme_tri.hxx
```diff
@@ -0,0 +1,338 @@
...
+    return num_hanging_faces + 1;
+  }
+
+  /** This defines how an element is refined into subelements using a specified subelement type.
+   * \param [in] elem The element to be refined.
+   * \param [in] length   The length of the output array \a c must match the number of subelements.   
+   *                      See \ref element_get_num_children.
+   * \param [in, out] c An array of allocated elements that will be filled with the subelements of \a elem. 
+   * \param [in] type The subelement type to be used for refinement. This is a binary encoding of the hanging faces.
+   */
```
spenke91, line 155:
```suggestion
  /** This defines how an element is refined into subelements using a specified subelement type.
   * \param [in] elem     The element to be refined.
   * \param [in] length   The length of the output array \a c must match the number of subelements.   
   *                      See \ref element_get_num_children.
   * \param [in, out] c   An array of allocated elements that will be filled with the subelements of \a elem. 
   * \param [in] type     The subelement type to be used for refinement. This is a binary encoding of the hanging faces.
   */
```

src/t8_schemes/t8_subelement/specializations/t8_scheme_tri.hxx
```diff
@@ -0,0 +1,338 @@
...
+  void
+  subelement_get_children (const t8_element_t *elem, [[maybe_unused]] const int length, t8_element_t *c[],
+                           int type) const noexcept
+  {
+    const TSubelementType *element = this->as_subelement (elem);
+    TSubelementType **subelements = reinterpret_cast<TSubelementType **> (c);
+    const int num_subelements = this->subelement_get_num_children (elem, type);
```
spenke91, line 162: Should we use `length` here somehow or assert it matches `num_subelements`? If we really don't use `length` at all, we should maybe say so in the doxygen description.

src/t8_schemes/t8_subelement/specializations/t8_scheme_tri.hxx
```diff
@@ -0,0 +1,338 @@
...
+   *                      See \ref element_get_num_children.
+   * \param [in, out] c An array of allocated elements that will be filled with the subelements of \a elem. 
+   * \param [in] type The subelement type to be used for refinement. This is a binary encoding of the hanging faces.
+   */
+  void
+  subelement_get_children (const t8_element_t *elem, [[maybe_unused]] const int length, t8_element_t *c[],
+                           int type) const noexcept
+  {
+    const TSubelementType *element = this->as_subelement (elem);
+    TSubelementType **subelements = reinterpret_cast<TSubelementType **> (c);
```
spenke91, line 161: We have this "two-fold view" on element arrays as t8_element_t and subelements a few times. Should we make it more obvious they are pointing at the same memory, e.g., by calling it `c_as_subelements`?

src/t8_schemes/t8_subelement/specializations/t8_scheme_tri.hxx
```diff
@@ -0,0 +1,338 @@
...
+    }
+  }
+
+  /** Convert points in the reference space of a (triangular) subelement to points in the reference
+   * space of the tree.
+   * \param [in] elem       The subelement.
+   * \param [in] ref_coords The coordinates in \f$ [0,1]^2 \f$ of the points in the subelement's reference space.
+   * \param [in] num_coords The number of points to convert.
+   * \param [out] out_coords The coordinates of the points in the reference space of the tree.
+   */
```
spenke91, line 191:
```suggestion
  /** Convert points in the reference space of a (triangular) subelement to points in the reference
   * space of the tree.
   * \param [in]  elem        The subelement.
   * \param [in]  ref_coords  The coordinates in \f$ [0,1]^2 \f$ of the points in the subelement's reference space.
   * \param [in]  num_coords  The number of points to convert.
   * \param [out] out_coords  The coordinates of the points in the reference space of the tree.
   */
```

src/t8_schemes/t8_subelement/specializations/t8_scheme_tri.hxx
```diff
@@ -0,0 +1,338 @@
...
+  }
+
+ private:
+  /** Check whether a given face of the parent triangle is hanging.
+   * \param [in] type  The subelement type (binary code over the faces, f0 is the most significant bit).
+   * \param [in] iface The face to check.
+   * \return           True if \a iface is hanging for \a type.
+   */
+  static bool
+  face_is_hanging (const unsigned type, const int iface) noexcept
```
spenke91, line 224: maybe avoid `type` and call it `subelem_type` or something to avoid confusing it with the type of the triangle (0 or 1)

src/t8_schemes/t8_subelement/specializations/t8_scheme_tri.hxx
```diff
@@ -0,0 +1,338 @@
...
+    // If that bit is 1, the face is hanging.
+    // 1u is for lowest bit extraction.
+    return ((type >> ((T8_ELEMENT_NUM_FACES[T8_ECLASS_TRIANGLE] - 1) - iface)) & 1u) != 0u;
+  }
+
+  /** Compute the integer coordinates of the three vertices of a triangular subelement.
+   *
+   * For this, we first define the order of the subelements and the subelement vertices:
+   * All subelements of a transition cell share one common point, which we define as \a m_c, and each subelement is
+   * spanned by \a m_c together with two consecutive points of a \b path along the boundary of the
```
spenke91, line 236:
```suggestion
   * All subelements of a transition cell share one common vertex, which we define as \a m_c, and each subelement is
   * spanned by \a m_c together with two consecutive vertices of a \b path along the boundary of the
```

src/t8_schemes/t8_subelement/specializations/t8_scheme_tri.hxx
```diff
@@ -0,0 +1,338 @@
...
+   * parent triangle. This defines the numbering completely:
+   *   - The \a main \a face is the lowest-indexed hanging face \a fA.
```
spenke91, line 238: I think there is a slight convention or recommondation in doxygen to use \a for variables, arguments etc and \b for general emphasis. I don't know whether you want to stick to this, though. In case you do, maybe you can double-check the usage here.

src/t8_schemes/t8_subelement/specializations/t8_scheme_tri.hxx
```diff
@@ -0,0 +1,338 @@
...
+   *      vertex 0 = \a m_c, vertex 1 = path[ \a i ], vertex 2 = path[ \a i+1 ].
+   *
+   * Since the path has (number of hanging faces + 2) points, there are (number of hanging faces + 1)
+   * subelements, which matches \ref element_get_num_children. No case distinction is needed:
+   * one hanging face yields a path of three points, two hanging faces a path of four.
+   *
+   * \verbatim
+        f2 hanging                                f2 hanging
+        one hanging face (here f2)            two hanging faces (here f1 and f2) (not nicely displayed)
+
```
spenke91, line 255: Thanks for this beautiful ASCII art! :-)

src/t8_schemes/t8_subelement/specializations/t8_scheme_tri.hxx
```diff
@@ -0,0 +1,338 @@
...
+    return T8_TRI_MAX_SUBELEMENT_TYPE;
+  }
+
+  /** Get the number of subelements an element is refined into for a specific type.
+   * \param [in] elem   The element whose number of children is returned.
+   * \param [in] subelement_type The subelement type used for refinement.
+   * \return                     The number of subelements the triangle is split into (hanging faces + 1).
+   */
+  static int
+  subelement_get_num_children ([[maybe_unused]] const t8_element_t *elem, int subelement_type)
```
spenke91, line 139: Can't we make this a one-liner with `std::popcount` like you do further below?
Also, it might make sense to add a helper function `subelement_get_num_hanging_faces` or something like this (so far, we just use the computation twice, so I'd say its borderline, but if we use it more often one common helper would avoid code duplication and reduce the risk of typo errors)

src/t8_schemes/t8_subelement/specializations/t8_scheme_tri.hxx
```diff
@@ -0,0 +1,338 @@
...
+
+    /* The corners of the parent triangle. */
+    std::array<std::array<int, 2>, 3> parent_coords;
+    for (int icorner = 0; icorner < T8_ELEMENT_NUM_CORNERS[T8_ECLASS_TRIANGLE]; ++icorner) {
+      underlying_scheme.element_get_vertex_integer_coords (this->subelement_to_standalone (subelement), icorner,
+                                                           parent_coords[icorner].data ());
+    }
+
+    /* Lambda for the midpoints of a face of the parent triangle. */
+    const auto face_midpoint = [&parent_coords] (const int iface) {
```
spenke91, line 293: How about sth. like `compute_face_midpoint` emphasize its a function?

src/t8_schemes/t8_subelement/specializations/t8_scheme_tri.hxx
```diff
@@ -0,0 +1,338 @@
...
+    /* Build the path: Walk the parent edges from the first to the second end vertex of the main face, the way
+     * that does not traverse the main face itself.
+     */
+    const int start_vertex = t8_face_vertex_to_tree_vertex[T8_ECLASS_TRIANGLE][main_face][0];
+    const int end_vertex = t8_face_vertex_to_tree_vertex[T8_ECLASS_TRIANGLE][main_face][1];
+
+    // The path has maximal length 4 for 2 hanging faces.
+    // For the path we use the property of the triangle enumeration that the face has always the id of the opposite
+    // vertex (so the only vertex it is not adjacent to). Therefore we can use the face ids to get the midpoints of the hanging faces.
+    std::array<std::array<int, 2>, 4> path;
```
spenke91, line 316: `path_coords`, `vertex_coords_on_path`, or sth like that?

src/t8_schemes/t8_subelement/specializations/t8_scheme_quads.hxx
```diff
@@ -0,0 +1,378 @@
...
+*/
+
+/** \file t8_scheme_quads.hxx
+ * Subelement scheme specialization for quadrilateral elements. A quad is transitioned into
+ * triangular subelements (e.g. to resolve hanging nodes). The subelement type is a binary
+ * code over the quad's four faces indicating which of them are hanging; type 0 means the
+ * element is not a subelement (it is just the underlying standalone quad). This file only
+ * implements the quad-specific logic. The functionality shared by all subelement schemes
+ * lives in \ref t8_subelement_scheme.hxx.
+ */
```
spenke91, line 30: The file contents of this file and `t8_scheme_tri.hxx` is already pretty tailored to the hanging-node resolution, right? For, say, GPU-suitable subpatches a lot of the things would ot apply / look different, such as the subelement shape. Should we maybe 
(a) rename the files or the folder structure to be hanging-node specific or 
(b) split the files into aspects relevant for all triangle/quad-based subelement schemes and those specific to hanging-node resolution?

I guess (b) would be cleanest?

src/t8_schemes/t8_subelement/specializations/t8_scheme_quads.hxx
```diff
@@ -0,0 +1,378 @@
...
+  TUnderlyingScheme underlying_scheme {}; /**< Instance of the underlying standalone scheme. */
+
+  /** Compute the number of corners of an element.
+   * \param [in] elem The subelement.
+   * \return          The number of corners of \a elem.
+   */
+  static int
+  subelement_get_num_corners ([[maybe_unused]] const TSubelementType *elem) noexcept
+  {
+    return T8_ELEMENT_NUM_CORNERS[T8_ECLASS_TRIANGLE];
```
spenke91, line 82: Regarding my comment above about splitting this file into different levels of abstraction, maybe we can use a template parameter or something like this for the ECLASS of the subelements to keep parts like this generic.

src/t8_schemes/t8_subelement/specializations/t8_scheme_quads.hxx
```diff
@@ -0,0 +1,378 @@
...
+   */
+  static int
+  subelement_get_num_children ([[maybe_unused]] const t8_element_t *elem, int subelement_type)
+  {
+    int num_hanging_faces = 0;
+    /* Count the number of ones of the binary subelement type. This number equals the number of hanging faces. */
+    for (int i = 0; i < T8_ELEMENT_NUM_FACES[T8_ECLASS_QUAD]; ++i) {
+      num_hanging_faces += (subelement_type & (1 << i)) >> i;
+    }
+    return T8_ELEMENT_NUM_FACES[T8_ECLASS_QUAD] + num_hanging_faces;
```
spenke91, line 157:
```suggestion
    // Each original face "has" one triangular subelement, each split face two.
    return T8_ELEMENT_NUM_FACES[T8_ECLASS_QUAD] + num_hanging_faces;
```

src/t8_schemes/t8_subelement/specializations/t8_scheme_quads.hxx
```diff
@@ -0,0 +1,378 @@
...
+        + - - - - - - x         x - - - - - x        x - - x - - x        x - - - - - x        x - - - - - x
+   * \endverbatim
+   * Subelement ids are counted clockwise, starting with the (lower) left subelement with id 0.
+   * Note that we do not change the underlying quadrant.
+   */
+  void
+  subelement_get_children (const t8_element_t *elem, [[maybe_unused]] const int length, t8_element_t *c[],
+                           int type) const noexcept
+  {
+    const TSubelementType *element = this->as_subelement (elem);
```
spenke91, line 183:
```suggestion
    const TSubelementType *sub_elem = this->as_subelement (elem);
```
or similar to stick to the naming convention you use.

src/t8_schemes/t8_subelement/specializations/t8_scheme_quads.hxx
```diff
@@ -0,0 +1,378 @@
...
+    TSubelementType **subelements = reinterpret_cast<TSubelementType **> (c);
+    const int num_subelements = this->subelement_get_num_children (elem, type);
```
spenke91, line 185: see comment(s) in triangular subelement scheme

src/t8_schemes/t8_subelement/specializations/t8_scheme_quads.hxx
```diff
@@ -0,0 +1,378 @@
...
+  }
+
+ private:
+  /** Check whether a given face of the parent quad is hanging (and therefore split in half).
+   * \param [in] type  The subelement type (binary code over the faces, order is (f0 ,..., f_{numfaces-1})).
+   * \param [in] iface The face to check.
+   * \return           True if \a iface is hanging for \a type.
+   */
+  static bool
+  face_is_split (const unsigned type, const int iface) noexcept
```
spenke91, line 247: I think in `t8_scheme_tri` this is called `face_is_hanging`.

src/t8_schemes/t8_subelement/specializations/t8_scheme_quads.hxx
```diff
@@ -0,0 +1,378 @@
...
+  {
+    return ((type >> ((T8_ELEMENT_NUM_FACES[T8_ECLASS_QUAD] - 1) - iface)) & 1u) != 0u;
+  }
+
+  /** For each parent face, its two vertices in clockwise order. */
+  static constexpr int face_to_clockwise_vertex[4][2] = { { 0, 2 }, { 3, 1 }, { 1, 0 }, { 2, 3 } };
```
spenke91, line 253: I think we should make this consistent with the definition of `static constexpr int vertex_offset[4][2] = { { 0, 0 }, { 1, 0 }, { 0, 1 }, { 1, 1 } };`, so either both here or in `vertex_coords_of_subelement`

src/t8_schemes/t8_subelement/specializations/t8_scheme_quads.hxx
```diff
@@ -0,0 +1,378 @@
...
+    const int split = location[1];
+    const int sub_face_id = location[2];
+    /* Check, whether the get_location function provides meaningful location data. */
+    T8_ASSERT (face_number == 0 || face_number == 1 || face_number == 2 || face_number == 3);
+    T8_ASSERT ((split == 0 && sub_face_id == 0) || (split == 1 && (sub_face_id == 0 || sub_face_id == 1)));
+
+    /** The vertex offsets of a quad (as multiples of its edge length). */
+    static constexpr int vertex_offset[4][2] = { { 0, 0 }, { 1, 0 }, { 0, 1 }, { 1, 1 } };
+    /** Function lambda to get the vertex coordinates of the parent element. */
+    const auto vertex_coords_parent = [&] (const int vertex) {
```
spenke91, line 299:
```suggestion
    const auto get_vertex_coords_parent = [&] (const int vertex) {
```

src/t8_schemes/t8_subelement/specializations/t8_scheme_quads.hxx
```diff
@@ -0,0 +1,378 @@
...
+      return std::array<int, 2> { origin[0] + len * vertex_offset[vertex][0],
+                                  origin[1] + len * vertex_offset[vertex][1] };
+    };
+    /** Function lambda to get the midpoint of a face of the parent element. */
+    const auto vertex_midpoint_coords_parent = [&] (const int face) {
```
spenke91, line 304: `compute_vertex_midpoint_coords_parent`?

src/t8_schemes/t8_subelement/specializations/t8_scheme_quads.hxx
```diff
@@ -0,0 +1,378 @@
...
+    /* Vertex 0 is always the centre of the transition cell. */
+    vertex_coords[0] = { origin[0] + len / 2, origin[1] + len / 2 };
+    /* Vertices 1 and 2 are the face's two clockwise vertices, unless the face is split: then one of
+     * them is replaced by the face midpoint. */
+    const std::array<int, 2> vertex_start = vertex_coords_parent (face_to_clockwise_vertex[face_number][0]);
+    const std::array<int, 2> vertex_end = vertex_coords_parent (face_to_clockwise_vertex[face_number][1]);
+    const std::array<int, 2> face_midpoint = vertex_midpoint_coords_parent (face_number);
+
+    vertex_coords[1] = (split && sub_face_id) ? face_midpoint : vertex_start;
+    vertex_coords[2] = (split && !sub_face_id) ? face_midpoint : vertex_end;
```
spenke91, line 322: I already surrendered regarding this implicit is-nonzero check for boolean-like integers, but I really don't like it for other integer variables like the sub_face_id here. So I'd prefer
```suggestion
    vertex_coords[1] = (split && (sub_face_id==1)) ? face_midpoint : vertex_start;
    vertex_coords[2] = (split && (sub_face_id==0)) ? face_midpoint : vertex_end;
```
but feel free to overrule me, because it's probably just a matter of time I'll have to give up on these too :-)

src/t8_forest/t8_forest_subelement.hxx
```diff
@@ -0,0 +1,65 @@
...
+ *  Functionality to handle subelements in a forest.
+ */
+#pragma once
+
+#include <t8.h>
+#include "t8_forest_general.h"
+#include <t8_eclass/t8_eclass.h>
+#include <t8_schemes/t8_scheme.hxx>
+#include <t8_element/t8_element.h>
+
```
spenke91, line 33: In the long run, a function like `t8_forest_is_conforming` would also be useful I think

test/t8_forest/t8_gtest_subelement.cxx
```diff
@@ -0,0 +1,102 @@
...
+
+  /* Repartition the forest containing subelements (exercises MPI_Pack / MPI_Unpack). */
+  t8_forest_t forest_partitioned;
+  t8_forest_init (&forest_partitioned);
+  t8_forest_set_partition (forest_partitioned, forest, 0);
+  t8_forest_commit (forest_partitioned);
+
+  /* Subelements and leaf count must remain consistent after repartitioning. */
+  EXPECT_TRUE (t8_forest_has_global_subelements (forest_partitioned));
+  EXPECT_EQ (t8_forest_get_global_num_leaf_elements (forest_partitioned), num_leaves_sub);
```
spenke91, line 86: We could additionally check the number of trees etc but it's probably not necessary, right?

example/subelements/t8_quads_hanging_nodes.cxx
```diff
@@ -0,0 +1,115 @@
...
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with t8code; if not, write to the Free Software Foundation, Inc.,
+  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+*/
+
+/** \file t8_quads_hanging_nodes.cxx
+ * This is an example to demonstrate hanging node resolution for quads.
```
spenke91, line 24:
```suggestion
 * This is an example to demonstrate hanging node-resolution for quads. 
```

example/subelements/t8_quads_hanging_nodes.cxx
```diff
@@ -0,0 +1,115 @@
...
+#include <t8.h>                                       /* General t8code header, always include this. */
+#include <t8_cmesh/t8_cmesh.h>                        /* cmesh definition and basic interface. */
+#include <t8_cmesh/t8_cmesh_examples.h>               /* A collection of exemplary cmeshes */
+#include <t8_forest/t8_forest_general.h>              /* forest definition and basic interface. */
+#include <t8_forest/t8_forest_io.h>                   /* save forest */
+#include <t8_forest/t8_forest_geometrical.h>          /* geometrical information of the forest */
+#include <t8_forest/t8_forest_subelement.hxx>         /* Function for adding subelements. */
+#include <t8_schemes/t8_subelement/t8_subelement.hxx> /* Subelement refinement scheme. */
+#include <t8_types/t8_vec.h>                          /* Basic operations on 3D vectors. */
+#include <t8_eclass/t8_eclass.h>
```
spenke91, line 36: Kudos for the comments on what the headers include! 

```suggestion
#include <t8_types/t8_vec.h>                          /* Basic operations on 3D vectors. */
#include <t8_eclass/t8_eclass.h>                      /* Element-class definitions. */
```

example/subelements/t8_hanging_nodes_shock.cxx
```diff
@@ -0,0 +1,252 @@
...
+  forest_adapt = t8_forest_new_adapt (forest, t8_adapt_callback, 1, 0, &adapt_data);
+  return forest_adapt;
+}
+
+/** Adapt a forest around the circle of radius 0.6 (second adaptation cycle).
+ * Same criterion as \ref t8_adapt_forest but with a larger radius. 
+ * \param[in] forest Forest to be adapted.
+ */
+t8_forest_t
+t8_adapt_forest_2and (t8_forest_t forest)
```
spenke91, line 139:
```suggestion
t8_adapt_forest_2nd (t8_forest_t forest)
```
?

example/subelements/t8_hanging_nodes_shock.cxx
```diff
@@ -0,0 +1,252 @@
...
+  /* This is the inverse of the previous step and is required before adapting again. */
+  forest = t8_forest_discard_subelements (forest);
+  std::cout << "Subelements removed: " << t8_forest_has_global_subelements (forest) << std::endl;
+  const char *prefix_removed_sub = "t8_discarded_subelements1";
+  t8_forest_write_vtk (forest, prefix_removed_sub);
+  t8_global_productionf (" [subelements] Wrote adapted forest with discarded subelements to vtu files: %s*\n",
+                         prefix_removed_sub);
+
+  /* --- Second cycle: adapt around the larger circle. --- */
+  forest = t8_adapt_forest_2and (forest);
```
spenke91, line 228:
```suggestion
  forest = t8_adapt_forest_2nd (forest);
```

example/subelements/t8_hanging_nodes_shock.cxx
```diff
@@ -0,0 +1,252 @@
...
+t8_adapt_forest_2and (t8_forest_t forest)
+{
+  struct t8_adapt_data adapt_data = {
+    { 0, 1, 0 }, /* Center of the circle. */
+    0.6,         /* Radius */
+    0.1,         /* Delta (transition band width) */
+    2,           /* Minlevel */
+    6            /* Maxlevel */
+  };
+
```
spenke91, line 148: Wouldn't it be easier to define the whole adapt_data somewhere accessible from main and just adjust the radius then via `adapt_data.radius`?
Then we could also get rid of these two adapt wrappers. However, both is fine.

example/subelements/t8_hanging_nodes_shock.cxx
```diff
@@ -0,0 +1,252 @@
...
+  const int level = 0;
+  t8_forest_t forest = t8_forest_new_uniform (cmesh, t8_scheme_new_subelement (), level, 0, comm);
+  const char *prefix = "t8_uniform";
+  t8_forest_write_vtk (forest, prefix);
+  t8_global_productionf (" [subelements] Uniform forest wrote to file: %s*\n", prefix);
+
+  /* --- Adapt the forest: refine near the first circle, creating hanging nodes. --- */
+  forest = t8_adapt_forest (forest);
+  std::cout << "Subelements before removing: " << t8_forest_has_global_subelements (forest) << std::endl;
+  prefix = "t8_adapted1";
```
spenke91, line 200: Might be convenient to adjust the naming because Paraview otherwise shows the two `t8_adapted` ptvu files as series in the open-file menu. Not a big problem, but maybe it would be nice to avoid it, e.g.,

```suggestion
  prefix = "t8_adapted_1st";
```

src/t8_schemes/t8_subelement/specializations/t8_scheme_tri.hxx
```diff
@@ -0,0 +1,338 @@
...
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with t8code; if not, write to the Free Software Foundation, Inc.,
+  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+*/
+
+/** \file t8_scheme_tri.hxx
```
spenke91, line 23: I would suggest to use more descriptive names for these two files. I know the context is clear from the path, but the path is often not shown, e.g., in vs code. So maybe something like `t8_subelement_scheme_tri.hxx` or `t8_scheme_hanging_nodes_tri.hxx`?

spenke91 (changes requested):
Thanks a lot @lenaploetzke for this great feature! I'm glad to see the subelement concept being molded into an actual implementation 👍 
I like the overall code design and think it is great that we barely have to change the interface of the normal schemes. As you will see in one of the comments, I feel like we might be missing one layer of inheritance between "a subelement scheme for triangles" and "a subelement schemes for triangles that resolves haning nodes". 

Aside from that, most of my comments are minor suggestion and remarks, as usual :-)

test/t8_gtest_adapt_callbacks.cxx
lenaploetzke: Oh you are right!

example/subelements/t8_hanging_nodes_shock.cxx
lenaploetzke: Yes, nice!

src/t8_schemes/t8_subelement/t8_subelement_scheme.hxx
lenaploetzke: Its the same for the other schemes, but it may be valuable to call the deinit of them nevertheless.

src/t8_schemes/t8_subelement/t8_subelement_scheme.hxx
lenaploetzke: yes this is right (and this is done in all schemes with their element types).

test/t8_forest/t8_gtest_subelement.cxx
lenaploetzke: Hm yes we could but i think it must happen something very weird if the tree number is changed right?

src/t8_forest/t8_forest_subelement.hxx
lenaploetzke: Yes you are right. I implemented it but then noticed that i obviously need face nieghbors for this. Therefore it currently works only for forests without subelements... I would still leave it in the code because it still has a value for the test and neighbors will be added soon for the subelements. I added a check if we have subelements...

src/t8_schemes/t8_subelement/specializations/t8_scheme_quads.hxx
lenaploetzke: I think there is really not much which is relevant for all tri/quad-based subelements. 
But you are very right that the naming is not good. What do you think about t8_scheme_hanging_nodes_tri /_quad for the file and for the scheme t8_subelem_scheme_hanging_nodes_tri ?

src/t8_schemes/t8_subelement/specializations/t8_scheme_quads.hxx
lenaploetzke: The problem is that we can think of a subelement where we maybe have a mixture of quads and triangles (we also discussed another hanging node resolution scheme for quads with quads and triangles). So even this function is not generic i think. 
So i am not sure if it really makes sense to have different layers of abstraction.

src/t8_schemes/t8_subelement/specializations/t8_scheme_tri.hxx
lenaploetzke: I also like this but i wait for the decision with the abstraction layer before doing this

src/t8_schemes/t8_subelement/specializations/t8_scheme_tri.hxx
lenaploetzke: Yes, with the popcount this gets really easy so i dont add an extra function for this at the moment but keep it in mind if we need it more often :)

src/t8_schemes/t8_subelement/specializations/t8_scheme_tri.hxx
lenaploetzke: Yes nice idea, i did this here and and in the quad scheme.

src/t8_schemes/t8_subelement/specializations/t8_scheme_tri.hxx
lenaploetzke: I used \c now, hope this is fine

src/t8_schemes/t8_subelement/t8_subelement_scheme.hxx
lenaploetzke: In short this is because get_linear_id of a subelement returns just the linear id of the parent. But i think there is no good alternative for this. Therefore in some cases, another subelement is found with the same type but not the same id and checked for equality. I think everything should work without the check. This is not very nice but i also do not know another way at the moment.

src/t8_schemes/t8_subelement/README.md
spenke91: Thanks for adjusting. As I mentioned offline, I changed my mind here a little: Using the terms level, parent, etc. is also fine for the subelements. But just a comment, don't change it back, step is fine either way 🙂

src/t8_schemes/t8_subelement/t8_subelement_scheme.hxx
spenke91: Logically yes, but not with `reinterpret_cast`, so I was extra careful ;-)

test/t8_forest/t8_gtest_subelement.cxx
spenke91: Yeah, true! 😄

src/t8_schemes/t8_subelement/specializations/t8_scheme_hanging_nodes_tri.hxx
```diff
@@ -0,0 +1,338 @@
...
+    return T8_TRI_MAX_SUBELEMENT_TYPE;
+  }
+
+  /** Get the number of subelements an element is refined into for a specific type.
+   * \param [in] elem   The element whose number of children is returned.
+   * \param [in] subelement_type The subelement type used for refinement.
+   * \return                     The number of subelements the triangle is split into (hanging faces + 1).
+   */
+  static int
+  subelement_get_num_children ([[maybe_unused]] const t8_element_t *elem, int subelement_type)
```
spenke91: Fine with me👍

test/t8_forest/t8_gtest_subelement.cxx
```diff
@@ -50,26 +50,27 @@ TEST (t8_gtest_subelement, hybrid_hanging_nodes_visualization)
...
-  EXPECT_FALSE (t8_forest_has_global_subelements (forest));
+  EXPECT_FALSE (t8_forest_has_subelements (forest));
+  EXPECT_FALSE (t8_forest_is_conforming (forest));
   const t8_gloidx_t num_leaves_adapted = t8_forest_get_global_num_leaf_elements (forest);
 
   // Check that discarding without subelements just does nothing.
   forest = t8_forest_discard_subelements (forest);
-  EXPECT_FALSE (t8_forest_has_global_subelements (forest));
+  EXPECT_FALSE (t8_forest_has_subelements (forest));
   EXPECT_EQ (t8_forest_get_global_num_leaf_elements (forest), num_leaves_adapted);
```
spenke91: Now that I look at this again: Instead of checking the number of leaves, it might make sense to use
`EXPECT_FOREST_EQ` to actually compare the forests before and after the subelement steps (which of course would mean we'd have to store the forest from before). 
This would be the most thorough test of whether the two match, but like so often, just think about it and decide yourself :-)
(Same holds for the second subelement cycle below of course)

src/t8_schemes/t8_subelement/t8_subelement_scheme.hxx
spenke91: Sorry, I don't understand that yet 😇 Isn't `el1->subelement_id` the child id of the subelement within the parent (normal) element? To check for equality, we should compare them too, shouldn't we? How is their linear ID being that of the parent a problem in this context?

src/t8_schemes/t8_subelement/t8_subelement_scheme.hxx
```diff
@@ -272,6 +273,8 @@ struct t8_subelement_scheme_common:
 
   /** Check if two elements are equal. 
   * \note For subelements, it is only checked that the type is equal and not the id!!
+  * (This is because the linear id is the same for all subelements and sometimes, the wrong subelement is found. 
+  * At the moment, this is not a problem.)
```
spenke91, line 282: See comment below, I don't get that yet 🙈

src/t8_schemes/t8_subelement/specializations/t8_scheme_hanging_nodes_quads.hxx
```diff
@@ -0,0 +1,378 @@
...
+  TUnderlyingScheme underlying_scheme {}; /**< Instance of the underlying standalone scheme. */
+
+  /** Compute the number of corners of an element.
+   * \param [in] elem The subelement.
+   * \return          The number of corners of \a elem.
+   */
+  static int
+  subelement_get_num_corners ([[maybe_unused]] const TSubelementType *elem) noexcept
+  {
+    return T8_ELEMENT_NUM_CORNERS[T8_ECLASS_TRIANGLE];
```
spenke91: True, good point!

spenke91 (commented):
Thanks a lot for the update and replies. Here's some minor comments.
I think it's easiest to discuss the open questions, like a possible additional abstraction layer, in our meeting today 👍

src/t8_schemes/t8_subelement/specializations/t8_scheme_hanging_nodes_quads.hxx
```diff
@@ -0,0 +1,378 @@
...
+*/
+
+/** \file t8_scheme_quads.hxx
+ * Subelement scheme specialization for quadrilateral elements. A quad is transitioned into
+ * triangular subelements (e.g. to resolve hanging nodes). The subelement type is a binary
+ * code over the quad's four faces indicating which of them are hanging; type 0 means the
+ * element is not a subelement (it is just the underlying standalone quad). This file only
+ * implements the quad-specific logic. The functionality shared by all subelement schemes
+ * lives in \ref t8_subelement_scheme.hxx.
+ */
```
lenaploetzke: Done

src/t8_schemes/t8_subelement/t8_subelement_scheme.hxx
spenke91: After our discussion yesterday, it is clear now :-)

spenke91 (dismissed):
Looks great, thanks! :-) The only open optional comment is the `EXPECT_FOREST_EQ` one, which, as I said, is optional 👍

test/t8_forest/t8_gtest_subelement.cxx
lenaploetzke: Yes and no, i thinkfor the second cycle we cannot do this because of the "partition" part (where we partition with subelements) so we cannot use FOREST_EQ there right? But here, it is a very good idea :)

spenke91: Well, I guess we could repartition the forest again (and the "original" one too), then the two should be the same again. But again, might be an overkill 🙂

spenke91 (approved):
Looks great 👍

test/t8_forest/t8_gtest_subelement.cxx
lenaploetzke: Hm yes i guess we could do this ,but i also think it may be an overkill here.

Issue #2035, opened 2025-12-10
external header includes internal header


lukasdreyer:
# Bug report

**Describe the bug**
The external header t8_cmesh.hxx includes the internal header t8_cmesh_types.hxx, using it to access internal types in the templated function register_geometry.
Thus, external applications are not able to include t8_cmesh.hxx.

**To Reproduce**
'#include <t8_cmesh/t8_cmesh.hxx>'
from an external application using t8code as a library.

**Expected behavior**
clean inclusion

**Estimated priority**
Which of these is most applicable (remove the others):

"Priority: high"  Should be solved as soon as possible

PR #2433, opened 2026-09-18
Bugfix: Feature: Add geometry handler getter to cmesh

### LINKED ISSUE #2434: Bugfix: Feature: Add geometry handler getter to cmesh

### LINKED ISSUE #2435: Cmesh.hxx uses internal geometry_handler
# Bug report

**Describe the bug**
A subissue of #2035 
The public cmesh.hxx accesses the internal variable geometry_handler and thus needs an internal header.

**To Reproduce**
include cmesh.hxx in an external library

**Expected behavior**
bugfree linkage

**Log files**

**Environment (please complete the following information):**

**Estimated priority**
"Priority: high"  Should be solved as soon as possible

**Additional context**
See #2035

holke:
Closes #2434

Related to #2035 , fixes parts of it.
Closes #2435 (subissue of #2035)

We have the issue that the public cmesh.hxx includes the private cmesh_types.h header.
Thus, no external library can include cmesh.hxx.

One of the sub-issues was the geometry handler accessed via cmesh->geometry_handler.
Solved by adding a new getter function for the geometry handler.

Since it happens in a template, i cannot move the function into a cxx file (where the internal header could be included).
Thus, another solution was required.

I am not fully happy with this solution, since now users can get non-const access to the geometry_handler and could mess up any of our geometry settings.
We just need to trust them not to do so.

#### General
#### Tests

#### Scripts and Wiki
#### License


src/t8_cmesh/t8_cmesh.hxx
```diff
@@ -106,11 +106,12 @@ template <typename geometry_type, typename... _args>
 inline geometry_type *
 t8_cmesh_register_geometry (t8_cmesh_t cmesh, _args &&...args)
 {
-  if (cmesh->geometry_handler == NULL) {
+  t8_geometry_handler_c *geometry_handler = t8_cmesh_get_geometry_handler (cmesh);
+  if (geometry_handler == NULL) {
     /* The handler was not constructed, do it now. */
-    cmesh->geometry_handler = new t8_geometry_handler ();
+    geometry_handler = new t8_geometry_handler ();
```
sandro-elsweijer: we need a register geometry handler function here, because you are only assigning the new handler to the local pointer. the cmesh pointer will still point to nullptr.
I had the same problem here and introduced a new function for this kind of problem:
https://github.com/DLR-AMR/t8code/blob/bf61cad528ff99b28ba6cb6f22ece07dab38f459/src/t8_cmesh/t8_cmesh_boundary_conditions/t8_cmesh_boundary_conditions.hxx#L47-L57

holke: Yes, thanks for pointing this out.

It took a little extra effort, especially since we now hide the geometry_handler in the detail namespace.

src/t8_cmesh/t8_cmesh_internal/t8_cmesh_geometry_internal.hxx
```diff
@@ -0,0 +1,37 @@
+/*
+  This file is part of t8code.
+  t8code is a C library to manage a collection (a forest) of multiple
+  connected adaptive space-trees of general element classes in parallel.
+
+  Copyright (C) 2015 the developers
```
sandro-elsweijer, line 6:
```suggestion
  Copyright (C) 2026 the developers
```

src/t8_cmesh/t8_cmesh.hxx
```diff
@@ -94,25 +94,6 @@ t8_cmesh_get_first_element_of_process (const uint32_t process, const uint32_t mp
...
-inline geometry_type *
-t8_cmesh_register_geometry (t8_cmesh_t cmesh, _args &&...args)
-{
-  if (cmesh->geometry_handler == NULL) {
-    /* The handler was not constructed, do it now. */
-    cmesh->geometry_handler = new t8_geometry_handler ();
-  }
-  return cmesh->geometry_handler->register_geometry<geometry_type> (std::forward<_args> (args)...);
-}
-
```
sandro-elsweijer: I thought the whole purpose was to fix this and not to delete this?

src/t8_cmesh/t8_cmesh_geometry.cxx
```diff
@@ -52,11 +54,75 @@ t8_cmesh_set_tree_geometry (t8_cmesh_t cmesh, const t8_gloidx_t gtreeid, const t
...
+ * \note                   The return value might be NULL if no geometry handler exists.
+ * \note                   Handle with care. This function should be used by t8code devs only.
+ */
+t8_geometry_handler_c *
+t8_cmesh_get_geometry_handler_c (const t8_cmesh_t cmesh)
+{
+  T8_ASSERT (t8_cmesh_is_initialized (cmesh) || t8_cmesh_is_committed (cmesh));
+
+  return cmesh->geometry_handler;
+}
```
sandro-elsweijer, line 69: Do we even need the C version?
The user should not directly interact with the handler and since it can also just really be used with Cpp, we do not need a C interface.

src/t8_cmesh/t8_cmesh_geometry.cxx
```diff
@@ -52,11 +54,75 @@ t8_cmesh_set_tree_geometry (t8_cmesh_t cmesh, const t8_gloidx_t gtreeid, const t
...
+
+/* Set a geometry handler or construct a new geometry_handler for a cmesh and add it to the cmesh.
+ * \param [in] cmesh      The cmesh to be considered. Must be initialized. Does not need to be committed.
+ * \param [in] new_handler  The geometry handler to be set. If nullptr then a new handler will be allocated.
+ * \return                On success, the new geometry_handler. nullptr on failure (out of memory).
+ */
+detail::t8_geometry_handler *
+t8_cmesh_set_geometry_handler (t8_cmesh_t cmesh, detail::t8_geometry_handler *new_handler)
+{
+  T8_ASSERT (t8_cmesh_is_initialized (cmesh) || t8_cmesh_is_committed (cmesh));
```
sandro-elsweijer, line 103: I do not think we should be able to set a new geometry handler for an already committed cmesh

src/t8_cmesh/t8_cmesh_geometry.cxx
```diff
@@ -52,11 +54,75 @@ t8_cmesh_set_tree_geometry (t8_cmesh_t cmesh, const t8_gloidx_t gtreeid, const t
...
+  T8_ASSERT (t8_cmesh_get_geometry_handler (cmesh) == nullptr);
+
+  // If not present allocate a new handler
+  if (new_handler == nullptr) {
+    new_handler = new detail::t8_geometry_handler ();
+  }
+  else {
+    // Increase reference count of handler
+    new_handler->ref ();
+  }
```
sandro-elsweijer, line 115: I would say the cmesh takes ownership of the handler just as our forest does with the scheme and the cmesh.
It would be weird if the cmesh would behave differently here.
```suggestion
```

src/t8_cmesh/t8_cmesh.h
```diff
@@ -61,6 +61,24 @@ typedef struct t8_ctree *t8_ctree_t;
...
+ * We need it so that we can use t8_geometry_handler_c pointers in .c files
+ * without them seeing the actual C++ code (and then not compiling)
+ * TODO: Delete this when the cmesh is a proper cpp class.
+ */
+#ifdef __cplusplus
+
+namespace detail
+{
+struct t8_geometry_handler;
+}
```
sandro-elsweijer: No forward declaration needed here.
Since this part is only compiled in cpp you can directly include the correct header file.
This way `t8_geometry_handler_c` and `detail::t8_geometry_handler` really are the same in every cpp file which includes this header and reinterpret_cast is no longer needed.
```suggestion
#include <t8_geometry/t8_geometry_handler.hxx>
```

src/t8_cmesh/t8_cmesh_geometry.cxx
```diff
@@ -52,11 +54,75 @@ t8_cmesh_set_tree_geometry (t8_cmesh_t cmesh, const t8_gloidx_t gtreeid, const t
...
+  // Convert the handler to C pointer and add to cmesh
+  t8_geometry_handler_c *new_handler_c = detail::t8_geom_handler_to_c (new_handler);
+  return cmesh->geometry_handler = new_handler_c;
```
sandro-elsweijer, line 118: This should work if you include the handler header in cmesh.h as discussed above
```suggestion
  return cmesh->geometry_handler = new_handler;
```

src/t8_cmesh/t8_cmesh_geometry.hxx
```diff
@@ -42,6 +43,42 @@ T8_EXTERN_C_BEGIN ();
...
+detail::t8_geometry_handler *
+t8_cmesh_get_geometry_handler (const t8_cmesh_t cmesh);
+
+/** Construct a new geometry_handler for a cmesh and add it to the cmesh.
+ * \param [in] cmesh      The cmesh to be considered. Must be initialized. Does not need to be committed.
+ * \return                On success, the new geometry_handler. nullptr on failure (out of memory).
+ * \note                  Handle with care. This function should be used by t8code devs only.
+ */
+detail::t8_geometry_handler *
+t8_cmesh_add_geometry_handler (t8_cmesh_t cmesh);
```
sandro-elsweijer, line 60: These should not bee needed to be wrapped in T8_EXTERN_C since they are pure cpp functions

src/t8_geometry/t8_geometry.h
```diff
@@ -59,13 +59,6 @@ typedef enum t8_geometry_type {
  */
 typedef struct t8_geometry t8_geometry_c;
 
-/** This typedef holds virtual functions for the geometry handler.
- * We need it so that we can use t8_geometry_handler_c pointers in .c files
- * without them seeing the actual C++ code (and then not compiling)
- * TODO: Delete this when the cmesh is a proper cpp class.
- */
-typedef struct t8_geometry_handler t8_geometry_handler_c;
-
```
sandro-elsweijer, line 68: why do we place the typedefs in t8_cmesh.h and not here?

src/t8_geometry/t8_geometry_handler.hxx
```diff
@@ -339,4 +345,20 @@ struct t8_geometry_handler
...
+  return reinterpret_cast<t8_geometry_handler_c *> (handler);
+}
+
+/** Conversion helpers to convert from C to C++ */
+inline detail::t8_geometry_handler *
+t8_geom_handler_from_c (t8_geometry_handler_c *handler)
+{
+  return reinterpret_cast<detail::t8_geometry_handler *> (handler);
+}
+
```
sandro-elsweijer, line 361: this should not be needed with the proposed include in the typedef

```suggestion
```

src/t8_cmesh/t8_cmesh.h
sandro-elsweijer, line 74: No forward declaration needed here.
Since this part is only compiled in cpp you can directly include the correct header file.
This way `t8_geometry_handler_c` and `detail::t8_geometry_handler` really are the same in every cpp file which includes this header and reinterpret_cast is no longer needed.
```suggestion
#include <t8_geometry/t8_geometry_handler.hxx>
```

src/t8_cmesh/t8_cmesh_geometry.cxx
sandro-elsweijer: and why do we put the setter for the handler into an internal file, but the getter not? Imho the user should never be exposed to the handler.

sandro-elsweijer (changes requested):

